PR #7, opened 2017-12-18
Added missing named lambda parameter

LouisCAD:


robstoll:
Thanks for your contribution. `subject` is actually a property of IAssertionPlant and not a param of the lambda. So there is no need to define it as param (correct me, but I think it is even wrong, would not compile). 
Nevertheless, maybe you are not the only one which thought a param is missing. Do you want to add an explanation instead? Somewhere near "In the above example, subject within the assertion group block refers to myPerson."

LouisCAD:
Oh I see.
It should still compile and execute fine. You'd have to use `this.subject` to access the hidden property of `IAssertionPlant`, and `subject` for the lambda parameter.

My confusion came from the fact that `subject` was close to "person", so I'd write that it's the subject of the assertion to emphasize the distinction with the fact that it's a person.

robstoll:
Please feel free to improve the documentation or close the pull request and I look into it myself

README.md
```diff
@@ -233,7 +233,7 @@ So the above output can be read as "I assert, Person's name (which is actually `
 :information_source: You can also use `it` as alternative for `subject`. The above could have been written as:
 ```kotlin
 assert(person) {
-    property(it::name).toBe("Peter")
+    property(it::name).toBe("Peter") // The assertion subject is also passed as an implicit lambda parameter so you can use it
```
robstoll: That's actually not true. `it` is a property as well. I added it because people are used to use `it` as shortcut for whatever their main focus is.

README.md
```diff
@@ -213,7 +213,7 @@ data class Person(val name: String, val isStudent: Boolean)
 val myPerson = Person("Robert", false) 
 
 assert(myPerson) {
-    property(subject::name).toBe("Peter")
+    property(subject::name).toBe("Peter") // subject is a property of the receiver. This is the assertion subject
```
robstoll: I would prefer to keep explanations in the text and not in code

README.md
LouisCAD: Why not use lambda parameter instead? This creates confusion...

robstoll: Yeah I see that it creates confusion. I should either remove `it` entirely or pass it as param. I am not sure which is better. Right now, I think passing it as param is overkill. It is a lambda with receiver where `subject` is available anyway. What do you think?

LouisCAD: If only `subject` is needed, then I'd not add any receiver, and I'd just pass the subject as single lambda parameter, giving `it` and the freedom to name it as one likes.
If the receiver is needed, then I think keeping `subject` only would be right and avoid any confusion or ambiguity.

robstoll: It has to be a receiver lambda, otherwise you could not use the assertion functions (they are all defined on IAssertionPlant). I'll consider to remove `it`. Thank you very much for your time and for your feedback :-)

LouisCAD: You're welcome :-)

PR #19, opened 2018-04-26
first steps need more guidance, installation + verbs

robstoll:

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,8 @@ dependencies {
     testCompile "ch.tutteli:atrium-cc-en_UK-robstoll:$atrium_version"
 }
 ```
+We have defined a dependency to the bundle `atrium-cc-en_UK-robstoll` in the above example. 
+Adjust it if you use [another bundle](#apis) (e.g. `atrium-cc-infix-en_UK-robstoll` if you use the infix api).
 
 *maven*:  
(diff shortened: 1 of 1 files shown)


robstoll:
Thanks for your feedback. I probably have to give an example which puts an emphasis on why you want to use your own assertion verb and not a predefined assertion verb or at least a function which delegates to the predefined assertion verb. 
Say you are using Atrium in 20 projects, in each you use the predefined `expect`. At some point you get the business requirement, that you should generate an HTML report out of the spec (a feature which is not yet available in Atrium but which I plan to implement in the future). Now you have to use your own assertion verb, because the predefined assertion verb does not do anything like that. That means you have to touch 20 projects, use search for the import and replace it with your own assertion verb. If you haved used an own assertion verb from the beginning then you would only adjust the ReporterBuilder in one place (ideally you did not define the assertion verb in 20 places :wink:).
I don't remember in which context something similar happened to me but I remember that it was very cumbersome and I wished that the library was more flexible; told me from the beginning that there was a way to get around this problem (I think it was not even possible).

I hope you see why I encourage users to define an own assertion verb. Nevertheless, I see that it is too much at the beginning and that I should adjust the design so that one can specify/configure a ReporterBuilder even if one uses a predefined assertion verb. I'll think about it... thank's again for your feedback :+1:

Issue #66, opened 2019-01-12
allow to pass single Char to starts(Not)With/ends(Not)With


robstoll:
*Platform* (JVM, JS, android): all

## Code related feature
check if a CharSequence startsWith/endsWith a certain Char should also be supported next to CharSequence.
```
assert("hello world").startsWith('h')
```

Add overloads to startsWith, startsNotWith, endsWith and endsNotWith in:
- api-fluent-en_GB-common

which expects a `Char` and delegates to the existing function by converting it to a `String`.
(the functions are located in files named charSequenceAssertions.kt)

# Your first contribution?
- write a comment if you would like to take this task over, I might give you some hints in addition or update the task in case things have changed in the meantime
- do not hesitate to contact me via slack if you need help: https://kotlinlang.slack.com/team/U3DE1TXKP ([Invite yourself](http://slack.kotlinlang.org/) in case you do not have an account yet)

kssc0112:
If overload startsWith like 
``` fun startsWith(subjectProvider: SubjectProvider<CharSequence>, expected: Char): Assertion```
I get the following error in the spec. Is there a workaround for this? If not, I'll continue with creating startsWithChar, startsNotWithChar and so on.

<img width="1239" alt="Screen Shot 2019-07-17 at 11 03 02 PM" src="https://user-images.githubusercontent.com/15913002/61433098-3d3f9680-a8e7-11e9-80e8-f714b9a933b1.png">

**Comment by robstoll** [comment]
there is a workaround, see for instance ListFeatureAssertionsSpec

**Comment by robstoll** [comment]
roughly:
```
import kotlin.reflect.KFunction1
class CharSequenceAssertionsSpec : ch.tutteli.atrium.specs.integration.CharSequenceAssertionsSpec(
    ...
    startsWithChar.name to startsWithChar
){
    companion object{
        val startsWithChar: KFunction1<Expect<CharSequence>, Char, Expect<CharSequence>> = Expect<CharSequence>::get
    }
}
```

kssc0112:
@robstoll Thanks. Almost done for a PR. What is the purpose of calling forSubjectLess and forChecking APIs in CharSequenceAssertionsSpec? Do we want it for the Char versions as well?

Ex. `startsWithPair.forChecking("a", "abc", "xyz")`
`startsWithPair.forSubjectLess("")`

robstoll:
those are two basic specs which every assertion function should hold -- so yes, please add it also for Char. 

[...]

PR #100, opened 2019-07-15
switch to 'implement' instead of 'compile' in build.gradle where possible

kssc0112:
Following changes were announced with Gradle 3.0:
1) The 'compile' configuration is now deprecated and should be replaced by 'implementation'.
2) The 'testCompile' configuration is now deprecated and should be replaced by' testImplementation'.

----

### DIFF
--- a/apis/atrium-api-cc-en_UK/build.gradle
+++ b/apis/atrium-api-cc-en_UK/build.gradle
@@ -3,10 +3,10 @@ description = 'Deprecated, use atrium-api-cc-en_GB-jvm instead.'
 ext.jacoco_additional = [prefixedProject('translations-en_GB-jvm')]
 
 dependencies {
-    compile prefixedProject('domain-api-deprecated')
-    compile prefixedProject('domain-builders-jvm')
+    api prefixedProject('domain-api-deprecated')
+    api prefixedProject('domain-builders-jvm')
 
     testRuntimeOnly prefixedProject('domain-robstoll-jvm')
     testRuntimeOnly prefixedProject('core-robstoll-jvm')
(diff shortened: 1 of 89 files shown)


kssc0112:
@robstoll Do you have any idea why 'compileKotlinCommon' task is failing?

robstoll:
@kssc0112 the problem is that `implementation` does not propagate dependencies to consumer modules in contrast to `compile`. I was a bit sloppy in the description, sometimes we have to use `api` instead of `implementation` since we want to propagate the dependencies as transitive dependencies (see https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_recognizing_dependencies for details). Following a list of modules which should be depended on with `api` instead `implementation`:
- domain-api-deprecated
- domain-builders-...
- api-...
- verbs-...
- core-api-...
- domain-api-...

kssc0112:
@robstoll Make sense. Thanks. Will make the update.

kssc0112:
@robstoll I am still getting an error with on atrium-domain-builders-common-compileKotlinCommon task. Can't really tell what is wrong. Please see the attached stacktrace.
[atrium-domain-builders-common-compileKotlinCommon_error.log](https://github.com/robstoll/atrium/files/3400378/atrium-domain-builders-common-compileKotlinCommon_error.log)

robstoll:
I see the problem, it comes from using kbox, I am using it everywhere, so it should be `api` as well. I'll fix it myself. Thank you very much for your PR :slightly_smiling_face::+1: 

Would you like to to have a go at another Issue? Maybe more Kotlin related? I would suggest #66

PR #105, opened 2019-07-20
allow to pass single Char to starts(Not)With/ends(Not)With

kssc0112:
----

### DIFF
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceAssertions.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceAssertions.kt
@@ -107,6 +107,15 @@ fun <T : CharSequence> Expect<T>.startsWith(expected: CharSequence)
     = addAssertion(ExpectImpl.charSequence.startsWith(this, expected))
 
 /**
+ * Expects that the subject of the assertion starts with [expected].
+ *
+ * @return This plant to support a fluent API.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ */
+fun <T : CharSequence> Expect<T>.startsWith(expected: Char)
+    = startsWith(expected.toString())
+
+/**
  * Expects that the subject of the assertion does not start with [expected].
  *
  * @return This plant to support a fluent API.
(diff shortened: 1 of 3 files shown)


domain/api/atrium-domain-api-common/src/main/kotlin/ch/tutteli/atrium/domain/creating/CharSequenceAssertions.kt
```diff
@@ -24,9 +24,13 @@ interface CharSequenceAssertions {
     fun <T : CharSequence> containsNotBuilder(subjectProvider: SubjectProvider<T>): CharSequenceContains.Builder<T, NotSearchBehaviour>
 
     fun startsWith(subjectProvider: SubjectProvider<CharSequence>, expected: CharSequence): Assertion
+    fun startsWith(subjectProvider: SubjectProvider<CharSequence>, expected: Char): Assertion
```
robstoll: Yapp, I still think delegation is good enough (this applies to the other methods you added here as well). Unless you see a use case where we want to treat `Char` different than `CharSequence`, in this case we can discuss it. Otherwise I think it only clutters the API without real value.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceAssertions.kt
```diff
@@ -106,6 +106,15 @@ fun <T : CharSequence> Expect<T>.containsRegex(pattern: String, vararg otherPatt
...
     = addAssertion(ExpectImpl.charSequence.startsWith(this, expected))
 
+/**
+ * Expects that the subject of the assertion starts with [expected].
+ *
+ * @return This plant to support a fluent API.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ */
+fun <T : CharSequence> Expect<T>.startsWith(expected: Char)
+    = addAssertion(ExpectImpl.charSequence.startsWith(this, expected))
```
robstoll:
```suggestion
    = startsWith(expected.toString())
```

domain/api/atrium-domain-api-common/src/main/kotlin/ch/tutteli/atrium/domain/creating/CharSequenceAssertions.kt
kssc0112: Make sense. Will update

robstoll:
LGTM many thanks @kssc0112

robstoll:
@kssc0112 off to a hat-trick? I would suggest #108 as next issue

PR #123, opened 2019-08-23
Shortcuts for path existance

### LINKED ISSUE #108: shorcut for Path.exists and Path.existsNot
*Platform* (JVM, JS and/or Android): jdk8

## Code related feature
```
expect(Paths.get("xyz")).exists()
expect(Paths.get("nonExisting").existsNot()
//instead of
expect(Paths.get("xyz")).feature(Files::exists, arrayOf()) { toBe(true) }
expect(Paths.get("xyz")).feature(Files::notExists, arrayOf()) { toBe(true) }
```

*domain*
- extend PathAssertions with `existsNot` (`exists` is already provided)
- modify PathAssertionsBuilder, delegate to pathAssertions (as for `exists`)
- delegate implementation to robstoll-lib in PathAssertionsImpl (as for `exists`)

*lib*
- implement _exists and _existsNot (I have already setup a dependency to niok which provides Path.exists)

*api*
- add  `existsNot` (`exists` is already provided)
- extend PathAssertionsSpec in specs-common

## Your first contribution?
  ([Invite yourself](http://slack.kotlinlang.org/) in case you do not have an account yet).

jGleitz:
Implements Matchers for `Path#exists` and `Path#existsNot`.

Deviates from the proposed course of action in #108 as so far as I have adapted the message to be less noisy while keeping the same amount of information. The message in this PR match the ones used for `Collection#isEmpty` and the like.

So instead of:

```
expect: notExisting        (sun.nio.fs.UnixPath <749882615>)
◆ ▶ exists: false
    ◾ to be: true
```
this PR creates this message:
```
expect: notExisting        (sun.nio.fs.UnixPath <749882615>)
◆ to: exist
```

just like we get:
```
expect: [not-empty]        (java.util.Collections.SingletonList <1286606146>)
◆ is: empty
```

closes #108 

----

### DIFF
--- a/domain/api/atrium-domain-api-jvm/src/main/kotlin/ch/tutteli/atrium/domain/creating/PathAssertions.kt
+++ b/domain/api/atrium-domain-api-jvm/src/main/kotlin/ch/tutteli/atrium/domain/creating/PathAssertions.kt
@@ -17,5 +17,6 @@ val pathAssertions by lazy { loadSingleService(PathAssertions::class) }
  * which an implementation of the domain of Atrium has to provide.
  */
 interface PathAssertions {
-    fun <T: Path> exists(assertionContainer: Expect<T>): Assertion
+    fun <T : Path> exists(assertionContainer: Expect<T>): Assertion
+    fun <T : Path> existsNot(subjectProvider: Expect<T>): Assertion
 }
--- a/domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/pathAssertions.kt
+++ b/domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/pathAssertions.kt
@@ -2,15 +2,16 @@ package ch.tutteli.atrium.domain.robstoll.lib.creating
 
 import ch.tutteli.atrium.assertions.Assertion
 import ch.tutteli.atrium.creating.Expect
+import ch.tutteli.atrium.domain.builders.ExpectImpl
+import ch.tutteli.atrium.reporting.RawString
+import ch.tutteli.atrium.translations.DescriptionAnyAssertion.NOT_TO
+import ch.tutteli.atrium.translations.DescriptionAnyAssertion.TO
+import ch.tutteli.atrium.translations.DescriptionPathAssertion.EXIST
+import ch.tutteli.niok.exists
+import ch.tutteli.niok.notExists
 import java.nio.file.Path
 
-//TODO #104 remove annotation
-@Suppress("UNUSED_PARAMETER")
 fun <T : Path> _exists(assertionContainer: Expect<T>): Assertion =
-    TODO(
-        """
-        #104 create a feature assertion for Path::exists and assert it is true.
-        Notice, there is a bug in the new-type-inference https://youtrack.jetbrains.com/issue/KT-32851 
-        the bug is only within IDEA, check gradle if it works
-        """
-    )
+    ExpectImpl.builder.createDescriptive(assertionContainer, TO, RawString.create(EXIST)) { it.exists }
+fun <T : Path> _existsNot(subjectProvider: Expect<T>): Assertion =
+    ExpectImpl.builder.createDescriptive(subjectProvider, NOT_TO, RawString.create(EXIST)) { it.notExists }
--- a/misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
+++ b/misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
@@ -5,6 +5,7 @@ import ch.tutteli.atrium.api.fluent.en_GB.toThrow
 import ch.tutteli.atrium.api.verbs.internal.expect
 import ch.tutteli.atrium.specs.*
 import ch.tutteli.atrium.translations.DescriptionAnyAssertion
+import ch.tutteli.atrium.translations.DescriptionPathAssertion
 import ch.tutteli.spek.extensions.TempFolder
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.Suite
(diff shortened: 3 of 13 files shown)


jGleitz:
Blocked by https://github.com/robstoll/niok/issues/6

jGleitz:
See also #124 (to not compile with Java 9, but Java 12 in CI)

domain/api/atrium-domain-api-jvm/src/main/kotlin/ch/tutteli/atrium/domain/creating/PathAssertions.kt
```diff
@@ -17,5 +17,6 @@ val pathAssertions by lazy { loadSingleService(PathAssertions::class) }
  * which an implementation of the domain of Atrium has to provide.
  */
 interface PathAssertions {
-    fun <T: Path> exists(assertionContainer: Expect<T>): Assertion
+    fun <T : Path> exists(subjectProvider: SubjectProvider<T>): Assertion
```
robstoll: Please use `Expect` instead of `SubjectProvider`, most probably I am going to deprecate SubjectProvider with 1.0.0. 
It was only introduced to ease the migration from `Assert` to `Expect`

The same applies to other places in your PR

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
```diff
@@ -57,5 +57,38 @@ abstract class PathAssertionsSpec(
...
+                    verbs.check(file).existsNotFun()
+                }.toThrow<AssertionError> {
+                    messageContains(expectedMessageIfExisting)
+                }
+            }
+        }
+
+        context("existing folder") {
+            it("throws an AssertionError") {
+                val folder = tempFolder.newFile("exists-though-shouldnt")
```
robstoll: newFolder

translations/en_GB/atrium-translations-en_GB-common/src/main/kotlin/ch/tutteli/atrium/translations/DescriptionAnyAssertion.kt
```diff
@@ -7,6 +7,8 @@ import ch.tutteli.atrium.reporting.translating.StringBasedTranslatable
  * Contains the [DescriptiveAssertion.description]s of the assertion functions which are applicable to [Any].
  */
 enum class DescriptionAnyAssertion(override val value: String) : StringBasedTranslatable {
+    TO("to"),
```
robstoll: I have divided the `Description` enums into types where they are used. This should either be placed into `DescriptionPathAssertion` or into `DescriptionBasic` if it is something we want to use in other places as well.
Maybe it makes even sense to move `TO_BE` and `NOT_TO` to DescriptionBasic - up to you if you want to do that as well

robstoll:
Please rebase on master, it already uses the new version of niok. Notice that you will most likely run into a conflict in PathAssertionsSpec.kt as I have removed AssertionVerbFactory. Just remove it as well and you should be good again.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/pathAssertions.kt
```diff
@@ -1,16 +1,18 @@
...
-fun <T : Path> _exists(assertionContainer: Expect<T>): Assertion =
-    TODO(
-        """
-        #104 create a feature assertion for Path::exists and assert it is true.
-        Notice, there is a bug in the new-type-inference https://youtrack.jetbrains.com/issue/KT-32851 
-        the bug is only within IDEA, check gradle if it works
-        """
-    )
+fun <T : Path> _exists(subjectProvider: SubjectProvider<T>): Assertion =
+    ExpectImpl.builder.createDescriptive(subjectProvider, TO, RawString.create(EXIST)) { it.exists }
```
robstoll: Good idea :+1:

domain/api/atrium-domain-api-jvm/src/main/kotlin/ch/tutteli/atrium/domain/creating/PathAssertions.kt
jGleitz: done

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
jGleitz: Oh, right. Done.

translations/en_GB/atrium-translations-en_GB-common/src/main/kotlin/ch/tutteli/atrium/translations/DescriptionAnyAssertion.kt
jGleitz: I strongly feel that `TO` should be where `TO_BE` is. That is why I used `DescriptionAnyAssertion` in the first place.

I will move all four to `DescriptionBasic` in a separate pull request so this one keeps reviewable. I have already made the changes, the pull request is pending because I get some weird encoding issues in the tests.

jGleitz:
I have incorporated all changes and rebased on `master`. CodeCov fails, but the reason seems stupid (and also wrong!)

domain/api/atrium-domain-api-jvm/src/main/kotlin/ch/tutteli/atrium/domain/creating/PathAssertions.kt
```diff
@@ -17,5 +17,6 @@ val pathAssertions by lazy { loadSingleService(PathAssertions::class) }
  * which an implementation of the domain of Atrium has to provide.
  */
 interface PathAssertions {
-    fun <T: Path> exists(assertionContainer: Expect<T>): Assertion
+    fun <T : Path> exists(assertionContainer: Expect<T>): Assertion
+    fun <T : Path> existsNot(subjectProvider: Expect<T>): Assertion
```
robstoll: also use assertionContainer instead of subjectProvider, same same at other places

robstoll:
I'll do the tiny correction myself, thank you very much for your PR and other remarks and ideas. Fancy to continue with another one?

PR #125, opened 2019-08-26
Description Enum Refactoring

jGleitz:
:warning: Do not merge this PR before #123 is merged!

As discussed in #123:

> I have divided the `Description` enums into types where they are used. This should either be placed into `DescriptionPathAssertion` or into `DescriptionBasic` if it is something we want to use in other places as well.
Maybe it makes even sense to move `TO_BE` and `NOT_TO` to `DescriptionBasic` - up to you if you want to do that as well

I moved all four constants to `DescriptionBasic`. `TO_BE` and `NOT_TO_BE` still exist in `DescriptionAnyAssertion` for backwards compatibility.

----


robstoll:
Perfect, would you like to do the same for de_CH Descriptions in translations/de_CH/?

jGleitz:
> Perfect, would you like to do the same for de_CH Descriptions in translations/de_CH/? 

I am not sure what you mean, I made the change in `DescriptionAnyAssertion` in both `en_GB` and `de_CH`.

robstoll:
> I am not sure what you mean, I made the change in DescriptionAnyAssertion in both en_GB and de_CH.

Sorry, my mistake, I saw only en_GB this morning

core/robstoll-lib/atrium-core-robstoll-lib-jvm/src/test/resources/ch/tutteli/atrium/core/robstoll/lib/reporting/translating/Atrium_fr.properties
```diff
@@ -1,6 +1,6 @@
-ch.tutteli.atrium.translations.DescriptionAnyAssertion-TO_BE = est
-ch.tutteli.atrium.translations.DescriptionAnyAssertion-NOT_TO_BE = n'est pas
-ch.tutteli.atrium.translations.DescriptionAnyAssertion-IS_NOT_SAME = n'est pas la m�me instance que
+ch.tutteli.atrium.translations.DescriptionBasic-TO_BE = est
+ch.tutteli.atrium.translations.DescriptionBasic-NOT_TO_BE = n'est pas
+ch.tutteli.atrium.translations.DescriptionAnyAssertion-IS_NOT_SAME = n'est pas la m\u00EAme instance que
 ch.tutteli.atrium.specs.AssertionVerb-ASSERT = il applique que
-ch.tutteli.atrium.specs.reporting.translating.TranslationSupplierSpec$TestTranslatable-DATE_KNOWN=%tD �tait %<tA!!
-ch.tutteli.atrium.specs.reporting.translating.TranslationSupplierSpec$TestTranslatable-PLACEHOLDER=Caract�re de remplacement %s
+ch.tutteli.atrium.specs.reporting.translating.TranslationSupplierSpec$TestTranslatable-DATE_KNOWN=%tD �tait %<tA!!
```
robstoll: That does not look right, but specs are green, so I guess its OK :laughing:

jGleitz: Does not look right at all!
I had encoding issues with même, which I could only solve by replacing it with an escaped unicode character. I am curios why there is no test failure for this one…

robstoll: The problem is actually intellij, it saves properties file as UTF-8 where resource-bundle properties should be saved as ISO-8859-1

robstoll: I found the problem, spek1 has the problem that it does not execute two branches with the same name. Could you please fix this and the next line and I will point out in another comment where the fix needs to apply

jGleitz: Done. But please see comment below.

jGleitz:
Note: `Atrium_it.properties` and `TranslatorIntSpec$TestTranslatable_it.properties` are still not used anywhere (changing them does not make any test fail). Is this another case of a test case not being executed?

robstoll:
Thanks for the PR, I'll look into the issue with TranslatorIntSpec$TestTranslatable_it.properties and look for a fix for PropertiesPerEntityAndLocaleTranslationSupplier. 
As I plan to move away from ResourceBundle latest with 1.0.0 anyway (I'll probably switch to gettext) I am not even sure if it is worth to fix it.

PR #140, opened 2019-09-07
Elaborate path assertions (existence, type, permissions)

### LINKED ISSUE #111: shortcut for Path.isDirectory and Path.isRegularFile
*Platform* (JVM, JS and/or Android): jdk8

## Code related feature
```
expect(Paths.get("xyz")).isDirectory()
expect(Paths.get("xyz")).isRegularFile()
//instead of
expect(Paths.get("xyz")).feature(Files::isDirectory) { toBe(true) }
expect(Paths.get("xyz")).feature(Files::isRegularFile) { toBe(true) }
```

*domain*
- extend PathAssertions with `isDirectory` and `isRegularFile` (as for `exists`)
- modify PathAssertionsBuilder, delegate to pathAssertions (as for `exists`)
- delegate implementation to robstoll-lib in PathAssertionsImpl (as for `exists`)

*lib*
- implement _isDirectory and _isRegularFile in pathAssertions.kt in robstoll-lib (I have already setup a dependency to niok which provides Path.isDirectory and Path.isRegularFile) => see _exists as example, better use a descriptive Assertion instead of creating a feature assertion, less noisy, more readable, I would use `DescriptionBasic.IS` as description

*api*
- add  `isDirectory` and `isRegularFile` (`exists` as guideline)  and `@since 0.9.0` (adopt to current version) to the KDOC
- extend PathAssertionsSpec in specs-common

## Your first contribution?
  ([Invite yourself](http://slack.kotlinlang.org/) in case you do not have an account yet).

### LINKED ISSUE #112: shortcut for Path. isReadable and isWritable
*Platform* (JVM, JS and/or Android): jdk8

## Code related feature
```
expect(Paths.get("xyz")).isReadable()
expect(Paths.get("xyz")).isWritable()
//instead of
expect(Paths.get("xyz")).feature(Files::isReadable) { toBe(true) }
expect(Paths.get("xyz")).feature(Files::isWritable) { toBe(true) }
```

*domain*
- extend PathAssertions with `isReadable` and `isWritable` (as for `exists`)
- modify PathAssertionsBuilder, delegate to pathAssertions (as for `exists`)
- delegate implementation to robstoll-lib in PathAssertionsImpl (as for `exists`)

*lib*
- implement _isReadable and _isWritable in pathAssertions.kt in robstoll-lib (I have already setup a dependency to niok which provides Path.isReadable and Path.isWritable) => see _exists as example, better use a descriptive Assertion instead of creating a feature assertion, less noisy, more readable, I would use `DescriptionBasic.IS` as description.

*api*
- add  `isReadable` and `isWritable` (`exists` as guideline)  and `@since 0.9.0` (adopt to current version) to the KDOC
- extend PathAssertionsSpec in specs-common

## Your first contribution?
  ([Invite yourself](http://slack.kotlinlang.org/) in case you do not have an account yet).

jGleitz:
Implements assertions for a path
 * being a directory
 * being a file
 * being readable
 * being writable

also adapts the existing existence assertions.

The implementation tries to give helpful hints about what actually happened in the file system, so users do not have to check themselves (checking might not even be possible). The implementation realises these hints
 * when the subject path does not exist, show the last existing parent directory
 * when one part of the subject’s path is a file, show
 * when one part of the subject’s path cannot be accessed, show this part and its permissions
 * show all symbolic links that have been involved in resolving the path.

Please note that I was often not sure which type of `Assertion` to use. I appreciate input on this.

Examples:

```
 assert: /tmp/spek10214843601528508958/test        (sun.nio.fs.UnixPath <1744723183>)
◆ to be: a file
  ⚬ was: a directory
```
```
assert: /tmp/spek10474502568126922701/__linkTo_exist        (sun.nio.fs.UnixPath <1409474925>)
◆ to: exist
    » followed the symbolic link /tmp/spek10474502568126922701/__linkTo_exist to /tmp/spek10474502568126922701/startFile/i/dont/exist
    » failure at parent path: /tmp/spek10474502568126922701/startFile        (sun.nio.fs.UnixPath <1011844624>)
      » was a file instead of a directory
```
```
assert: /tmp/spek12943295927624369491/a        (sun.nio.fs.UnixPath <1986769819>)
◆ to be: readable
      ❗❗ found a symbolic link loop: /tmp/spek12943295927624369491/a -> /tmp/spek12943295927624369491/b -> /tmp/spek12943295927624369491/a
    » access threw FileSystemException: /tmp/spek12943295927624369491/a: Zu viele Ebenen aus symbolischen Links or unable to access attributes of symbolic link
```
```
assert: /tmp/spek12594952494284002728/startDir/i/dont/exist        (sun.nio.fs.UnixPath <327819564>)
◆ to be: a directory
    » failure at parent path: /tmp/spek12594952494284002728/startDir        (sun.nio.fs.UnixPath <165175844>)
      » access was denied
      » the POSIX permissions are u=w g= o=
```
```
assert: /tmp/spek14050191638722983707/__linkTo_not-writable        (sun.nio.fs.UnixPath <1825779019>)
◆ to be: writable
    » followed the symbolic link /tmp/spek14050191638722983707/__linkTo_not-writable to /tmp/spek14050191638722983707/not-writable
    » a directory exists at this location, but it is not writable
    » the POSIX permissions are u=rx g= o=x
```
```
assert: /tmp/spek11308856912373601693/startDir/i/dont/exist        (sun.nio.fs.UnixPath <125066784>)
◆ to be: a directory
    » no file system entry exists at this location
    » the closest existing parent directory is /tmp/spek11308856912373601693/startDir
```

closes #111
closes #112

----


jGleitz:
@robstoll GitHub Actions are not configured for pull requests, so this PR is not being built on Windows. Can you fix this?

(Testing on Windows would be especially important for this PR).

domain/builders/atrium-domain-builders-common/src/main/kotlin/ch/tutteli/atrium/domain/builders/creating/anyAssertionsBuilders.kt
```diff
@@ -20,7 +20,7 @@ import kotlin.reflect.KClass
  * In detail, it implements [AnyAssertions] by delegating to [anyAssertions]
  * which in turn delegates to the implementation via [loadSingleService].
  */
-object AnyAssertionsBuilder : AnyAssertions {
+object AnyAssertionsBuilder : AnyAssertions by anyAssertions {
```
robstoll: please revert

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/build.gradle
```diff
@@ -8,6 +8,8 @@ ext.jacoco_additional = [
     prefixedProject('core-robstoll-lib-jvm'),
 ]
 
+spek.version = "$spek2_version"
```
robstoll: is this really required? I might be wrong but I thought spek2 is already available

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/build.gradle
```diff
@@ -20,7 +22,9 @@ dependencies {
     //TODO remove with 1.0.0
     implementation prefixedProject('api-cc-en_GB-jvm')
 
+    testImplementation spekExtensions()
     testImplementation prefixedProject('verbs-internal-jvm')
+    testImplementation prefixedProject('translations-en_GB-jvm')
```
robstoll: I guess this is not required

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
```diff
@@ -0,0 +1,101 @@
...
+ * Adds explanatory hints for all involved symbolic links to [hintList].
+ */
+fun addAllLevelResolvedSymlinkHints(
+    absolutePath: Path,
+    hintList: Deque<Assertion>
+) = addAllLevelResolvedSymlinkHints(absolutePath, hintList, LinkedList(), absolutePath)
+
+private fun addAllLevelResolvedSymlinkHints(
+    absolutePath: Path,
+    hintList: Deque<Assertion>,
```
robstoll: why not a Hashset?

domain/builders/atrium-domain-builders-common/src/main/kotlin/ch/tutteli/atrium/domain/builders/creating/anyAssertionsBuilders.kt
jGleitz: Oh, that was an experiment. Should not have been commited. Done.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/build.gradle
jGleitz: Spek2 is not available and compilation fails without this line.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/build.gradle
jGleitz: The translations are not available and compilation fails without this line.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
jGleitz: For `hintList`: We never call `contains` or the like on it, using a `HashSet` would not bring any advantage, would it? (It would need to be `LinkedHashSet` because the hints are ordered)
For `visitedList`: We call `indexOf` on it, so i would not expect any gain from using a set, either.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/build.gradle
robstoll: The actual problem is that this project depends on `verbs-internal-jvm` instead of `specs-jvm`. Could you please change that (was a failure on my part)? spekExtensions and translations are then not required anymore

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
robstoll: Sorry, making reviews on mobile, I intended to target visitedList not hintList
I think Set is more suitable for two reasons. 1) you don't want duplicate paths in the visitedList (I guess) and Set re-states this intention, this is the main reason IMO. 2) Set.contains is faster than LinkedList.indexOf as indexOf needs to travers the whole LinkedList if it isn't in there, having O(n) complexity in contrast to O(1) - however, the list will be small and only be used in case of an error where performance isn't key anyway.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
```diff
@@ -0,0 +1,101 @@
...
+import ch.tutteli.atrium.reporting.RawString
+import ch.tutteli.atrium.translations.DescriptionPathAssertion.FAILURE_DUE_TO_LINK_LOOP
+import ch.tutteli.atrium.translations.DescriptionPathAssertion.HINT_FOLLOWED_SYMBOLIC_LINK
+import java.io.IOException
+import java.nio.file.Files
+import java.nio.file.Path
+import java.util.*
+
+inline fun explainForResolvedLink(path: Path, block: (realPath: Path) -> Assertion): Assertion {
+    val hintList = LinkedList<Assertion>()
```
robstoll: Just a matter of taste, you can keep it this way if you want. I would move it into addAllLevelResolvedSymlinkHints

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
```diff
@@ -0,0 +1,101 @@
...
+    val realPath = addAllLevelResolvedSymlinkHints(
+        path.toAbsolutePath(),
```
robstoll: I would also move the `toAbsolutePath` call into addAllLevelResolvedSymlinkHints the one which currently expects two arguments

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
```diff
@@ -0,0 +1,101 @@
...
+        hintList
+    )
+    val blockResult = block(realPath)
+    return if (hintList.isNotEmpty()) {
+        when (blockResult) {
+            is AssertionGroup -> hintList.addAll(blockResult.assertions)
```
robstoll: That seems smelly to me. An AssertionGroup has more than just the assertions. Why not just adding the group?

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
```diff
@@ -0,0 +1,101 @@
...
+import ch.tutteli.atrium.domain.builders.ExpectImpl
+import ch.tutteli.atrium.reporting.RawString
+import ch.tutteli.atrium.translations.DescriptionPathAssertion.FAILURE_DUE_TO_LINK_LOOP
+import ch.tutteli.atrium.translations.DescriptionPathAssertion.HINT_FOLLOWED_SYMBOLIC_LINK
+import java.io.IOException
+import java.nio.file.Files
+import java.nio.file.Path
+import java.util.*
+
+inline fun explainForResolvedLink(path: Path, block: (realPath: Path) -> Assertion): Assertion {
```
robstoll: maybe you can find a more descriptive name than block?

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
jGleitz: The key part here is that we want to call `indexOf`, and not `contains`! This is because `indexOf` gives us the number of items we can drop from `hintList` (because we do not want “followed symbolic link…” hints for links that a part of circle)

Regarding duplicates: We actually add a duplicated: the start of a loop. But this can be solved differently.

So would you still prefer a `LinkedHashSet`, even though the complexity will remain O(n)?

robstoll: Sorry, I am probably too tired now, I watched out for a use of visitedIndex other than `!= -1` and did not see it

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
```diff
@@ -0,0 +1,101 @@
...
+ */
+private fun addOneStepResolvedSymlinkHint(absolutePath: Path, resultList: Deque<Assertion>): Path? {
+    try {
+        val nextPath = Files.readSymbolicLink(absolutePath)
+        resultList.add(
+            ExpectImpl.builder.explanatory
+                .withExplanation(HINT_FOLLOWED_SYMBOLIC_LINK, absolutePath, nextPath)
+                .build()
+        )
+        return nextPath
```
robstoll: I would remove both `return` and turn this method into a method with expression body. But up to you

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+    return currentPath
+}
+
+/**
+ * If [absolutePath] is surely a symlink, adds an explanatory hint to [resultList] and returns the link target.
+ * Return `null` and does not modify [resultList] otherwise.
+ */
+private fun addOneStepResolvedSymlinkHint(absolutePath: Path, resultList: Deque<Assertion>): Path? {
```
robstoll: I would rename resultList to hintList (or the other way round, rename hintList to resultList in other places)

jGleitz:
Note: making permission checks compatible to Windows will require significant changes to the PR.

robstoll:
I'll continue with my review tomorrow or next week. Regarding compatibility with Windows. I guess this is necessary, we cannot put something in which breaks on windows. Do you want to figure out how this can be done so that it works on windows as well?

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
```diff
@@ -0,0 +1,101 @@
...
+        return nextPath
+    } catch (e: IOException) {
+        // either this is not a link, or we cannot check it. The best we can do is assume it is not a link.
+        return null
+    }
+}
+
+private fun hintForLinkLoop(loop: List<Path>, startIndex: Int): Assertion {
+    val loopRepresentation = loop.subList(startIndex, loop.size).joinToString(" -> ")
+    return ExpectImpl.builder.explanatoryGroup.withWarningType.withAssertion(
```
robstoll: you could also use `withExplanatoryAssertion` instead of `withAssertion` and use a placeholder in `FAILURE_DUE_TO_LINK_LOOP` to pass the loopRepresentation

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/pathAssertions.kt
```diff
@@ -1,18 +1,296 @@
...
+import java.nio.file.AccessDeniedException
+import java.nio.file.NoSuchFileException
 import java.nio.file.Path
+import java.nio.file.attribute.BasicFileAttributes
+import java.nio.file.attribute.PosixFilePermission
+import java.nio.file.attribute.PosixFilePermission.*
 
 fun <T : Path> _exists(assertionContainer: Expect<T>): Assertion =
-    ExpectImpl.builder.createDescriptive(assertionContainer, TO, RawString.create(EXIST)) { it.exists }
+    ExpectImpl.changeSubject.unreported(assertionContainer) {
```
robstoll: you could move 
```
    ExpectImpl.changeSubject.unreported(assertionContainer) {
        it.runCatchingIo { readAttributes<BasicFileAttributes>() }
    }.let { fileAttributesAssertionContainer ->
```
to an own function an reuse it here and in _existsNot

jGleitz:
I will work on Windows compatibility. However, I cannot predict how long it will take me.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/build.gradle
jGleitz: Done.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/pathAssertions.kt
```diff
@@ -1,18 +1,296 @@
...
+                    is Success -> describeWas(result.value.fileType)
+                    is Failure -> hintForIoException(realPath, result.exception)
+                }
+            }
+        }
+        .withDescriptionAndRepresentation(TO_BE, RawString.create(typeName))
+        .build()
+}
+
+fun hintForActualPosixPermissions(filePermissions: Set<PosixFilePermission>) =
```
robstoll: could be made `private` as well no?

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/build.gradle
jGleitz: Done.

robstoll:
> I will work on Windows compatibility. However, I cannot predict how long it will take me.

That's fine, we are not in a rush

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
jGleitz: The list creation cannot be moved there because the function does not return a reference to the list. It just adds elements to it (hence the name).

I do not like having functions modify passed parameters. But returning a `Pair` is not much better, I think.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
jGleitz: Done

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
jGleitz: Done.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/build.gradle
jGleitz: Done.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/build.gradle
jGleitz: Done.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
jGleitz: Done.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/pathAssertions.kt
jGleitz: Done.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/pathAssertions.kt
jGleitz: Done.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
jGleitz: I am trying to solve the following problem here:

The `block` will sometimes create an `ExplanatoryAssertionGroup`. If I include it directly, it will be indented, which I do not want. This is why I unwrap the `AssertionGroup` here.

You you have a better solution for this? I understand that the type check for `AssertionGroup` is not a good solution.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
jGleitz: Good idea! Done.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/test/kotlin/ch/tutteli/atrium/assertions/filesystem/SymbolicLinkResolvingSpec.kt
```diff
@@ -0,0 +1,182 @@
...
+                    get(1).describesLink(grandparentlink, grandparent)
+                    get(2).describesLink(innerLink, target)
+                    get(3).isSameAs(testAssertion)
+                }
+            }
+        }
+
+        it("adds an explanation for link loops") {
+            val a = tempFolder.tmpDir.resolve("linkA")
+            val b = tempFolder.newSymbolicLink("linkB", a)
```
robstoll: I was thinking that adding a test case with a relative symlink might be a good idea as well. e.g. 
```
val c = tempFolder.tmpDir.resolve("nonExisting").resolve("..").resolve("linkA")
val b = tempFolder.newSymbolicLink("linkB", c)
```

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/test/kotlin/ch/tutteli/atrium/assertions/filesystem/SymbolicLinkResolvingSpec.kt
```diff
@@ -0,0 +1,182 @@
...
+import ch.tutteli.atrium.domain.builders.ExpectImpl
+import ch.tutteli.atrium.domain.robstoll.lib.creating.filesystem.explainForResolvedLink
+import ch.tutteli.atrium.reporting.RawString
+import ch.tutteli.atrium.reporting.translating.TranslatableWithArgs
+import ch.tutteli.atrium.reporting.translating.Untranslatable
+import ch.tutteli.atrium.translations.DescriptionPathAssertion.FAILURE_DUE_TO_LINK_LOOP
+import ch.tutteli.atrium.translations.DescriptionPathAssertion.HINT_FOLLOWED_SYMBOLIC_LINK
+import ch.tutteli.spek.extensions.TempFolder
+import com.nhaarman.mockitokotlin2.any
+import com.nhaarman.mockitokotlin2.doReturn
```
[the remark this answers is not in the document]
robstoll: let's keep this we can switch to mockk later on

jGleitz:
I give up. On Windows, it seems to be impossible to tell from Java the difference between an empty folder and one that we may not read. I tried every trick and examination I could think of, but those two cases always look the same.

Windows programs _do_ know the difference, so there is a way on operating system level. But I have found no way from Java.

If nobody has another idea on how to differentiate these cases, I will stop trying and just add a comment on the assertions. This would be a shame because differentiating manually between missing files and ones that cannot be accessed can be difficult. It would have been very useful to report on that in atrium.

jGleitz:
Thinks that do _not_ work to differentiate between an empty and a not-readable folder:
* checking for the Exception type on child access (`NoSuchFileException` vs. `AccessDeniedException`)
* checking `isReadable` (or the file counterpart `canRead`)
* checking `isExecutable`
* listing the files using `Files.createDirectoryStream`, `File.list` or `File.listFiles`
* checking whether `..` exists in the directory
* examining the directory’s size
* examining the ACL manually, because I do not know which groups the current user is in

robstoll:
I'll take a look next week. @jGleitz is it OK for you if we merge the PR into a separate feature branch. I would continue from there (please push new changes beforehand)

jGleitz:
> is it OK for you if we merge the PR into a separate feature branch

absolutely. I will address the other points we discussed in this PR, push them on to a branch, tidy up the commits and then we can merge it into a feature branch.

jGleitz:
> I'll take a look next week.

Thank you!

robstoll:
I am actually unsure what you try to achieve. 
> This would be a shame because differentiating manually between missing files and ones that cannot be accessed can be difficult.

I guess you would like to give the hint that a file exists in case it is not readable and if it does not exist then check if the parent directory exists and is readable etc. Right? 
Following a scenario:
C:\tmp <- everyone has access
C:\tmp\noAccess <- only admin has access (user cannot even traverse/execute)
C:\tmp\noAccess\stillNoAccess\userFolder\test.txt <- java has access

Following a test program:
```
fun main(args: Array<String>) {
   var path: Path? = Paths.get("C:/tmp/noAccess/stillNoAccess/userFolder/test.txt")
   while(path != null){
       print(path)
       path = path.parent
   }
}
private fun print(path: Path) {
   println(path.absolutePathAsString)
   println("r: " + path.isReadable)
   println("e: " + path.exists)
   println("d: " + path.isDirectory)
   println("f: " + path.isRegularFile)
}
```
And the output for Windows 7
```
C:\tmp\noAccess\stillNoAccess\userFolder\test.txt
r: true     <-- works on Windows but not Ubuntu (see output below for without execution rights)    
e: true
d: false
f: true
C:\tmp\noAccess\stillNoAccess\userFolder
r: true
e: true
d: true
f: false
C:\tmp\noAccess\stillNoAccess
r: false       
e: false
d: false     <-- does not even know if it is a folder as it does not have access for noAccess
f: false
C:\tmp\noAccess
r: false      
e: false     <-- that's a surprise, if it is known to be a folder, it should be known that it exists
d: true      <-- this one works fortunately and could be used as fallback
f: false
C:\tmp
r: true
e: true
d: true
f: false
```
Giving the user the right to traverse the folder `noAccess` (but not reading it) gives different results:
```
C:\tmp\noAccess\stillNoAccess
r: false
e: false     <-- still false
d: true      <-- fallback works now though
f: false
```

Doing the same on ubuntu 19.04 with execution rights
```
/tmp/noAccess/stillNoAccess/userFolder/test.txt
r: true
e: true
e: false
d: false
f: true
/tmp/noAccess/stillNoAccess/userFolder
r: true
e: true
e: false
d: true
f: false
/tmp/noAccess/stillNoAccess
r: false
e: true       
e: false
d: true
f: false
/tmp/noAccess
r: false
e: true     <--- that's different than windows
e: false
d: true     <--- that's the same though
f: false
/tmp
r: true
e: true
e: false
d: true
f: false
```
Without execution rights:
```
/tmp/noAccess/stillNoAccess/userFolder/test.txt
r: false 
e: false     <--- out of luck, Windows is better in this regard
d: false
f: false
/tmp/noAccess/stillNoAccess/userFolder
r: false
e: false   
d: false   
f: false
/tmp/noAccess/stillNoAccess
r: false
e: false   <--- same picture as for Windows
d: false    
f: false
/tmp/noAccess
r: false
e: true
d: true
f: false
```

To conclude: As we can see, there are scenarios both on Windows and on Ubuntu (and other unix-based systems I guess) were we cannot tell the difference between not readable/not traversable/not existing.

I guess your implementation has to be changed as it makes sometimes assumptions which might not be true. However, I think your idea of providing diagnosis information is still very valuable. I would be more than happy if I get an error reporting which looks similar to:
```
expect: /tmp/noAccess/stillNoAccess/userFolder/test.txt  (sun.nio.fs.UnixPath <1744723183>)
◆ is: readable
  ⚬ we already checked for you:
   ✘ /tmp/noAccess/stillNoAccess/userFolder/test.txt exists
   ✘ /tmp/noAccess/stillNoAccess/userFolder exists
   ✘ /tmp/noAccess/stillNoAccess exists
   ✔ /tmp/noAccess exists
   ✘ /tmp/noAccess is readable
   ✔ /tmp/noAccess is a directory
```
This should still be possible no? (using the `isDirectory` fallback on Windows). Please let me know if this already brings you further, otherwise I dig deeper into it next week

jGleitz:
> Following a scenario:
> C:\tmp <- everyone has access
> C:\tmp\noAccess <- only admin has access (user cannot even traverse/execute)
> C:\tmp\noAccess\stillNoAccess\userFolder\test.txt <- java has access

Wait… this is possible? There can be a file I have access to even though I cannot access the parent? I did not know that. This is not possible on Linux, is it?

robstoll:
> This is not possible on Linux, is it?
See the examples above for ubuntu, it is possible. One cannot access test.txt if one does not have execution rights but otherwise the file can be accessed.

jGleitz:
Okay, it is _not_ possible on Linux:

> The directory can be accessed with cd; this is the only permission bit that in practice can be considered to be "inherited" from the ancestor directories, in fact if any folder in the path does not have the x bit set, the final file or folder cannot be accessed either, regardless of its permissions; see path_resolution(7) for more information. 

https://wiki.archlinux.org/index.php/File_permissions_and_attributes

I already feared I lacked fundamental operating system knowledge :sweat: 

Is it possible on Windows? I did a quick test and was not able to access a file in a folder I had no permission to, even if I explicitly disabled inheritance and gave myself permissions on the file.

jGleitz:
> One cannot access test.txt if one does not have execution rights but otherwise the file can be accessed.

Yes, okay, I knew that. But this does not break any assumption I have in the code.

robstoll:
> I did a quick test and was not able to access a file in a folder I had no permission to

Maybe they improved things, it worked on win7, one just needs to know the exact path. Traversing the directory structure is not possible but accessing `userFolder` directly works

robstoll:
Does the `isDirectory` fallback help you? to achieve your goals?

jGleitz:
> Does the isDirectory fallback help you? to achieve your goals?

I do not think so, but maybe I also got something wrong.

In your example, I would like to know whether I am permitted to know whether `C:\tmp\noAccess\stillNoAccess` exists. And even with your analysis, I would not know how to do this. I hence can not discriminate between “`C:\tmp\noAccess\stillNoAccess` does surely not exist” and “We are not permitted to read directory entries of `C:\tmp\noAccess`”

However, this discussion might be obsolete, anyway. On Unix, it is clear that if we are not allowed to traverse `C:\tmp\noAccess`, that this folder is the root error and the user has to fix it in any case. On Windows, you claim that just because we cannot traverse `C:\tmp\noAccess` does not mean in principle that we cannot access `C:\tmp\noAccess\stillNoAccess\userFolder\test.txt`. So on Windows, unlike on Unix, it would not be correct to show the user only a hint for missing privileges on `C:\tmp\noAccess`, even if could find that out reliably.

jGleitz:
> I would be more than happy if I get an error reporting which looks similar to:

I like that your proposed reporting is very explicit, but I misslike that it is very verbose. In my experience, it is common to have relatively long paths when testing. So your output would span tens of lines.

The question remains: What do we do after an `NoSuchFileException` on Windows? I see three possibilities:

 a) Show a message that reads “file was not found or access to it was denied” (downside: the message is measleading on UNIX and I do not know whether we can reliably detect whether we have the permission problem. Windows supports different file systems)
 b) Show diagnostics for every path component, even though this might get verbose (even with only one line per component!)
 c) Magically figure out how Windows file permissions work :).

robstoll:
> However, this discussion might be obsolete, anyway.

Good, that was my conclusion as well :) IMO we can neglect the special case of Windows, let's go ahead and ignore that. In most cases the hint will be correct. 

> I do not think so, but maybe I also got something wrong.

I don't know how you intend to check if a folder exists on Windows if not using `isDirectory` as  `exists` seems to be buggy:
```
C:\tmp\noAccess
r: false      
e: false     <-- that's a surprise, if it is known to be a folder, it should be known that it exists
d: true      <-- this one works fortunately and could be used as fallback
```

jGleitz:
> I don't know how you intend to check if a folder exists on Windows if not using `isDirectory`

The code always used `isDirectory`, which is why I never knew that `exists` is buggy: https://github.com/robstoll/atrium/pull/140/files#diff-5a4f9880b24f11ca3d58e1b7eeacd51aR222

jGleitz:
> IMO we can neglect the special case of Windows, let's go ahead and ignore that.

So you propose to leave the code as it is in this PR and report „no such file“ for cases where access restrictions are the actual error?

Please note that this also affects the `existsNot` matcher: It would succeed for an existing file we must not access (On Unix, it would not succeed because we realise that we do not have access). The matcher would be in accordance with the behaviour of Files.existsNot, though…

robstoll:
Sorry, this comment slipped through my fingers: https://github.com/robstoll/atrium/pull/140#issuecomment-531485114

I only meant that we can ignore cases like `C:\tmp\noAccess\stillNoAccess\userFolder\test.txt` which behave differently on Linux than on Windows.

> I like that your proposed reporting is very explicit, but I misslike that it is very verbose. In my experience, it is common to have relatively long paths when testing. So your output would span tens of lines.

perfectly fine if the hint is shorter and something along the line of.
```
» closest existing path: /tmp
  + is directory: true  
  + is readable: true
```
Have you thought about providing examples in ReadmeSpec? I think this would make it way easier to talk about. Especially as you have already though about different use cases I don't even have in mind probably. 

I prefer to be misleading on Windows systems than on linux. But we could also add the addition `or access to it was denied` only for windows systems.

jGleitz:
> Have you thought about providing examples in ReadmeSpec? I think this would make it way easier to talk about. Especially as you have already though about different use cases I don't even have in mind probably.

Will do. I wanted to fix bugs first. But now I will leave things as they are for know and incorporate all the changes you suggested, including adding examples to the README

> ```
>» closest existing path: /tmp
>  + is directory: true  
>  + is readable: true

that looks promising! Let’s discuss further as soon as I have added the examples.

jGleitz:
The builds now fail because `:atrium-specs-android` fails to resolve `niok` at compilation. @robstoll, can you have a look at that?

misc/specs/atrium-specs-jvm/build.gradle
```diff
@@ -32,4 +32,5 @@ dependencies {
     //TODO #116 remove once there aren't any spek1 specs anymore
     api "org.jetbrains.spek:spek-api:$spek_version", excludeKotlin
     api mockito(), excludeKotlin
```
robstoll: add the same to specs-android, should fix the build

misc/specs/atrium-specs-jvm/build.gradle
```diff
@@ -32,4 +32,5 @@ dependencies {
...
+    api niok(), excludeKotlin
```
robstoll: add the same to specs-android, should fix the build

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
jGleitz: Done.

misc/specs/atrium-specs-jvm/build.gradle
jGleitz: Done. It fixed the build.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
```diff
@@ -6,25 +6,25 @@ import ch.tutteli.atrium.assertions.builders.withExplanatoryAssertion
...
 import ch.tutteli.atrium.translations.DescriptionPathAssertion.FAILURE_DUE_TO_LINK_LOOP
 import ch.tutteli.atrium.translations.DescriptionPathAssertion.HINT_FOLLOWED_SYMBOLIC_LINK
+import ch.tutteli.niok.followSymbolicLink
 import java.io.IOException
-import java.nio.file.Files
 import java.nio.file.Path
 import java.util.*
 
-inline fun explainForResolvedLink(path: Path, block: (realPath: Path) -> Assertion): Assertion {
+inline fun explainForResolvedLink(path: Path, assertionCreator: (realPath: Path) -> Assertion): Assertion {
```
robstoll: the name is better but usually an assertionCreator has the type `Expect<T>.() -> Unit`. I suggest we use it only for this. How about assertionProvider/Supplier?

Daydreamzz:
N1

Daydreamzz:
https://tse1.mm.bing.net/th?id=OGC.cd0ad3250f0840b643adfafbf0bb4475&pid=Api&rurl=https%3a%2f%2fmedia.tenor.co%2fimages%2fcd0ad3250f0840b643adfafbf0bb4475%2fraw&ehk=ye1BRACX3zWshk5cN62qbw

robstoll:
@jGleitz let me know once you have pushed everything you wanted and I'll take over

jGleitz:
When merging with `master`, I reverted the changes made in #206 for the tests of `#exists` and `#existsNot`. I think merging those two is more awkward with my new test structure than having them separated is.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/test/kotlin/ch/tutteli/atrium/assertions/filesystem/SymbolicLinkResolvingSpec.kt
jGleitz: Done.

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
```diff
@@ -1,93 +1,512 @@
...
-        context("non existing") {
-            it("throws an AssertionError") {
+    fun Suite.it(description: String, skip: Skip = Skip.No, timeout: Long = delegate.defaultTimeout) =
+        SymlinkTestBuilder(tempFolder) { prefix, body -> it(prefix + description, skip, timeout, body) }
+
+    fun Suite.itPrintParentAccessDeniedDetails(block: (Path) -> Unit) {
+        it("prints parent permission errors") withAndWithoutSymlink { maybeLink ->
+            val start = tempFolder.newFolder("startDir")
+            val doesNotExist = maybeLink.create(start.resolve("i").resolve("dont").resolve("exist"))
+            val parentFailureMessage = String.format(FAILURE_DUE_TO_PARENT.getDefault(), start)
```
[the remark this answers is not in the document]
jGleitz: Done.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
```diff
@@ -0,0 +1,93 @@
...
+    } else {
+        blockResult
+    }
+}
+
+/**
+ * Resolves the provided [path] and returns the resolved target (if resolving is possible).
+ * Adds explanatory hints for all involved symbolic links to [hintList].
+ */
+fun addAllLevelResolvedSymlinkHints(path: Path, hintList: Deque<Assertion>): Path {
```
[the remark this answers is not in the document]
jGleitz: No, because it is called from a public inline function.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/test/kotlin/ch/tutteli/atrium/assertions/filesystem/SymbolicLinkResolvingSpec.kt
```diff
@@ -0,0 +1,182 @@
...
+                expect(resultAssertion).isSameAs(testAssertion)
+            }
+
+            it("adds an explanation for one symbolic link") {
+                val target = tempFolder.tmpDir.resolve("notExisting")
+                val link = tempFolder.newSymbolicLink("link", target)
+
+                val resultAssertion = explainForResolvedLink(link, expectListener)
+                expect(resultAssertion).isA<AssertionGroup>().feature({ p(it::assertions) }) {
+                    get(0).describesLink(link, target)
```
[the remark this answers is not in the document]
jGleitz: Much better. Thanks!

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/pathAssertions.kt
```diff
@@ -1,18 +1,343 @@
...
+    get() = when {
+        isRegularFile -> A_FILE
+        isDirectory -> A_DIRECTORY
+        isSymbolicLink -> A_SYMBOLIC_LINK
+        else -> A_UNKNOWN_FILE_TYPE
+    }
+
+private fun formatPosixPermissions(filePermissions: Set<PosixFilePermission>): StringBuilder {
+    val permissionString = StringBuilder(3 * 5 + 2)
+    permissionString
```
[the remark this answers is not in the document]
jGleitz: This _is_ the order `ls -la` uses: owner, group and others. You might be irritated because of the different wording in Java and Linux:

Linux | Java
---|---
user | owner
group | group
others | others

The prefixes are the ones `chmod` accepts:

> A  combination of the letters `ugoa` controls which users' access to the file will be changed: the user who owns it (`u`), other users in the file's group (`g`), other users not in the  file's  group (`o`), or all users (`a`).

source: `man chmod`

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/pathAssertions.kt
```diff
@@ -1,18 +1,343 @@
...
+private inline fun <T : Path, R> changeSubjectToFileAttributes(
+    assertionContainer: Expect<T>,
+    block: (Expect<IoResult<BasicFileAttributes>>) -> R
+): R = ExpectImpl.changeSubject.unreported(assertionContainer) {
+    it.runCatchingIo { readAttributes<BasicFileAttributes>() }
+}.let(block)
+
+private fun hintForOwnersAndPermissions(path: Path): MutableList<Assertion> {
+    val hintList = LinkedList<Assertion>()
+    // TODO: use niok once https://github.com/robstoll/niok/issues/11 is fixed
```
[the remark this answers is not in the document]
jGleitz: Done.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
```diff
@@ -0,0 +1,93 @@
...
+    }
+    return currentPath
+}
+
+/**
+ * If [absolutePath] is surely a symlink, adds an explanatory hint to [hintList] and returns the link target.
+ * Return `null` and does not modify [hintList] otherwise.
+ */
+private fun addOneStepResolvedSymlinkHint(absolutePath: Path, hintList: Deque<Assertion>) = try {
+    val nextPath = Files.readSymbolicLink(absolutePath)
```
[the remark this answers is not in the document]
jGleitz: Done.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/pathAssertions.kt
```diff
@@ -1,18 +1,343 @@
...
+            when (explanation) {
+                is AssertionGroup -> explanation
+                else -> ExpectImpl.builder.explanatoryGroup.withDefaultType
+                    .withAssertion(explanation)
+                    .build()
+            }
+        ).build()
+
+private fun hintForOtherIoException(exception: IOException) =
+    ExpectImpl.builder.explanatory
```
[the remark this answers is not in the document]
jGleitz: Thanks for the tip! I have created a constant `IO_EXCEPTION_STACK_TRACE_LENGTH` for the maximum reported stack trace. I dislike this approach because it is essentially hardcoded, but I also do not see a nice way to expose this to the API to be changeable. Thoughts?

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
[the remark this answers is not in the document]
robstoll: then I suggest you make it `@PublishedApi internal`

core/api/atrium-core-api-common/src/main/kotlin/ch/tutteli/atrium/assertions/builders/descriptiveWithFailureHint.kt
```diff
@@ -118,6 +118,15 @@ interface DescriptiveAssertionWithFailureHint {
         fun <T> showOnlyIfSubjectDefined(subjectProvider: SubjectProvider<T>): Descriptive.DescriptionOption<FinalStep> =
             showOnlyIf { subjectProvider.maybeSubject.isDefined() }
 
+        /**
+         * Defines that the failure hint shall be shown if the subject is defined and the given [predicate] holds for it
+         */
+        fun <T> showBasedOnDefinedSubjectOnlyIf(
```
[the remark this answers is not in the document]
jGleitz: I am determined to get this PR through until Sunday latest. I can, however, still make a separate PR for the library changes if you like.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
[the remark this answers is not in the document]
jGleitz: Oh, right. Done.

core/api/atrium-core-api-common/src/main/kotlin/ch/tutteli/atrium/assertions/builders/descriptiveWithFailureHint.kt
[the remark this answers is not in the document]
robstoll: As you wish really, was only a secondary object.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/filesystem/symbolicLinkResolving.kt
jGleitz: Done.

jGleitz:
I have now added three examples to the README. I feel like more examples are overkill. I am happy to expand or improve the README!

@robstoll: How shall we proceed on this PR?

robstoll:
I did not have a look at the PR in this stage yet but I am not sure what you mean with:

> How shall we proceed on this PR?

I intend to merge it (either directly into master or into a feature branch if I want to do more things) or are there open points I don't have in mind now?

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/pathAssertions.kt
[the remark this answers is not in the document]
robstoll: I suggest we address this in a different issue, I have created #234

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/pathAssertions.kt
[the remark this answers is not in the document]
robstoll: Apparently a misunderstanding on my part, thanks for the explanation 👍

samples/readme-examples/src/main/kotlin/readme/examples/ReadmeSpec.kt
```diff
@@ -1,23 +1,30 @@
 package readme.examples
 
 //snippet-import-start
+//snippet-import-end
```
robstoll: that was surely intellij messing things up, please revert.

samples/readme-examples/src/main/kotlin/readme/examples/ReadmeSpec.kt
```diff
@@ -107,9 +114,11 @@ class ReadmeSpec : Spek({
 
     //@formatter:off
     test("ex-property-methods-group") {
-        expect(myPerson) { // forms an assertion group block
+        expect(myPerson) {
+            // forms an assertion group block
```
robstoll: please revert

samples/readme-examples/src/main/kotlin/readme/examples/ReadmeSpec.kt
```diff
@@ -107,9 +114,11 @@ class ReadmeSpec : Spek({
...
 
-            feature({ f(it::firstName) }) { // forms an assertion group block
+            feature({ f(it::firstName) }) {
```
robstoll: please revert

samples/readme-examples/src/main/kotlin/readme/examples/ReadmeSpec.kt
```diff
@@ -470,13 +503,15 @@ object OwnPerson : Spek({
...
         expect(Person("Susanne", "Whitley", 43, listOf(Person("Petra", "Whitley", 12, listOf()))))
-            .children { // using the fun -> assertion group, ergo sub-assertions don't fail fast
+            .children {
+                // using the fun -> assertion group, ergo sub-assertions don't fail fast
                 none { feature { f(it::firstName) }.startsWith("Ro") }
                 all { feature { f(it::lastName) }.toBe("Whitley") }
             } // subject is still Person here
-            .apply { // only evaluated because the previous assertion group holds
+            .apply {
+                // only evaluated because the previous assertion group holds
```
robstoll: please revert changes

README.md
```diff
@@ -1324,6 +1327,63 @@ In case you should miss an assertion function, then please
...
+expect: /root/.ssh/config        (sun.nio.fs.UnixPath <1234789>)
+◆ to be: writable
+    » failure at parent path: /root        (sun.nio.fs.UnixPath <1234789>)
+      » access was denied
+      » the owner is root, the group is root
+      » the permissions are u=rwx g= o=
+```
+</ex-path-writable>
+
+Even in more complicated scenarios, atrium explains step by step what happened:
```
robstoll:
```suggestion
Even in more complicated scenarios, Atrium explains step by step what happened:
```

robstoll:
IMO we can merge the PR into master once you have reverted the auto-formatting issues in README. Note the `//@formatter:off` I have used, you could tell intellij to ignore them for formatting as well (I don't remember the configuration by heart but I am sure you will find it, otherwise let me know)

jGleitz:
I reverted the formatting changes.

jGleitz:
> IMO we can merge the PR into master

I am happy to merge this! However, in your earlier comments, you sounded like you would like to revise the failure messages before merging.

robstoll:
> I am happy to merge this! However, in your earlier comments, you sounded like you would like to revise the failure messages before merging.

oh... I forgot to mention that in my last comment, I really like the error message as shown in the README :heart_eyes:

robstoll:
@jGleitz as a follow-up task: can you please have a look at codecov why it dropped from 88% to 77%

robstoll:
@jGleitz hm... master fails now due to the merge, could you have a look?

jGleitz:
I will look into why the test fails.

jGleitz:
About code coverage: The main issue I see is that Windows-specific code is (of course) marked as not-covered. Can we somehow tell JaCoCo to ignore those parts?

jGleitz:
@robstoll I see no failure on `master` have you re-started the job?

This particular test case uses mockk and runs for ~3.3s on my machine. I cannot explain why mockk needs so much time. If we observe this more often, we could increase the timeout of that test case or see whether going back to mockito will help (I expect that it will).

robstoll:
I restarted it and I thought it failed again, that's why I wrote here but apparently it passed the second time. So yeah, let's hope it was a glitch and does not occur regularly.

robstoll:
About the coverage part, I think it makes sense to **not** ignore it. Instead we can record the coverage also for a Windows build => I have created a task for it #235

jGleitz:
Well yes, this is of course a much better solution!

Issue #129, opened 2019-08-27
elementsOf(Iterable) for contains.inAnyOrder.atXy


robstoll:
*Platform* (jvm, jdk8, js, android): all

## Code related feature
```
val otherIterable : List<Int>
expect(listOf(1, 2)).contains.inAnyOrder.atLeast(1).elementsOf(otherIterable)

//instead of

expect(listOf(1, 2)).contains.inAnyOrder.atLeast(1).elementsOf(otherIterable.first(), *otherIterable.drop(1).toTypedArray())
```

Following the things you need to do:

*api*
- add fun `elementsOf` in apis/fluent-en_GB/atrium-fluent_en_GB-common/../iterableContainsInAnyOrderCreators.kt (see other functions there)
  - it is enough if you delegate to `values`
  - `require` that the `Iterable.hasNext`
  - add `@since 0.9.0` to the KDOC
- extend IterableContainsInAnyOrderAtLeast1ValuesAssertionsSpec (add another include, similar to  BuilderSpec)

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime,
  we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](http://slack.kotlinlang.org/) in case you do not have an account yet).

Oyedotuncaleb:
Great

kellybrewer1967:
I don't have a computer. Just this phone

kellybrewer1967:
I need a 64 bit codes to download on phone for fx player.

sanatik:
I'll work on this

robstoll:
implemented by a0b539c0

PR #157, opened 2019-09-29
Add elementsOf function for contains.inAnyOrder

sanatik:
Discussed in #129 
Added `elementsOf` function for `contains.inAnyOrder`
Function allows you to search in `Iterable` entries that exists in expected `Iterable`

----


robstoll:
@sanatik thank you very much for the contribution :slightly_smiling_face: :+1:

sanatik:
@robstoll you are welcome!
I would love to pick something else, checking issues now :)
Let me know if you have some ideas

robstoll:
@sanatik how about: https://github.com/robstoll/atrium/issues/158 I just created it (kind of for you :wink:). It goes a bit deeper into Atrium but should still be fairly simple. I think that would be a good next step.

PR #159, opened 2019-10-01
Change Github Emoji to Unicode emoji

Frandall:
Replace emoji that was defined using github markup with unicode based emoji

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 Atrium is an open-source multiplatform assertion library for Kotlin with support for JVM, JS and Android.
 It is designed to support multiple [APIs](#api-styles), different error reporting styles and [Internationalization](#internationalization-1) (i18n). 
 The project was inspired by AssertJ at first but moved on and provides now more 
-flexibility, features and hints to its users (so to you :wink:).
+flexibility, features and hints to its users (so to you &#x1F609;).
 
 Atrium is designed to be extensible as well as configurable 
 and allows you to extend it with your own assertion functions, customise reporting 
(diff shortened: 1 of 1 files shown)


robstoll:
@Frandall First of all, thank you very much for your contribution. :slightly_smiling_face: :+1:
I need your opinion, I am no longer sure.... I am disappointed how my browser (firefox) renders ⚠ , it's not even yellow :cry: 
:warning: looks much better.  I am no longer sure if it is a good idea to change the icons to utf-8. The primary usage is github. What do you think?

Or maybe we could change these icons to something else. Following a suggestion, but maybe you come up with a better one:
- :warning:  => 😱
- :information_source: => 💬
- :interrobang: => 🤔

Frandall:
Well, from your issue, you concern about external sites not rendering properly. What external sites that you had found not rendering it correctly?

robstoll:
Sad enough I don't remember

Frandall:
If you want to change the icon, I think we need to find more suitable than 
⚠️ => 😱
ℹ️ => 💬
⁉️ => 🤔

robstoll:
ideas?

Frandall:
How about
⚠️ => ❗
ℹ️ => 👓
⁉️ => 💬

robstoll:
Yeah... go for it. 
Just one note about your change though. I would prefer if you could use utf-8 characters instead of escaping sequences. Or have you done this for a particular reason?

robstoll:
what browsers do you have in mind?

Frandall:
I mainly use chrome ^^

robstoll:
I mean what browser do you have in mind which do not support utf8 characters yet

Frandall:
Hmm, I think in this case all of them support utf8, because people who find this will have some degree of knowledge to use the latest of popular browser. I think I will use the utf-8 character.

robstoll:
LGTM, thank you very much for the discussion and contribution :slightly_smiling_face:

robstoll:
Fancy another one? Have a look at https://github.com/robstoll/atrium/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22 e.g. 160

Frandall:
I'll let you know in 3 days.

PR #161, opened 2019-10-02
Overload `CharSequence.containsRegex` to expect `Regex`

neelkamath:
*domain*

- [x] modify CharSequenceContainsAssertions, expect a `List<Regex>` instead of a `List<String>` for regex
- [x] adopt CharSequenceAssertionsBuilder, add the new method but keep the one which expects `List<String>` and delegate to the new one
- [x] adopt the signature of CharSequenceAssertionsImpl, keep only `List<Regex>`

*lib*

- [x] add an additional type parameter SC to CharSequenceContains.Searcher which represents the search criterion => use it for searchFor
  - [x] use Any as second type parameter in ch.tutteli.atrium.assertions.charsequence.contains.CharSequenceContains.Searcher
  - [x] use Regex for RegexSearcher and IgnoringCaseRegexSearcher
  - [x] use String for IndexSearcher and IgnoringCaseIndexSearcher
  - [x] adopt CharSequenceContainsAssertionCreator.search (use SC)
  - [x] adopt the signature of _containsRegex
  - [x] adopt createAssertionGroup and other functions in ch/tutteli/atrium/domain/robstoll/lib/creating/charsequence/contains/creators/creators.kt
this means i.a. that we need to map expected from Any to String with toString() for _containsValues and _containsValuesIgnoringCase
  - [x] adopt ch.tutteli.atrium.assertions.charsequence.contains.CharSequenceContains.Searcher (use Any as second type parameter)

*api*

- [x] provide an additional overload for containsRegex in charSequenceAssertions which expects Regex instead of string (see existing fun)
- [x] provide an additional overload for CharSequenceContains.CheckerOption<T, NoOpSearchBehaviour>.regex in charSequenceContainsCreators.kt which expects Regex
- [x] add a RegexSpec to ch.tutteli.atrium.api.fluent.en_GB.CharSequenceContainsRegexAssertionsSpec (similar to StringSpec)

I confirm that I have read the Contributor Agreements v1.0, agree to be bound on them and confirm that my contribution is compliant.


neelkamath:
@robstoll 

For the task **deprecate CharSequenceContainsAssertions.regexIgnoringCase and delegate to the new regex using `Regex(..., RegexOption.IGNORE_CASE)``**,

I updated the file `ch/tutteli/atrium/domain/robstoll/creating/charsequence/contains/creators/CharSequenceContainsAssertionsImpl.kt` to look like:

```kotlin
@Deprecated(
        "use regex(checkerOption, expected)",
        ReplaceWith(
            "regex(checkerOption, expected)",
            "ch.tutteli.atrium.domain.robstoll.creating.charsequence.contains.creators.CharSequenceContainsAssertionsImpl.regex"
        )
    )
    override fun <T : CharSequence> regexIgnoringCase(
        checkerOption: CharSequenceContains.CheckerOption<T, IgnoringCaseSearchBehaviour>,
        expected: List<String>
    ): AssertionGroup = regex(checkerOption, expected.map { Regex(it, RegexOption.IGNORE_CASE) })
```

The call to `regex()` gives me an error because `checkerOption` doesn't satisfy the interface.

I don't think I'm supposed to implement the interface myself like this:
```kotlin
@Deprecated(
        "use regex(checkerOption, expected)",
        ReplaceWith(
            "regex(checkerOption, expected)",
            "ch.tutteli.atrium.domain.robstoll.creating.charsequence.contains.creators.CharSequenceContainsAssertionsImpl.regex"
        )
    )
    override fun <T : CharSequence> regexIgnoringCase(
        checkerOption: CharSequenceContains.CheckerOption<T, IgnoringCaseSearchBehaviour>,
        expected: List<String>
    ): AssertionGroup = regex(
        object : CharSequenceContains.CheckerOption<T, IgnoringCaseSearchBehaviour> {
            override val checkers: List<CharSequenceContains.Checker>
                get() = TODO("not implemented")
            override val containsBuilder: CharSequenceContains.Builder<T, IgnoringCaseSearchBehaviour>
                get() = TODO("not implemented")
        },
        expected.map { Regex(it, RegexOption.IGNORE_CASE) }
    )
```

So what should I do?

robstoll:
Right, my fault. Skip this point. Wait with the *lib* section, I have to quickly review my description

robstoll:
@neelkamath note that I am not writing down every single step you need to do. E.g. you also need to adopt ch.tutteli.atrium.assertions.charsequence.contains.CharSequenceContains.Searcher (use Any as second type parameter). You can run `./gr build` in the root of the project to check if everything still compiles

neelkamath:
I'm stuck with what "adopt ..." means 😅

robstoll:
What I meant with adopt: change it in such a way that the compiler is happy again. Are you using intellij? It should complain, that Searcher requires two type parameters

neelkamath:
Am I overloading incorrectly?

For the code:
```kotlin
    fun <T : CharSequence> regex(
        checkerOption: CharSequenceContains.CheckerOption<T, NoOpSearchBehaviour>,
        expected: List<String>
    ): AssertionGroup

    fun <T : CharSequence> regex(
        checkerOption: CharSequenceContains.CheckerOption<T, NoOpSearchBehaviour>,
        expected: List<Regex>
    ): AssertionGroup
```

I get the error:
```
e: /Users/neelkamath/Documents/GitHub/atrium/domain/api/atrium-domain-api-common/src/main/kotlin/ch/tutteli/atrium/domain/creating/charsequence/contains/creators/CharSequenceContainsAssertions.kt: (44, 5): Platform declaration clash: The following declarations have the same JVM signature (regex(Lch/tutteli/atrium/domain/creating/charsequence/contains/CharSequenceContains$CheckerOption;Ljava/util/List;)Lch/tutteli/atrium/assertions/AssertionGroup;):
    fun <T : CharSequence> regex(checkerOption: CharSequenceContains.CheckerOption<T, NoOpSearchBehaviour>, expected: List<String>): AssertionGroup defined in ch.tutteli.atrium.domain.creating.charsequence.contains.creators.CharSequenceContainsAssertions
    fun <T : CharSequence> regex(checkerOption: CharSequenceContains.CheckerOption<T, NoOpSearchBehaviour>, expected: List<Regex>): AssertionGroup defined in ch.tutteli.atrium.domain.creating.charsequence.contains.creators.CharSequenceContainsAssertions
e: /Users/neelkamath/Documents/GitHub/atrium/domain/api/atrium-domain-api-common/src/main/kotlin/ch/tutteli/atrium/domain/creating/charsequence/contains/creators/CharSequenceContainsAssertions.kt: (49, 5): Platform declaration clash: The following declarations have the same JVM signature (regex(Lch/tutteli/atrium/domain/creating/charsequence/contains/CharSequenceContains$CheckerOption;Ljava/util/List;)Lch/tutteli/atrium/assertions/AssertionGroup;):
    fun <T : CharSequence> regex(checkerOption: CharSequenceContains.CheckerOption<T, NoOpSearchBehaviour>, expected: List<String>): AssertionGroup defined in ch.tutteli.atrium.domain.creating.charsequence.contains.creators.CharSequenceContainsAssertions
    fun <T : CharSequence> regex(checkerOption: CharSequenceContains.CheckerOption<T, NoOpSearchBehaviour>, expected: List<Regex>): AssertionGroup defined in ch.tutteli.atrium.domain.creating.charsequence.contains.creators.CharSequenceContainsAssertions
```

robstoll:
No you are not, the problem comes from how the JVM handles generics. It applies a type erasure strategy (see e.g. https://www.geeksforgeeks.org/type-erasure-java/) You can add `@JvmName("stringRegex")` to the first overload (the one which expects `List<String>`)

neelkamath:
I don't think I can use the `@JvmName()` annotation on `interface` functions because it gives me the error `'@JvmName' annotation is not applicable to this declaration`. For example:
```kotlin
@JvmName("stringRegex")
fun <T : CharSequence> regex(
    checkerOption: CharSequenceContains.CheckerOption<T, NoOpSearchBehaviour>,
    expected: List<String>
): AssertionGroup
```

I tried annotating the implementing methods of the two classes. For example:
```kotlin
@JvmName("stringRegex")
override inline fun <T : CharSequence> regex(
    checkerOption: CharSequenceContains.CheckerOption<T, NoOpSearchBehaviour>,
    expected: List<String>
) = charSequenceContainsAssertions.regex(checkerOption, expected)
```
But that still gives me the `Platform declaration clash: The following declarations have the same JVM signature` error.

robstoll:
That's right, but you shouldn't add the method to the interface anyway, only on the builder (see description)

neelkamath:
@robstoll I'm confused with the `domain` work. For `modify CharSequenceContainsAssertions, expect a List<Regex> instead of a List<String> for regex`, am I supposed to _replace_ the `String` function with `regex`, or simply provide an overload in the builder?

robstoll:
@neelkamath for 
- `CharSequenceContainsAssertions` you shall remove the function with `List<String>` and add one for `List<Regex>` (or simply change `List<String>` to `List<Regex>`)
- for `CharSequenceAssertionsBuilder` you shall add an additional overload which expects `List<Regex>`

neelkamath:
I get the error:
```
> Task :atrium-api-cc-de_CH-jvm:compileTestKotlin FAILED
e: /Users/neelkamath/Documents/GitHub/atrium/apis/cc-de_CH/atrium-api-cc-de_CH-jvm/src/test/kotlin/ch/tutteli/atrium/api/cc/de_CH/CharSequenceContainsSpecBase.kt: (21, 87): Cannot choose among the following candidates without completing type inference: 
public fun <T : CharSequence> CharSequenceContains.CheckerOption<CharSequence, NoOpSearchBehaviour>.regex(pattern: String, vararg otherPatterns: String): AssertionPlant<CharSequence> defined in ch.tutteli.atrium.api.cc.de_CH
public fun <T : CharSequence> CharSequenceContains.CheckerOption<CharSequence, NoOpSearchBehaviour>.regex(pattern: Regex, vararg otherPatterns: Regex): AssertionPlant<CharSequence> defined in ch.tutteli.atrium.api.cc.de_CH

FAILURE: Build failed with an exception.
```
because of the way I overloaded code in the `apis` directory like:
```kotlin
fun <T : CharSequence> Expect<T>.containsRegex(pattern: String, vararg otherPatterns: String): Expect<T> =
    contains.atLeast(1).regex(pattern, *otherPatterns)

// New code:
fun <T : CharSequence> Expect<T>.containsRegex(pattern: Regex, vararg otherPatterns: Regex): Expect<T> =
    containsRegex(pattern.toString(), *otherPatterns.map { it.toString() }.toTypedArray())
```
Compilation errors occur in the `apis` directory's test suite such as:
```kotlin
abstract class CharSequenceContainsSpecBase {
    ...
    protected val regex = CharSequenceContains.CheckerOption<*, NoOpSearchBehaviour>::regex.name
    ...
}
```

domain/robstoll/atrium-domain-robstoll-common/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/creating/charsequence/contains/creators/CharSequenceContainsAssertionsImpl.kt
```diff
@@ -33,8 +33,8 @@ class CharSequenceContainsAssertionsImpl : CharSequenceContainsAssertions {
 
     override fun <T : CharSequence> regex(
         checkerOption: CharSequenceContains.CheckerOption<T, NoOpSearchBehaviour>,
-        expected: List<String>
-    ): AssertionGroup = _containsRegex(checkerOption, expected)
+        expected: List<Regex>
+    ): AssertionGroup = _containsRegex(checkerOption, expected.map { it.toString() })
```
robstoll: `it.pattern` but you need to adopt _containsRegex in a next step anyway. Thus you can also leave it this way

robstoll:
Please only add the function to fluent-en_GB and not to cc-de_CH (or any other API).

Concerning CharSequenceContainsSpecBase, you can help the kotlin compiler by specifying the KFunction type additional. Or in other words, do it in two steps:
```
private val regexKFun : KFunction3<CharSequenceContains.CheckerOption<*, NoOpSearchBehaviour>, String, Array<out String>, Expect<*>> =
        CharSequenceContains.CheckerOption<*, NoOpSearchBehaviour>::regex
protected val regex = regexKFun.name
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceAssertions.kt
```diff
@@ -96,6 +96,32 @@ fun <T : CharSequence> Expect<T>.containsNot(expected: Any, vararg otherExpected
...
+ *   `contains.atLeast(1).regex('a(b)?', 'a(b)?')`
+ *
+ * @param pattern The pattern which is expected to have a match against the input of the search.
+ * @param otherPatterns Additional patterns which are expected to have a match against the input of the search.
+ *
+ * @return This assertion container to support a fluent API.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ */
+fun <T : CharSequence> Expect<T>.containsRegex(pattern: Regex, vararg otherPatterns: Regex): Expect<T> =
+    containsRegex(pattern.toString(), *otherPatterns.map { it.toString() }.toTypedArray())
```
robstoll: as mentioned in KDoc, it is a shortcut for `contains.atLeast(1).regex(pattern, *otherPatterns)` please implement it this way (same as in line 97)

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceContainsCreators.kt
```diff
@@ -190,6 +190,32 @@ fun <T : CharSequence> CharSequenceContains.CheckerOption<T, NoOpSearchBehaviour
...
+ * @param pattern The pattern which is expected to have a match against the input of the search.
+ * @param otherPatterns Additional patterns which are expected to have a match against the input of the search.
+ *
+ * @return The [Expect] for which the assertion was built to support a fluent API.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ */
+fun <T : CharSequence> CharSequenceContains.CheckerOption<T, NoOpSearchBehaviour>.regex(
+    pattern: Regex,
+    vararg otherPatterns: Regex
+): Expect<T> = regex(pattern.toString(), *otherPatterns.map { it.toString() }.toTypedArray())
```
robstoll: You have implemented an overload for ExpectImpl.charSequence.contains.regex which expects `Regex` please use it here

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceContainsCreators.kt
```diff
@@ -190,6 +190,32 @@ fun <T : CharSequence> CharSequenceContains.CheckerOption<T, NoOpSearchBehaviour
...
+ * as well as the [otherPatterns] are expected to have a match, using a non disjoint search.
+ *
+ * By non disjoint is meant that `'aa'` in `'aaaa'` is found three times and not only two times.
+ * Also notice, that it does not search for unique matches. Meaning, if the input of the search is `'ab'` and [pattern]
+ * is defined as `'a(b)?'` and one of the [otherPatterns] is defined as `'a(b)?'` as well, then both match, even though
+ * they match the same sequence in the input of the search. Use an option such as [atLeast], [atMost] and [exactly] to
+ * control the number of occurrences you expect.
+ *
+ * Meaning you might want to use:
+ *   `contains.exactly(2).regex('a(b)?')`
```
robstoll: adjust the example, use Regex

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceAssertions.kt
```diff
@@ -96,6 +96,32 @@ fun <T : CharSequence> Expect<T>.containsNot(expected: Any, vararg otherExpected
...
+ * It is a shortcut for `contains.atLeast(1).regex(pattern, *otherPatterns)`.
+ *
+ * By non disjoint is meant that `'aa'` in `'aaaa'` is found three times and not only two times.
+ * Also notice, that it does not search for unique matches. Meaning, if the input of the search is `'ab'` and [pattern]
+ * is defined as `'a(b)?'` and one of the [otherPatterns] is defined as `'a(b)?'` as well, then both match, even though
+ * they match the same sequence in the input of the search. Use an option such as [atLeast], [atMost] and [exactly] to
+ * control the number of occurrences you expect.
+ *
+ * Meaning you might want to use:
+ *   `contains.exactly(2).regex('a(b)?')`
```
robstoll: adjust the example, use Regex

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsRegexAssertionsSpec.kt
```diff
@@ -17,6 +18,16 @@ class CharSequenceContainsRegexAssertionsSpec : Spek({
         "◆ ", "⚬ "
     )
 
+    object RegexSpec : ch.tutteli.atrium.specs.integration.CharSequenceContainsRegexAssertionsSpec(
+        getNameContainsRegex(),
+        getAtLeastTriple(),
+        getAtLeastIgnoringCaseTriple(),
+        getShortcutTriple(),
+        getAtMostTriple(),
+        getAtMostIgnoringCaseTriple(),
```
robstoll: Those test the functionality with `String` write own functions which use `Regex` and use those

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceAssertions.kt
neelkamath: Ah, ok. I had originally thought of delegating to the existing one as an overload of sorts. I'll update it.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceContainsCreators.kt
neelkamath: I had thought of changing `contains.exactly(2).regex('a(b)?')` to `contains.exactly(2).regex(Regex("""a(b)?"""))`, but the KDoc had used `'` instead of `"` for denoting `a(b)?`, so I thought it was pseudocode. If I understood correctly, I should also fix the quotes for the original `String` method's KDoc too, right?

robstoll: good catch, yes please also change the string example. please don't use triple quotes for Regex though. single quotes fit better here.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsRegexAssertionsSpec.kt
```diff
@@ -8,41 +8,49 @@ class CharSequenceContainsRegexAssertionsSpec : Spek({
...
         private fun getNameContainsRegex() = "$contains with search mode $regex"
 
-        private fun getAtLeastTriple() =
+        private fun getAtLeastTripleString() =
             { what: String, times: String -> "$contains $what $atLeast $times" } to
                 ("$contains.$atLeast.$regex" to Companion::containsAtLeast)
 
+        private fun getAtLeastTripleRegex() =
+            { what: Regex, times: String -> "$contains $what $atLeast $times" } to
+                ("$contains.$atLeast.$regex" to Companion::containsAtLeast)
```
robstoll: you also need two versions of containsAtLeast (same same but different for other functions)

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/CharSequenceContainsRegexAssertionsSpec.kt
```diff
@@ -200,3 +200,196 @@ abstract class CharSequenceContainsRegexAssertionsSpec(
         }
     }
 })
+
+abstract class CharSequenceContainsRegexAssertionsSpecRegex(
```
robstoll: no need to create this class, you can reuse `CharSequenceContainsRegexAssertionsSpec` also for regex.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsRegexAssertionsSpec.kt
neelkamath: Does that mean I should also overload `containsAtLeastIgnoringCase`? In so, I think I should also provide overloads for
```kotlin
@JvmName("regexIgnoringCase")
fun <T : CharSequence> CharSequenceContains.CheckerOption<T, IgnoringCaseSearchBehaviour>.regex
```
and
```kotlin
fun <T : CharSequence> CharSequenceContains.Builder<T, IgnoringCaseSearchBehaviour>.regex(
    pattern: String,
    vararg otherPatterns: String
): Expect<T>
```
in `ch/tutteli/atrium/api/fluent/en_GB/charSequenceContainsCreators.kt`.

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/CharSequenceContainsRegexAssertionsSpec.kt
neelkamath: In that case, how do I make the RegexSpex compile (since `CharSequenceContainsRegexAssertionsSpec` only accepts `String` versions)?

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsRegexAssertionsSpec.kt
robstoll: ah no, sorry, my fault, you don't have to overload it for those, so only two versions for containsAtLeast

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/CharSequenceContainsRegexAssertionsSpec.kt
robstoll: You have to implement containsAtLeastRegex in a way that it full fills the signature the spec requires (so the same as containsAtLeast) but uses the regex version internally and not the string version

neelkamath:
I couldn't find a function called `containsAtLeastRegex`. I thought of the `Regex` overload for `containsAtLeast`, but didn't understand how to make it fulfill the signature of `CharSequenceContainsRegexAssertionsSpec`.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsRegexAssertionsSpec.kt
```diff
@@ -32,17 +37,20 @@ class CharSequenceContainsRegexAssertionsSpec : Spek({
...
 
-        private fun getAtLeastTripleRegex() =
+        private fun getAtLeastTripleRegex(): Pair<(Regex, String) -> String, RegexFun> =
             { what: Regex, times: String -> "$contains $what $atLeast $times" } to
                 ("$contains.$atLeast.$regex" to Companion::containsAtLeast)
 
         private fun containsAtLeast(expect: Expect<CharSequence>, atLeast: Int, a: String, aX: Array<out String>) =
             expect.contains.atLeast(atLeast).regex(a, *aX)
 
+        private fun containsAtLeast(expect: Expect<CharSequence>, atLeast: Int, a: Regex, aX: Array<out Regex>) =
```
robstoll: please do not overload it, add suffix `Regex` and `String` as you did otherwise

robstoll: The function `containsAtLeastRegex` which you could not find is the one you have to write, sorry for the confusion :grin:

neelkamath: Sorry, I'm still kind of confused.
If I'm correct, I'm supposed to implement your stub like:
```kotlin
        private fun containsAtLeastString(expect: Expect<CharSequence>, atLeast: Int, a: String, aX: Array<out String>) =
            expect.contains.atLeast(atLeast).regex(a, *aX)

        private fun containsAtLeastRegex(expect: Expect<CharSequence>, atLeast: Int, a: Regex, aX: Array<out Regex>) =
            expect.contains.atLeast(atLeast).regex(a, *aX)
```
But without duplicating `CharSequenceContainsRegexAssertionsSpec` for the new Regex functions, I still get compile-time errors for
```kotlin
object RegexSpec : ch.tutteli.atrium.specs.integration.CharSequenceContainsRegexAssertionsSpec(
        getNameContainsRegex(),
        getAtLeastTripleRegex(),
        getAtLeastIgnoringCaseTripleRegex(),
        getShortcutTripleRegex(),
        getAtMostTripleRegex(),
        getAtMostIgnoringCaseTripleRegex(),
        "◆ ", "⚬ "
    )
```
because `CharSequenceContainsRegexAssertionsSpec` wants the original `String` versions.

robstoll:
@neelkamath note, before I can merge your PR eventually you need to add the following to the description:
```
I confirm that I have read the Contributor Agreements v1.0, agree to be bound on them and confirm that my contribution is compliant.
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsRegexAssertionsSpec.kt
```diff
@@ -37,26 +32,30 @@ class CharSequenceContainsRegexAssertionsSpec : Spek({
...
-        private fun containsAtLeast(expect: Expect<CharSequence>, atLeast: Int, a: Regex, aX: Array<out Regex>) =
-            expect.contains.atLeast(atLeast).regex(a, *aX)
+        private fun containsAtLeastRegex(expect: Expect<CharSequence>, atLeast: Int, a: String, aX: Array<out String>) =
+            expect.contains.atLeast(atLeast).regex(a.toRegex(), *aX.map { it.toRegex() }.toTypedArray())
 
         private fun getAtLeastIgnoringCaseTripleString() =
             { what: String, times: String -> "$contains $ignoringCase $what $atLeast $times" } to
                 ("$contains.$atLeast.$ignoringCase.$regex" to Companion::containsAtLeastIgnoringCase)
 
         private fun getAtLeastIgnoringCaseTripleRegex() =
```
robstoll: remove this function, is the same as getAtLeastIgnoringCaseTripleString

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsRegexAssertionsSpec.kt
```diff
@@ -69,32 +68,38 @@ class CharSequenceContainsRegexAssertionsSpec : Spek({
...
             expect.contains.atMost(atMost).regex(a, *aX)
 
+        private fun containsAtMostRegex(expect: Expect<CharSequence>, atMost: Int, a: String, aX: Array<out String>) =
+            expect.contains.atMost(atMost).regex(a.toRegex(), *aX.map { it.toRegex() }.toTypedArray())
+
         private fun getAtMostIgnoringCaseTripleString() =
             { what: String, times: String -> "$contains $ignoringCase $what $atMost $times" } to
                 ("$contains.$ignoringCase.$atMost.$regex" to Companion::containsAtMostIgnoringCase)
 
         private fun getAtMostIgnoringCaseTripleRegex() =
```
robstoll: remove this function, is the same as getAtMostIgnoringCaseTripleString

domain/robstoll-lib/atrium-domain-robstoll-lib-common/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/charsequence/contains/searchers/RegexSearcher.kt
```diff
@@ -8,7 +8,7 @@ import ch.tutteli.atrium.domain.creating.charsequence.contains.searchbehaviours.
  *
  * It performs a non-disjoint search, meaning searching 'aa?' in 'aaaa' has 4 hits.
  */
-class RegexSearcher : Searcher<NoOpSearchBehaviour> {
+class RegexSearcher : Searcher<NoOpSearchBehaviour, Regex> {
     override fun search(searchIn: CharSequence, searchFor: Any): Int = search(searchIn, Regex(searchFor.toString()))
```
robstoll: searchFor should then by Regex no?

neelkamath:
I'm unsure how to fix the tests

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsRegexAssertionsSpec.kt
```diff
@@ -34,11 +34,11 @@ class CharSequenceContainsRegexAssertionsSpec : Spek({
 
         private fun getAtLeastTripleString() =
             { what: String, times: String -> "$contains $what $atLeast $times" } to
-                ("$contains.$atLeast.$regex" to Companion::containsAtLeastString)
+                ("$contains.$atLeast.$regex" to ::containsAtLeastString)
 
         private fun getAtLeastTripleRegex() =
             { what: String, times: String -> "$contains $what $atLeast $times" } to
-                ("$contains.$atLeast.$regex" to Companion::containsAtLeastRegex)
+                ("$contains.$atLeast.$regex" to ::getAtLeastIgnoringCaseTripleString)
```
robstoll: that's a mistake, keep `containsAtLeastRegex` as before

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsRegexAssertionsSpec.kt
robstoll: You removed the wrong function, keep containsAtLeastRegex and remove getAtLeastIgnoringCaseTripleRegex instead

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsRegexAssertionsSpec.kt
```diff
@@ -82,25 +79,22 @@ class CharSequenceContainsRegexAssertionsSpec : Spek({
 
         private fun getAtMostTripleString() =
             { what: String, times: String -> "$contains $what $atMost $times" } to
-                ("$contains.$atMost.$regex" to Companion::containsAtMostString)
+                ("$contains.$atMost.$regex" to ::containsAtMostString)
 
         private fun getAtMostTripleRegex() =
             { what: String, times: String -> "$contains $what $atMost $times" } to
-                ("$contains.$atMost.$regex" to Companion::containsAtMostRegex)
+                ("$contains.$atMost.$regex" to ::getAtMostIgnoringCaseTripleString)
```
robstoll: same here, this way you don't test what you want

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsRegexAssertionsSpec.kt
neelkamath: If I delete `getAtLeastIgnoringCaseTripleRegex` and `getAtMostIgnoringCaseTripleRegex`, what do I replace them with in the test's arguments below?
```kotlin
    object RegexSpec : ch.tutteli.atrium.specs.integration.CharSequenceContainsRegexAssertionsSpec(
        getNameContainsRegex(),
        getAtLeastTripleRegex(),
        getAtLeastIgnoringCaseTripleRegex(),
        getShortcutTripleRegex(),
        getAtMostTripleRegex(),
        getAtMostIgnoringCaseTripleRegex(),
        "◆ ", "⚬ "
    )
```

robstoll: My bad, you are right. So keep both :)

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsRegexAssertionsSpec.kt
```diff
@@ -6,31 +6,54 @@ import org.spekframework.spek2.Spek
...
+        getShortcutTripleString(),
+        getAtMostTripleString(),
+        getAtMostIgnoringCaseTripleString(),
+        "◆ ", "⚬ "
+    )
+
+    object RegexSpec : ch.tutteli.atrium.specs.integration.CharSequenceContainsRegexAssertionsSpec(
+        getNameContainsRegex(),
+        getAtLeastTripleRegex(),
+        getAtLeastIgnoringCaseTripleRegex(),
```
robstoll: use 
```suggestion
        getAtLeastIgnoringCaseTripleString(),
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsRegexAssertionsSpec.kt
```diff
@@ -6,31 +6,54 @@ import org.spekframework.spek2.Spek
...
+        getShortcutTripleRegex(),
+        getAtMostTripleRegex(),
+        getAtMostIgnoringCaseTripleRegex(),
```
robstoll: use
```suggestion
        getAtMostIgnoringCaseTripleString(),
```

robstoll:
@neelkamath you still have unresolved issues, do you need help with them?

neelkamath:
Yes, I get a stack overflow error when running the tests

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsRegexAssertionsSpec.kt
robstoll: I see you have problems spinning your head around this. Please just apply my suggestion, you don't need an own getAtLeastIgnoringCaseTripleRegex it is the same as getAtLeastIgnoringCaseTripleString()

robstoll:
@neelkamath what do you prefer, solve it yourself (with my hints) or shall I take over and finish it?

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsRegexAssertionsSpec.kt
robstoll: same here

neelkamath:
I'm fine with hints, unless you'd prefer to finish it yourself.

domain/robstoll-lib/atrium-domain-robstoll-lib-common/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/charsequence/contains/searchers/IgnoringCaseRegexSearcher.kt
```diff
@@ -10,7 +10,7 @@ import ch.tutteli.atrium.domain.creating.charsequence.contains.searchbehaviours.
 class IgnoringCaseRegexSearcher : Searcher<IgnoringCaseSearchBehaviour, Regex> {
     private val searcher = RegexSearcher()
 
-    override fun search(searchIn: CharSequence, searchFor: Any): Int {
+    override fun search(searchIn: CharSequence, searchFor: Regex): Int {
         val pattern = Regex(searchFor.toString(), RegexOption.IGNORE_CASE)
```
robstoll: searchFor.pattern

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceAssertions.kt
```diff
@@ -96,6 +96,32 @@ fun <T : CharSequence> Expect<T>.containsNot(expected: Any, vararg otherExpected
...
+ * Meaning you might want to use:
+ *   `contains.exactly(2).regex(Regex("a(b)?"))`
+ * instead of:
+ *   `contains.atLeast(1).regex(Regex("a(b)?"), Regex("a(b)?"))`
+ *
+ * @param pattern The pattern which is expected to have a match against the input of the search.
+ * @param otherPatterns Additional patterns which are expected to have a match against the input of the search.
+ *
+ * @return This assertion container to support a fluent API.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
```
robstoll: Please add `@since 0.9.0`

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceContainsCreators.kt
```diff
@@ -190,6 +190,32 @@ fun <T : CharSequence> CharSequenceContains.CheckerOption<T, NoOpSearchBehaviour
...
+ * Meaning you might want to use:
+ *   `contains.exactly(2).regex(Regex("a(b)?"))`
+ * instead of:
+ *   `contains.atLeast(1).regex(Regex("a(b)?"), Regex("a(b)?"))`
+ *
+ * @param pattern The pattern which is expected to have a match against the input of the search.
+ * @param otherPatterns Additional patterns which are expected to have a match against the input of the search.
+ *
+ * @return The [Expect] for which the assertion was built to support a fluent API.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
```
robstoll: Please add `@since 0.9.0`

domain/robstoll-lib/atrium-domain-robstoll-lib-common/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/charsequence/contains/creators/creators.kt
```diff
@@ -15,17 +15,24 @@ import ch.tutteli.atrium.translations.DescriptionCharSequenceAssertion.VALUE
 fun <T : CharSequence> _containsValues(
     checkerOption: CharSequenceContains.CheckerOption<T, NoOpSearchBehaviour>,
     expected: List<Any>
-): AssertionGroup = checkOnlyAllowedTypeNotEmptyStringAndCreateAssertionGroup(checkerOption, IndexSearcher(), expected)
+): AssertionGroup = checkOnlyAllowedTypeNotEmptyStringAndCreateAssertionGroup(
+    checkerOption,
+    IndexSearcher(),
+    expected.map { it.toString() }
```
robstoll: toString should not be necessary here (see comment further down below)

domain/robstoll-lib/atrium-domain-robstoll-lib-common/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/charsequence/contains/creators/creators.kt
```diff
@@ -15,17 +15,24 @@ import ch.tutteli.atrium.translations.DescriptionCharSequenceAssertion.VALUE
...
 
 fun <T : CharSequence> _containsValuesIgnoringCase(
     checkerOption: CharSequenceContains.CheckerOption<T, IgnoringCaseSearchBehaviour>,
     expected: List<Any>
-): AssertionGroup =
-    checkOnlyAllowedTypeNotEmptyStringAndCreateAssertionGroup(checkerOption, IgnoringCaseIndexSearcher(), expected)
+): AssertionGroup = checkOnlyAllowedTypeNotEmptyStringAndCreateAssertionGroup(
+    checkerOption,
+    IgnoringCaseIndexSearcher(),
+    expected.map { it.toString() }
```
robstoll: toString should not be necessary here (see next comment)

domain/robstoll-lib/atrium-domain-robstoll-lib-common/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/charsequence/contains/creators/creators.kt
```diff
@@ -44,7 +51,7 @@ private fun <T : CharSequence, S : CharSequenceContains.SearchBehaviour> checkOn
             "Searching for an empty CharSequence does not make sense. You probably forgot to specify the search criterion."
         }
     }
-    return createAssertionGroup(checkerOption, searcher, expected, VALUE)
+    return createAssertionGroup(checkerOption, searcher, expected.map { it.toString() }, VALUE)
```
robstoll: that's good, toString should happen here but not elsewhere

domain/robstoll-lib/atrium-domain-robstoll-lib-common/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/charsequence/contains/creators/creators.kt
```diff
@@ -59,17 +66,22 @@ fun <T : CharSequence> _containsDefaultTranslationOfIgnoringCase(
...
 ): AssertionGroup = createAssertionGroup(checkerOption, RegexSearcher(), expected, STRING_MATCHING_REGEX)
 
 fun <T : CharSequence> _containsRegexIgnoringCase(
     checkerOption: CharSequenceContains.CheckerOption<T, IgnoringCaseSearchBehaviour>,
     expected: List<String>
-): AssertionGroup = createAssertionGroup(checkerOption, IgnoringCaseRegexSearcher(), expected, STRING_MATCHING_REGEX)
+): AssertionGroup = createAssertionGroup(
+    checkerOption,
+    IgnoringCaseRegexSearcher(),
+    expected.map { it.toRegex() },
```
robstoll: should not be necessary (revert the change)

robstoll:
@neelkamath I can merge your PR into a separate branch in case you aim to count this PR towards hacktoberfest -- IMO it is wort to be counted even if it is not entirely finished yet. You can then continue from the branch. Let me know if you want that.

domain/robstoll-lib/atrium-domain-robstoll-lib-common/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/charsequence/contains/creators/creators.kt
neelkamath: I think it fails to compile if I revert `expected.map { it.toRegex() }` to `expected`.

neelkamath:
It's ok, you can just merge once it's ready.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceAssertions.kt
```diff
@@ -179,6 +207,8 @@ fun <T : CharSequence> Expect<T>.endsNotWith(expected: Char) = endsNotWith(expec
  *
  * @return This assertion container to support a fluent API.
  * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ *
+ * @since 0.9.0
```
robstoll: please remove this one, that's not the one you added in this PR (the other one in file common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceContainsCreators.kt is still open though)

domain/robstoll-lib/atrium-domain-robstoll-lib-common/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/charsequence/contains/creators/creators.kt
robstoll: you are right. Well, in this case I have an additional suggestion. It does not really make sense to convert it here to regex and then unpack it in `IgnoreCaseRegexSearcher` just to repack it with the RgexOptin.IGNORE_CASE. Thus:
- use the following here: `Regex(it, RegexOption.IGNORE_CASE)`
- use the RegexSearcher one line above
- remove IgnoringCaseRegexSearcher

robstoll:
@neelkamath The build fails due to a missing type parameter for `misc/atrium-assertions/src/main/kotlin/ch/tutteli/atrium/assertions/charsequence/contains/creators/CharSequenceContainsAssertionCreator.kt`

domain/robstoll-lib/atrium-domain-robstoll-lib-common/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/charsequence/contains/creators/creators.kt
```diff
@@ -15,17 +15,24 @@ import ch.tutteli.atrium.translations.DescriptionCharSequenceAssertion.VALUE
...
+): AssertionGroup = checkOnlyAllowedTypeNotEmptyStringAndCreateAssertionGroup(
+    checkerOption,
+    IgnoringCaseIndexSearcher(),
+    expected.map { it.toString() }
+)
 
 private fun <T : CharSequence, S : CharSequenceContains.SearchBehaviour> checkOnlyAllowedTypeNotEmptyStringAndCreateAssertionGroup(
     checkerOption: CharSequenceContains.CheckerOption<T, S>,
-    searcher: CharSequenceContains.Searcher<S>,
+    searcher: CharSequenceContains.Searcher<S, String>,
```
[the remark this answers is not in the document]
robstoll: This one is still open IMO, or why did you mark it as `resolved`?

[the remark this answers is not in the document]
robstoll: Ah no, you are right, `expect: List<Any>` that's how it should be, sorry :)

neelkamath:
I get a compile-time error in `ch/tutteli/atrium/domain/robstoll/lib/creating/charsequence/contains/creators/creators.kt` for 
```kotlin
): AssertionGroup = createAssertionGroup(
    checkerOption,
    RegexSearcher(),
    expected.map { Regex(it, RegexOption.IGNORE_CASE) },
    STRING_MATCHING_REGEX
)
```
I guess I misinterpreted your suggestion

robstoll:
Sorry, I am doing the reviews mostly without actually looking at the code. This way I don't require that much time but forget sometimes things or make wrong assumptions. I checked out your code now and can see that I forgot the SearchBehaviour. We still require `IgnoreRegexSearch` because it implements the IgnoringCaseSearchBehaviour (in contrast to RegexSearcher). Which means your previous code was actually correct.

However, I still think it does not make sense to create a Regex and unpack it in  IgnoringCaseRegexSearcher. Thus:
use:
```
fun <T : CharSequence> _containsRegexIgnoringCase(
    checkerOption: CharSequenceContains.CheckerOption<T, IgnoringCaseSearchBehaviour>,
    expected: List<String>
): AssertionGroup = createAssertionGroup(
    checkerOption,
    IgnoringCaseRegexSearcher(),
    expected,
    STRING_MATCHING_REGEX
)
```
And change Ignoring.. to
```
class IgnoringCaseRegexSearcher : Searcher<IgnoringCaseSearchBehaviour, String>
```
(adopt the search method)

robstoll:
@neelkamath I see why the build is failing. It is due to the `expected.map { it.toString() }` in checkOnlyAllowedTypeNotEmptyStringAndCreateAssertionGroup. This changes also the presentation of the expected object we pass to `contains`. For instance: 
```
expect("hello number 1").contains('h', 1)
```
Currently we turn 'h' and 1 into String. You certainly did that on my suggestions. But we can easily fix it. change to `searcher: CharSequenceContains.Searcher<S, Any>,` and use `expected` instead of `expected.map { it.toString() }` + we need to use `Any` for IgnoringCaseIndexSearcher and for IndexSearcher

neelkamath:
Do you mean change the definitions of the classes such as from `class IgnoringCaseIndexSearcher : Searcher<IgnoringCaseSearchBehaviour, String>` to `class IgnoringCaseIndexSearcher : Searcher<IgnoringCaseSearchBehaviour, Any>`?

robstoll:
Yes

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceContainsCreators.kt
robstoll:
```suggestion
 * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
 *
 * @since 0.9.0
```

robstoll: sorry, was too fast, you added it in the meantime

robstoll:
@neelkamath the code looks good to me now but we need to fix the tests as the behaviour has changed a bit. Change Line 155 and line 193 in CharSequenceContainsRegexAssertionsSpec remove the wrapping `\"` this is because it is no longer a string but a REGEX

`./gr build` should then succeed. However, the backward-compatiblity test (which compiles and runs the tests of version 0.8.0 against the new version) as well as the binary backward compatibility test (which runs the pre-compiled 0.8.0 tests against the new version) will fail as they expect a string as well. 

We can fix the bc test by manipulating the sources upfront. This has to be done build.gradle of atrium-bc-test (please rebase on master, the atrium-bc-test has moved to /misc/tools/ as far as I can see this is not yet the case in your branch)

The bbc tests on the hand cannot be manipulated (well, they could on a byte level but that's too much of a hassle). Thus, I would say we can forgive those tests (I hope you are a merciful person as well :grin:) for this you have to declare that in the build.gradle of atrium-bc-test 

I can also take over from here (or maybe after you fixed the tests in the current version). Dealing with bc and bbc test is surely something different again. Let me know what you prefer

neelkamath:
My git and Gradle skills are limited to that of a noob, so I'd like if you could take over from here. Thanks.

robstoll:
@neelkamath thanks for your contribution :+1:

neelkamath:
Thanks for your unwavering support.
I think I'll try contributing after I've learned a little more about what I want to do exactly though.

PR #177, opened 2019-10-03
add File.asPath assertions

lelloman:
Being my first contribution attempt I'm opening this PR to check if I'm on the right track, I need to add JavaDoc though.
I also have a stupid question, but, (how) can I write a test for this new assertion?

### DIFF
--- a/apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/fileAssertions.kt
+++ b/apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/fileAssertions.kt
@@ -1,0 +1,29 @@
+package ch.tutteli.atrium.api.fluent.en_GB.jdk8
+
+import ch.tutteli.atrium.creating.Expect
+import ch.tutteli.atrium.domain.builders.ExpectImpl
+import java.io.File
+import java.nio.file.Path
+
+/**
+ * Turns `Expect<File>` into `Expect<Path>`.
+ *
+ * The transformation as such is not reflected in reporting.
+ * Use `feature(File::toPath)` if you want to show the transformation in reporting.
+ *
+ * @return The newly created [Expect] for the transformed subject.
+ */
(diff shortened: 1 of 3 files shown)


robstoll:
Looks good so far. I suggest you copy the KDoc of sequenceAssertions.kt -> asIterable and adopt it accordingly. You can copy ch.tutteli.atrium.specs.integration.ArrayAsIterableAssertionsSpec and adopt it for File

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/fileAssertions.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import ch.tutteli.atrium.creating.Expect
+import ch.tutteli.atrium.domain.builders.ExpectImpl
+import java.io.File
+import java.nio.file.Path
+
+fun <T : File> Expect<T>.asPath(): Expect<Path> =
+    ExpectImpl.changeSubject.unreported(this) { it.toPath() }
+
+fun <T : File> Expect<T>.asPath(assertionCreator: Expect<Path>.() -> Unit): Expect<Path> =
```
robstoll: This should return `Expect<T>`, wrap the whole expression into an `apply` (see sequenceAssertions for an example)

lelloman: Fixed

lelloman:
Added KDoc and one test case, I think this should be OK as the test does fail when using a different path in the assertion.

robstoll:
@lelloman thanks for the contribution :+1: 
Fancy another one? https://github.com/robstoll/atrium/issues?q=is%3Aissue+is%3Aopen+label%3Ahacktoberfest
I think 169 could be a good one, similar but requires a bit more

lelloman:
@robstoll No problem, actually I came for the t-shirt but I'm gonna stay for the code :) 
169 Seems doable, yeah I'd like to pick it up, I think I can complete it in ~ a week. I also asked for a slack account on kotlinlang, hope I can seek help there if I get lost.

PR #178, opened 2019-10-03
update spek specification link in CONTRIBUTING.md

lelloman:
The link to spec specification is giving 404.
----


robstoll:
Good catch, thanks for the correction.

Issue #193, opened 2019-10-08
upgrade spek to 2.0.8 and use spek's `include`


robstoll:
*Platform* (jvm, jdk8, js, android): all

## Code related feature
Since Atrium's `include` was pushed back to spek and is included since 2.0.7 we can now use it instead. Thus:
- upgrade spek to 2.0.8
- remove `include` from /misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/SpekExtensions.kt
- add imports for spek's include if required

Anubhav007:
I'll work on it.

robstoll:
I am unsure what your problem is with the last point.It might well be that you don't need to add any import (that's why I wrote `if required`). Try if the project still builds by running `./gr build` (`gr build` in case of windows) in the root of the project. If it does, then you can simply skip step 3.

Anubhav007:
So i tried running the build and there were compilation errors (eg - IterableAnyAssertionsSpec) where **include** is used. For step 1, I updated the **spek2_version = '2.0.8'** in build.gradle line 22.

robstoll:
Could you please push your changes to github and create a PR (prefix the title with [WIP] where WIP stands for work in progress)? This makes review much easier for me

robstoll:
Ah I think I see your problem. Notice that we are currently using Spek 1 in certain modules. You don't have to change these modules. We use spek2 currently in api/fluent-en_GB and extensions as well as in specs-common (specs-jvm etc.). Maybe even some more modules, in short: 
- ignore files which have an `import org.jetbrains.spek.api.Spek`
- only adapt files which have an `import org.spekframework.spek2.Spek`

Anubhav007:
@robstoll I have raised a WIP PR - https://github.com/robstoll/atrium/pull/195 Please let me know what is need to change there.

PR #195, opened 2019-10-09
WIP #193 upgrade spek to 2.0.8 and use spek's `include`

Anubhav007:
----

### DIFF
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsAtLeastAssertionsSpec.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/CharSequenceContainsAtLeastAssertionsSpec.kt
@@ -1,7 +1,6 @@
 package ch.tutteli.atrium.api.fluent.en_GB
 
 import ch.tutteli.atrium.creating.Expect
-import ch.tutteli.atrium.specs.include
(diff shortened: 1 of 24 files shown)


robstoll:
Have a look at the output of the build it clearly states what is wrong, it cannot find
```
import ch.tutteli.atrium.specs.include
```
because that is the one you removed in step 1 :slightly_smiling_face: 
Thus you need to remove this line

Anubhav007:
@robstoll I actually removed those imports but I was still getting errors wherever the include was used. Looks like gradle was not refreshing . The issue got resolved by upgrading the intelliJ. I will raise PR with the changes in a while.

robstoll:
You need to re-import a gradle project in Intellij if you modify something in a build.gradle file

robstoll:
@Anubhav007 try if spek 2.0.7 works

Anubhav007:
@robstoll Looks like only travis build is failing on updating to 2.0.7

robstoll:
@Anubhav007 looks already much better (I guess there is a bug in spek 2.0.8 , I'll investigate later). As last step you have to include the signature of `include` to org.spekframework.spek2.dsl.Root in /misc/specs/atrium-specs-js/src/main/kotlin/org/spekframework/spek2/dsl/Root.kt

Anubhav007:
@robstoll Could you help me with the signature part?

robstoll:
Just copy the signature from spek. Open up a spec and navigate to the definition of `include`

Anubhav007:
@robstoll The build is passing now

robstoll:
@Anubhav007 thanks for your contribution :+1:

PR #197, opened 2019-10-09
Path.endsNotWith

segunfamisa:
## Implement Path.endsNotWith #190 

### DIFF
--- a/domain/builders/atrium-domain-builders-jvm/src/main/kotlin/ch/tutteli/atrium/domain/builders/creating/PathAssertionsBuilder.kt
+++ b/domain/builders/atrium-domain-builders-jvm/src/main/kotlin/ch/tutteli/atrium/domain/builders/creating/PathAssertionsBuilder.kt
@@ -23,7 +23,7 @@ object PathAssertionsBuilder : PathAssertions {
     override fun <T : Path> endsWith(assertionContainer: Expect<T>, expected: Path) =
         pathAssertions.endsWith(assertionContainer, expected)
 
-    override fun <T : Path> endsNotWith(assertionContainer: Expect<T>, expected: T) =
+    override fun <T : Path> endsNotWith(assertionContainer: Expect<T>, expected: Path) =
         pathAssertions.endsNotWith(assertionContainer, expected)
 
     override inline fun <T : Path> exists(assertionContainer: Expect<T>) =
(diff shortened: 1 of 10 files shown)


domain/builders/atrium-domain-builders-jvm/src/main/kotlin/ch/tutteli/atrium/domain/builders/creating/PathAssertionsBuilder.kt
```diff
@@ -17,6 +17,9 @@ object PathAssertionsBuilder : PathAssertions {
     override fun <T : Path> endsWith(assertionContainer: Expect<T>, expected: Path) =
         pathAssertions.endsWith(assertionContainer, expected)
 
+    override fun <T : Path> endsNotWith(assertionContainer: Expect<T>, expected: Path) =
```
robstoll: please `inline` the function (could you do the same for endsWith, I apparently did not see it during the review)

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
```diff
@@ -112,5 +113,26 @@ abstract class PathAssertionsSpec(
             }
         }
     }
+
+    describeFun(endsNotWith.name) {
```
robstoll: personally I would fuse this describe with the above one as this is the opposite of endsWith. This way you could reuse the `context...` but it's also fine if you leave it this way

robstoll:
@segunfamisa please rebase on master and resolve the conflicts

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
robstoll: I see that all others are also this way, so let's keep it for now

segunfamisa:
Cool!

Thanks for the review. I'll inline those two functions (completely forgot
it could be used inline as well:)) and rebase.

segunfamisa:
Hey @robstoll 

Sorry, I hastily sent that on the go from my phone. I had thought you meant infix, not inline.

Is there any benefit of inlining it? just asking since I don't see any lambda functions that the compiler could inline. Also I'm getting compiler warnings saying `"...expected performance impact of inlining is insignificant"` 

But I'm also not sure maybe there is something I'm missing.

robstoll:
You should not get a compiler warning as there is `@file:Suppress("OVERRIDE_BY_INLINE", "NOTHING_TO_INLINE")` in the PathAssertionsBuilder file
I inline for backward compatibility reasons, see https://github.com/robstoll/atrium-roadmap/issues/25#issuecomment-526239006 => rough analysis => delegation in case you are curious

segunfamisa:
Thanks for the explanation. I'll have a look at the roadmap link you shared!

robstoll:
@segunfamisa thanks for the contribution :+1:

Fancy another one? (for instance #200 would be a good one to continue with):
https://github.com/robstoll/atrium/issues?q=is%3Aissue+is%3Aopen+label%3Ahacktoberfest+no%3Aassignee

(and don't forget to star Atrium :wink:)

Issue #200, opened 2019-10-10
simplify PathAssertionsSpec fuse describe of similar functions


robstoll:
*Platform* (jvm, jdk8, js, android): jdk8

## Code related feature

functions like exists and existsNot form a counterpart of each other. Thus should be reflected in tests IMO and we can also simplify the tests due to this.

Have a look at ch.tutteli.atrium.specs.integration.CharSequenceAssertionsSpec for instance isEmtpy, isNotEmpty => I would like to have the same for exists/existsNot, startsWith/startsNotWith, endsWith/endsNotWith

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime,
  we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

segunfamisa:
I'll work on this

Issue #167, opened 2019-10-02
compile android class files into dex bytecode


robstoll:
*Platform* (jvm, jdk8, js, android): android

## Code related feature
Currently we don't have a proof that the class files are actually compatible with d8. 
It would be nice if we could compile the resulting class files of a build of an `-android` module with the d8 dexer.

TODO:
- [ ] create a gradle task `checkDexer` (based on JavaExec) for all android projects => do that in the build.gradle in the root -> use `configure(getAndroidProjects()){...}`. The task shall convert all class files to dex-files using the d8 dexer
- [ ] make sure `check` depends on the new task and make sure it is run after `compileKotlin` (use dependsOn)

Some additional hints (the rest needs to be researched):
- docu on how JavaExec works https://docs.gradle.org/current/dsl/org.gradle.api.tasks.JavaExec.html
- dependency to r8 which includes d8 is described here https://mvnrepository.com/artifact/com.android.tools/r8/1.6.84 (not sure in which repositories it is published)
- repo of r8, probably contains further information on how to use it: https://r8.googlesource.com/r8

segunfamisa:
I'll work on this

robstoll:
@segunfamisa do you need help?

assaflei:
I'll try taking this

PR #206, opened 2019-10-11
Fuse describe of similar functions in PathAssertionsSpec

segunfamisa:
Fuse describe of similar functions in `PathAssertionsSpec` #200 

--

### DIFF
--- a/misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
+++ b/misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
@@ -144,18 +144,18 @@ abstract class PathAssertionsSpec(
         }
     }
 
-    describeFun(endsWith.name) {
+    describeFun(endsWith.name, endsNotWith.name) {
         val endsWithFun = endsWith.lambda
+        val endsNotWithFun = endsNotWith.lambda
+
         context("ends with") {
             it("does not throw") {
                 expect(Paths.get("/not/existed/for/test"))
                     .endsWithFun(Paths.get("for/test"))
             }
-        }
-        val expectedMessageIfNotEndsWith = "${DescriptionPathAssertion.ENDS_WITH.getDefault()}:"
 
-        context("not ends with") {
             it("throws an AssertionError") {
+                val expectedMessageIfNotEndsWith = "${DescriptionPathAssertion.ENDS_WITH.getDefault()}:"
                 expect {
                     expect(Paths.get("/not/existed/for/test"))
                         .endsWithFun(Paths.get("/for/test"))
@@ -164,11 +164,8 @@ abstract class PathAssertionsSpec(
                 }
             }
         }
-    }
 
-    describeFun(endsNotWith.name) {
-        val endsNotWithFun = endsNotWith.lambda
-        context("path ends with") {
+        context("path ends not with") {
             it("throws an AssertionError") {
                 expect {
                     expect(Paths.get("/path/ends/with/this"))
@@ -177,9 +174,7 @@ abstract class PathAssertionsSpec(
                     messageContains("${DescriptionPathAssertion.ENDS_NOT_WITH.getDefault()}:")
                 }
             }
-        }
 
-        context("path does not end with") {
             it("does not throw") {
                 expect(Paths.get("/path/ends/with/this"))
                     .endsNotWithFun(Paths.get("with/another"))
--- a/misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
+++ b/misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
@@ -131,35 +131,33 @@ abstract class PathAssertionsSpec(
         val endsWithFun = endsWith.lambda
         val endsNotWithFun = endsNotWith.lambda
 
-        context("ends with") {
-            it("${endsWith.name} does not throw") {
-                expect(Paths.get("/not/existed/for/test"))
+        context("path '/some/path/for/test'") {
+            it("${endsWith.name} 'for/test' does not throw") {
+                expect(Paths.get("/some/path/for/test"))
                     .endsWithFun(Paths.get("for/test"))
             }
 
-            it("${endsWith.name} throws an AssertionError") {
+            it("${endsWith.name} 'for/another' throws an AssertionError") {
                 expect {
-                    expect(Paths.get("/not/existed/for/test"))
-                        .endsWithFun(Paths.get("/for/test"))
+                    expect(Paths.get("/some/path/for/test"))
+                        .endsWithFun(Paths.get("for/another"))
                 }.toThrow<AssertionError> {
                     messageContains("${DescriptionPathAssertion.ENDS_WITH.getDefault()}:")
                 }
             }
-        }
 
-        context("path ends not with") {
-            it("${endsNotWith.name} throws an AssertionError") {
+            it("${endsNotWith.name} 'for/test' throws an AssertionError") {
                 expect {
-                    expect(Paths.get("/path/ends/with/this"))
-                        .endsNotWithFun(Paths.get("with/this"))
+                    expect(Paths.get("/some/path/for/test"))
+                        .endsNotWithFun(Paths.get("for/test"))
                 }.toThrow<AssertionError> {
                     messageContains("${DescriptionPathAssertion.ENDS_NOT_WITH.getDefault()}:")
                 }
             }
 
-            it("${endsNotWith.name} does not throw") {
-                expect(Paths.get("/path/ends/with/this"))
-                    .endsNotWithFun(Paths.get("with/another"))
+            it("${endsNotWith.name} 'for/another' does not throw") {
+                expect(Paths.get("/some/path/for/test"))
+                    .endsNotWithFun(Paths.get("for/another"))
             }
         }
     }
--- a/misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
+++ b/misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
@@ -34,9 +34,10 @@ abstract class PathAssertionsSpec(
     fun describeFun(vararg funName: String, body: Suite.() -> Unit) =
         describeFunTemplate(describePrefix, funName, body = body)
 
-    describeFun(exists.name) {
+    describeFun(exists.name, existsNot.name) {
         val existsFun = exists.lambda
-        context("non existing") {
+        val existsNotFun = existsNot.lambda
+        context("exists") {
             it("throws an AssertionError") {
                 expect {
                     expect(Paths.get("nonExistingFile")).existsFun()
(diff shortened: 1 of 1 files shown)


misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
```diff
@@ -34,10 +34,11 @@ abstract class PathAssertionsSpec(
...
         describeFunTemplate(describePrefix, funName, body = body)
 
-    describeFun(exists.name) {
+    describeFun(exists.name, existsNot.name) {
         val existsFun = exists.lambda
-        context("non existing") {
-            it("throws an AssertionError") {
+        val existsNotFun = existsNot.lambda
+        context("file or folder exists") {
+            it("${exists.name} throws an AssertionError when file does not exist") {
```
robstoll: this contradicts the given context which states the `file or folder exists`. move this to `file or folder does not exist`
Moreover I would introduce two context, one for `file exists` and one for `folder exists` (same same but different for `file does not exist` and `folder does not exist`)

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
```diff
@@ -46,67 +47,55 @@ abstract class PathAssertionsSpec(
                     )
                 }
             }
-        }
-        context("existing file") {
-            it("does not throw") {
+
+            it("${exists.name} does not throw when file exists") {
```
robstoll: once you have two contexts you can simplify this to 
```suggestion
            it("${exists.name} does not throw") {
```
Same for folder below and same for the other `it` in the `file does not exist` and `folder does not exist`

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
```diff
@@ -46,67 +47,55 @@ abstract class PathAssertionsSpec(
...
         }
     }
 
-    describeFun(startsWith.name) {
+    describeFun(startsWith.name, startsNotWith.name) {
         val startsWithFun = startsWith.lambda
+        val startsNotWithFun = startsNotWith.lambda
         context("starts with") {
-            it("does not throw") {
+            it("${startsWith.name} does not throw when tested path starts with expected path") {
```
robstoll: see comments for Path above, the same applies here as well

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
```diff
@@ -144,18 +128,18 @@ abstract class PathAssertionsSpec(
...
     }
 
-    describeFun(endsWith.name) {
+    describeFun(endsWith.name, endsNotWith.name) {
         val endsWithFun = endsWith.lambda
+        val endsNotWithFun = endsNotWith.lambda
+
         context("ends with") {
-            it("does not throw") {
+            it("${endsWith.name} does not throw when tested path ends with expected path") {
```
robstoll: see comments for Path above, the same applies here as well.

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
segunfamisa: Fixed that and other related comments, thanks!

I added the additional description because I thought it would be helpful for test reporting

Now the test report looks like this and I think it's not clear which context the failing test belongs to:

image

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
```diff
@@ -34,79 +34,70 @@ abstract class PathAssertionsSpec(
...
                     .startsWithFun(Paths.get("/some/path/"))
             }
-        }
-
-        val expectedMessageIfNotStartsWith = "${DescriptionPathAssertion.STARTS_WITH.getDefault()}:"
 
-        context("does not start with") {
-            it("throws an AssertionError") {
+            it("${startsWith.name} throws an AssertionError") {
+                val expectedMessageIfNotStartsWith = "${DescriptionPathAssertion.STARTS_WITH.getDefault()}:"
```
robstoll: could you please inline expectedMessageIfNotStartsWith, the assertion in line 105 will be more readable this way

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
```diff
@@ -115,25 +106,20 @@ abstract class PathAssertionsSpec(
...
                     .startsNotWithFun(Paths.get("/some/pa"))
             }
-        }
-
-        val expectedMessageIfStartsWith = "${DescriptionPathAssertion.STARTS_NOT_WITH.getDefault()}:"
 
-        context("starts with") {
-            it("throws an AssertionError") {
+            it("${startsNotWith.name} throws an AssertionError") {
+                val expectedMessageIfStartsWith = "${DescriptionPathAssertion.STARTS_NOT_WITH.getDefault()}:"
```
robstoll: same here, please inline

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
```diff
@@ -144,18 +130,18 @@ abstract class PathAssertionsSpec(
...
                 expect(Paths.get("/not/existed/for/test"))
                     .endsWithFun(Paths.get("for/test"))
             }
-        }
-        val expectedMessageIfNotEndsWith = "${DescriptionPathAssertion.ENDS_WITH.getDefault()}:"
 
-        context("not ends with") {
-            it("throws an AssertionError") {
+            it("${endsWith.name} throws an AssertionError") {
+                val expectedMessageIfNotEndsWith = "${DescriptionPathAssertion.ENDS_WITH.getDefault()}:"
```
robstoll: please inline

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
```diff
@@ -34,79 +34,70 @@ abstract class PathAssertionsSpec(
...
                 expect(Paths.get("/some/path/for/test"))
                     .startsWithFun(Paths.get("/some/path/"))
             }
-        }
-
-        val expectedMessageIfNotStartsWith = "${DescriptionPathAssertion.STARTS_WITH.getDefault()}:"
 
-        context("does not start with") {
-            it("throws an AssertionError") {
+            it("${startsWith.name} throws an AssertionError") {
```
robstoll: this does not make sense if the context is `starts with` then `startsWith` should not fail. Maybe you can improve this by changing the context description to `context("path /some/path/for/test")` and then the `it("${startsWith.name} for/test throws AssertionError")` => if you do this, then change also the path in lin 103

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
```diff
@@ -115,25 +106,20 @@ abstract class PathAssertionsSpec(
...
                 expect(Paths.get("/some/path/for/test"))
                     .startsNotWithFun(Paths.get("/some/pa"))
             }
-        }
-
-        val expectedMessageIfStartsWith = "${DescriptionPathAssertion.STARTS_NOT_WITH.getDefault()}:"
 
-        context("starts with") {
-            it("throws an AssertionError") {
+            it("${startsNotWith.name} throws an AssertionError") {
```
robstoll: same here, I guess you don't require two contexts, one path is enough, you could even inlcude endsWith/endsNotWith in this context

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
robstoll: Yeah... well, it does not include the hierarchy as intellij does
image
Maybe there is a config setting so that this is reflected as well. 
I think it will already be better if you change the `it` and include the path as well

robstoll: I found this: https://github.com/gradle/gradle/issues/5273
Let's ignore this for know, I guess gradle will fix this and include the hierarchy in the future

segunfamisa: I see. Sounds good. It should get fixed in the future. I'll address the other comments later in the day. Thanks.

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
segunfamisa: Thanks! I used the same context for `startsWith`/`startsNotWith`, but I had a different `describeFun` containing just one `context` for `endsWith`/`endsNotWith` like `...integration/CharSequenceAssertionsSpec`

I hope that's fine?

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
segunfamisa: Got it. I think I had misunderstood what `"context"` meant in this project and how it relates to the function being tested.

Thanks for the explanation!

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
```diff
@@ -134,36 +131,33 @@ abstract class PathAssertionsSpec(
         val endsWithFun = endsWith.lambda
         val endsNotWithFun = endsNotWith.lambda
 
-        context("ends with") {
-            it("${endsWith.name} does not throw") {
-                expect(Paths.get("/not/existed/for/test"))
+        context("path '/some/path/for/test'") {
```
robstoll: no need for the context as it is the same as above

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
```diff
@@ -90,41 +90,38 @@ abstract class PathAssertionsSpec(
     describeFun(startsWith.name, startsNotWith.name) {
```
robstoll: include endsWith and endsNotWith

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
```diff
@@ -90,41 +90,38 @@ abstract class PathAssertionsSpec(
...
         val startsWithFun = startsWith.lambda
         val startsNotWithFun = startsNotWith.lambda
-        context("starts with") {
-            it("${startsWith.name} does not throw") {
+
+        context("path '/some/path/for/test'") {
+            it("${startsWith.name} '/some/path/' does not throw") {
                 expect(Paths.get("/some/path/for/test"))
```
robstoll: `expect(Paths.get("/some/path/for/test"))` is always the same for the whole context (makes sense, as this is the given context). Therefore I would recommend to extract this to a local variable, you can name it `expectPath` for instance

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathAssertionsSpec.kt
segunfamisa: okay. 👍

robstoll:
@segunfamisa that was quick, LGTM now, I'll wait what CI has to say and merge then.
Already thank you for the contribution :+1: 
I saw on slack that you are an android dev. Maybe #167 would be good issue to continue with. It requires a bit of gradle knowledge, I can help out there if needed

Issue #207, opened 2019-10-11
Create sample project Atrium + jasmine


robstoll:
*Platform* (jvm, js, android): js
*Extension* (none, kotlin 1.3, jdk8): none

## Code-Documentation related feature
It would be nice if we have a working example similar to the one we created for Atrium + mocha (see samples/js/mocha) but this time with Atrium + jasmine.

The example project shall show how to setup a JS project using jasmine as test runner framework + Atrium and contain a simple test (can be the same as for mocha).

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime,
  we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](http://slack.kotlinlang.org/) in case you do not have an account yet).

bsemexan:
submitted PR #208 for this

PR #208, opened 2019-10-12
sample project Atrium + jasmine #207

bsemexan:
----


robstoll:
@bsemexan please include the jasmine module into the project (see settings.gradle in the root of the project) and see if it actually builds

bsemexan:
it did build, I pushed from within js folder so I forgot the settings file

Issue #116, opened 2019-08-12
Migrate Spek1 specs to Spek2


robstoll:
*Platform* (jvm, jdk8, js, android): jvm

## Test-Code related issue
We have not yet migrated all specs to spek2.
Following a list modules in which there are still spek1 specs
- atrium-core-robstoll-lib-jvm
- atrium-domain-robstoll-lib-jvm

Search for `import org.jetbrains.spek.api.Spek`and for `TODO #116` to find those specs

*TODO*:
- copy the spec from misc/atrium-spec to misc/specs/atrium-specs-common or move it
  from misc/atrium-specs-jvm to misc/atrium-specs-common
 
- migrate the spec, which most of the time involves:
  - search for `import org.jetbrains.spek.api.Spek` replace with `import org.spekframework.spek2.Spek`
  - search for `import org.jetbrains.spek.api.include` replace with `import ch.tutteli.atrium.specs.include`
  - search for `import ch.tutteli.atrium.verbs.internal.AssertionVerbFactory` replace with `import ch.tutteli.atrium.api.verbs.internal.AssertionVerbFactory`
  - search for `import ch.tutteli.atrium.spec.` replace with `import ch.tutteli.atrium.specs`
  - search for `test(` replace with `it(`
  - search for `group(` replace with `context(`
  - search for `on(` repalce with `context(`
  - search for `verbs.checkImmediately` replace with `verbs.check`
  - rewrite `describeFun` => see https://github.com/robstoll/atrium/pull/209/files#diff-676c33efc276af1070bfde5d55307d9dR23
  - plus some further adjustments like import `describe`

See https://spekframework.org/migration/ for additional hints on the topic

## Your first contribution?
- Write a comment `I'll work on ...` if you would like to migrate one of the specs 
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](http://slack.kotlinlang.org/) in case you do not have an account yet).

monst12:
Hey @robstoll, I would love working on this.

robstoll:
@monst12 nice :+1: I suggest you start off by migrating one spek and create a PR for it before you move on to the next one.

robstoll:
@monst12 do you still work on this issue?

[...]

PR #209, opened 2019-10-12
Migrate DelegatingAssertionCheckerSpec.kt to spek2

monst12:
As suggested I started migrating some few specs in misc/specs/atrium-specs-jvm and their connected classes.

Regarding the comment 
> TODO #116 migrate spek1 to spek2 - move to specs-common

inside some classes and the task of moving the specs into specs-common, I ran into a problem:
If I refactor (move) a JVM-specific spec into the common module, the Kotlin stdlib is missing. 
So I'm unsure whether I should only move non Kotlin dependent specs being to specs-common (and I just ignore "move to specs-common") or specs-common should depend on Kotlin-stdlib?
This question also applies to specs in atrium-domain-robstoll-lib-jvm and atrium-core-robstoll-lib-jvm (e.g. `AssertionFormatterControllerSpec ` in `atrium-core-robstoll-lib-jvm.ch.tutteli.atrium.core.robstoll.lib.reporting`).

Furthermore I had to add a new extension function (`Root.include` in `SpecBodyExtensions.kt`) to provide the same syntax in `TextIndentBasedAssertionGroupFormatterSpec`. 

I would greatly appreciate some help and feedback.

----

### DIFF
--- a/misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/checking/DelegatingAssertionCheckerSpec.kt
+++ b/misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/checking/DelegatingAssertionCheckerSpec.kt
@@ -10,9 +10,7 @@ import ch.tutteli.atrium.creating.AssertionHolder
 import ch.tutteli.atrium.creating.AssertionPlant
 import ch.tutteli.atrium.specs.AssertionVerb
 import ch.tutteli.atrium.specs.describeFunTemplate
-import io.mockk.slot
-import io.mockk.spyk
-import io.mockk.verify
+import io.mockk.*
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.Suite
 
@@ -40,7 +38,7 @@ abstract class DelegatingAssertionCheckerSpec(
     describeFun("check") {
         context("empty assertion list") {
             it("does not throw an exception") {
-                val testee = testeeFactory(spyk())
+                val testee = testeeFactory(mockk(relaxed = true))
                 testee.check(assertionVerb, 1, listOf())
             }
         }
@@ -54,7 +52,9 @@ abstract class DelegatingAssertionCheckerSpec(
             context(description) {
                 it("adds the assertion(s) to the subject plant") {
                     //arrange
-                    val subjectFactory = spyk<AssertionPlant<Int>>()
+                    val subjectFactory = mockk<AssertionPlant<Int>>()
+                    every { subjectFactory.addAssertion(any()) } returns subjectFactory
+
                     val testee = testeeFactory(subjectFactory)
                     //act
                     testee.check(assertionVerb, 1, assertions)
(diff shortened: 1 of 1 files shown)


misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/SpecBodyExtensions.kt
```diff
@@ -24,3 +26,6 @@ fun SpecBody.prefixedDescribe(prefix: String, description: String, body: SpecBod
 
 fun SpecBody.prefixedDescribe(prefix: String, suffix: String, description: String, body: SpecBody.() -> Unit) =
     group("${prefix}describe$suffix $description", body = body)
+
+//Tried copying Spek 1 wrap function
+fun Root.include(root: Root.() -> Unit) = root(this)
```
robstoll: td;lr please remove

I had added this functionality myself but pushed it back to spek in the meantime. Moreover #193 removed exactly this function and rewrote the code so that we are using spek's include.

robstoll:
specs-common is depending on kotlin-stdlib-common so the std-lib shouldn't be missing. Is intelliJ telling you this? How are you moving files?

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/AssertionFormatterSpecBase.kt
```diff
@@ -5,14 +5,14 @@ import ch.tutteli.atrium.core.coreFactory
 import ch.tutteli.atrium.reporting.*
 import ch.tutteli.atrium.reporting.translating.Translator
 import ch.tutteli.atrium.reporting.translating.UsingDefaultTranslator
-import org.jetbrains.spek.api.Spek
-import org.jetbrains.spek.api.dsl.Spec
-import org.jetbrains.spek.api.include
+import ch.tutteli.atrium.specs.include
```
robstoll: please rebase on master, this function does not longer exist in Atrium

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/checking/DelegatingAssertionCheckerSpec.kt
```diff
@@ -56,14 +54,14 @@ abstract class DelegatingAssertionCheckerSpec(
             context(description) {
                 it("adds the assertion(s) to the subject plant") {
                     //arrange
-                    val subjectFactory = mock<AssertionPlant<Int>>()
+                    val subjectFactory = spyk<AssertionPlant<Int>>()
```
robstoll: same here

robstoll:
@monst12 usually I wouldn't merge a PR either marked as Draft or prefix with WIP. Since we talked about via Slack I made the adjustements myself. Looking forwatd to the next PR 🙂

Issue #211, opened 2019-10-14
ReadmeSpec should always use en_GB for number formatting


robstoll:
*Platform* (jvm, js, android): all
*Extension* (none, kotlin 1.3, jdk8): none

## Documentation-Code related feature

The examples in ReadmeSpec are used to populate the examples in the README. 
Since the reporter in the ReadmeSpec is using the default Locale of a developer it can happen that some uses a different number formatting (happened this way in pull request #199, see changes in README.md). 
In order that this does not happen any more, we need to fix the Locale used for ReadmeSpec.

TODO: 
- adjust the reporter used for the ReadmeSpec (/samples/readme-examples/src/main/kotlin/readme/examples/ReadmeReporterFactory.kt)

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime,
  we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

mattyway:
I'll work on this

PR #212, opened 2019-10-14
Force locale to en_GB when building examples for readme

mattyway:
Fixes robstoll/atrium#211

----


robstoll:
@mattyway thanks for the contribution :+1:

PR #214, opened 2019-10-15
Use spek-js instead of dummy implementation #202

### LINKED ISSUE #202: Use spek-js instead of dummy implementation
*Platform* (jvm, jdk8, js, android): 

## Code related feature
We are using Spek to perform multiplatform test. As spek does not JS yet, we used a dummy implementation (just the signatures) so that we still can use Spek for the other platforms. 
This idea was adopted by spek with 2.0.6

Hence we should theoretically be able to use it instead of our dummy implementation. However, it did not work as expected. I have a clue why, your job would be to find out if this is the problem. 

TODO: 
- activate the spek-js dependency in misc/specs/atrium-specs-js/build.gradle
- remove the following folder /misc/specs/atrium-specs-js/src/main/kotlin/org/
- try `./gr build` => I guess it still fails (see https://youtrack.jetbrains.com/issue/KT-33611 for more information)
- ~~turn specs-js into a common-js module (it is currently a UMD module)~~ no longer necessary

## Your first contribution?

wudmer:
closes #202 
----


robstoll:
@wudmer thanks for the contribution :+1:

Issue #218, opened 2019-10-16
update tutteli-spek-extensions to 1.0.1


robstoll:
*Platform* (jvm, js, android): jvm
*Extension* (none, kotlin 1.3, jdk8): jdk8

## Code related feature

- update spekExtensions_versions to 1.0.1 in build.gradle in the root
- fix PathAssertionsSpec and PathFeatureAssertionsSpec
- compile the project with `./gr build` to verify that all problems have been addressed

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime,
  we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

ShradhaSangtani:
I'll work on it

PR #244, opened 2019-10-22
ref #218 update tutteli-spek-extensions to 1.0.1

ShradhaSangtani:
----


apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/PathAssertionsSpec.kt
```diff
@@ -1,6 +1,8 @@
 package ch.tutteli.atrium.api.fluent.en_GB.jdk8
 
 import ch.tutteli.atrium.creating.Expect
+import ch.tutteli.atrium.domain.builders.creating.PathAssertionsBuilder.exists
```
robstoll: please revert

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathFeatureAssertionsSpec.kt
```diff
@@ -51,12 +52,12 @@ abstract class PathFeatureAssertionsSpec(
             it("toBe(folder.parent) holds") {
                 val childFolder = tempFolder.newFolder("child")
                 val parentFolder = childFolder.parent
-                expect(childFolder).parentVal().toBe(parentFolder)
+                toBe(parentFolder,expect(childFolder))
```
robstoll: don't use `AnyAssertionsBuilder.toBe` it was perfectly fine how it was previously. I guess there is a misunderstanding in the goal of the task. You need to change how `tempFolder` is specified and used.

ShradhaSangtani:
@robstoll Getting the error -  **Cannot add extension with name 'kotlin', as there is an extension already registered with that name.** when i try to run the updated project. Can you help me with this?

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathFeatureAssertionsSpec.kt
ShradhaSangtani: So i was getting errors on both tempFolder and toBe. For tempFolder i could see there is another class MemoizedTempFolder. But it does not have a method perTest() or anything similar to that.

robstoll: That's correct, there isn't a perTest or perGroup anymore, this is control via CachingMode of memoized. See https://github.com/robstoll/tutteli-spek-extensions#MemoizedTempFolder for an example how it is used.

robstoll:
Thanks for pointing out the error, that's a bug on master. I have fixed it. Please rebase

PR #266, opened 2019-11-03
Symbolic Link Loop Reporting

### LINKED ISSUE #258: symlink loop detection reporting wrong
*Affected Version*: 0.9.0-alpha
*API* (fluent-en_GB, cc or cc-infix): fluent
*Platform* (jvm, js, android): jvm
*Extension* (none, kotlin 1.3, jdk8): jdk8

## How to reproduce the problem  
on a linux system:
```
sudo mv /tmp /tmp2
sudo ln -s /tmp2 /tmp
```
Run PathAssertionsSpec afterwards -> there will be failures similar to:
```
❗❗ found a symbolic link loop: /tmp2 -> /tmp/.... -> /tmp2
```

*Edit* most of them have more to do with test-setup and not with the detection as such. This was fixed in the tests. However, one remains and here it seems the reporting is wrong. I have put a TODO in the code where I assume could be the error (did not have time to investigate further)

## Describe the bug
A symlink loop is detected even though there isn't any.

## Pre-measurements
~~The tests involving a symlink are currently skipped if a symlink is detected in the chain of Java's temporary directory.~~ => I reduced the problem to the test case `prints an explanation for link loops`

## Regression-test
We could build on macOS in addition where this problem was detected. Maybe we would find some odd behaviour of macOS's filesystem.

jGleitz:
* Fix tests so that they can deal with symbolic links in the path to the temporary directory
 * Fix symlink loop detection logic so that it does not get triggered just because we dereference the same link twice
 * Print more elaborate messages on symlink loops, which are helpful in situations like this one:

```
expect: /tmp2/spek9177372137321997668/sneaky-loop/foo/link        (sun.nio.fs.UnixPath <1131490382>)
◆ to be: readable
    » followed the symbolic link /tmp2/spek9177372137321997668/sneaky-loop/foo/link to /tmp2/spek9177372137321997668/sneaky-loop/bar/link
    » followed the symbolic link /tmp2/spek9177372137321997668/sneaky-loop/bar to /tmp2/spek9177372137321997668/sneaky-loop/foo
      ❗❗ found a symbolic link loop: /tmp2/spek9177372137321997668/sneaky-loop/foo/link -> /tmp2/spek9177372137321997668/sneaky-loop/foo/link
    » access threw a FileSystemException:
      » message: "/tmp2/spek9177372137321997668/sneaky-loop/foo/link: Zu viele Ebenen aus symbolischen Links"        <823596548>
      » stacktrace: 
        ⚬ sun.nio.fs.UnixException.translateToIOException(UnixException.java:91)
        ⚬ …
```

(in the previous implementation, the `❗❗ found a symbolic link loop…` message would have been the first one).
 
Fixes #258 

----

### DIFF
--- a/domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/test/kotlin/ch/tutteli/atrium/assertions/filesystem/SymbolicLinkResolvingSpec.kt
+++ b/domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/test/kotlin/ch/tutteli/atrium/assertions/filesystem/SymbolicLinkResolvingSpec.kt
@@ -215,6 +215,8 @@ object SymbolicLinkResolvingSpec : Spek({
             val resultAssertion = explainForResolvedLink(a, resolvedPathConsumer)
             expect(resultAssertion).isA<AssertionGroup>()
                 .feature { p(it::assertions) }.containsExactly(
+                    { describesLink(a, b) },
+                    { describesLink(b, a) },
                     { describesLinkLoop(a, b, a) },
                     { isSameAs(testAssertion) }
                 )
@@ -229,7 +231,9 @@ object SymbolicLinkResolvingSpec : Spek({
             val resultAssertion = explainForResolvedLink(link, resolvedPathConsumer)
             expect(resultAssertion).isA<AssertionGroup>()
                 .feature { p(it::assertions) }.containsExactly(
-                    { describesLinkLoop(link, foolink, link) },
+                    { describesLink(link, foolink.resolve("link")) },
+                    { describesLink(foolink, foo) },
+                    { describesLinkLoop(link, link) },
                     { isSameAs(testAssertion) }
                 )
         }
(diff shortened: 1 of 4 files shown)


domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/test/kotlin/ch/tutteli/atrium/assertions/filesystem/SymbolicLinkResolvingSpec.kt
```diff
@@ -174,41 +188,79 @@ object SymbolicLinkResolvingSpec : Spek({
                         { isSameAs(testAssertion) }
                     )
             }
+
+            it("does not assume a link loop even if the same link appears multiple times") {
+                val testdir = tempFolder.newFolder("multi-non-loop").toRealPath()
+                val barLink = testdir.createSymbolicLink(testdir.resolve("bar"))
+                val target = testdir.resolve("target").createFile()
+                val testLink = barLink.resolve(barLink.fileName).resolve(barLink.fileName).resolve("target")
```
robstoll:
```suggestion
                val testLink = barLink.resolve(barLink.fileName).resolve(barLink.fileName).resolve(target.fileName)
```
Should also fix the build

jGleitz: Ah, thanks!

PR #267, opened 2019-11-09
WIP: Adding new map entry assertion for infix functions

npswedberg:
----


apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/MapEntryAssertionsSpec.kt
```diff
@@ -0,0 +1,21 @@
+package ch.tutteli.atrium.api.infix.en_GB
+
+import ch.tutteli.atrium.creating.Expect
+import ch.tutteli.atrium.specs.fun2
+import ch.tutteli.atrium.specs.notImplemented
+
+class MapEntryAssertionsSpec : ch.tutteli.atrium.specs.integration.MapEntryAssertionsSpec(
+    fun2(Expect<Map.Entry<String, Int>>::isKeyValue),
```
robstoll: Note, the number after `fun` designated the arity of the function. `isKeyValue` in the fluent-API requires two arguments, key and value, thus we used `fun2`. However, your infix function only has 1 argument, thus you have to use `fun1`. Moreover, we want to show the usage of isKeyValue in the companion object. See PairFeatureAssertionsSpec. Something along the line of:
```
    fun1(Expect<Map.Entry<String, Int>>::isKeyValue).name to Companion::first,
    // ...
) {
    companion object {
      fun isKeyValue(expect: Expect<Map.Entry<String, Int>>, key: String, value: Int) =
        expect isKeyValue (key to value)
      // ...
   }
```

robstoll:
I suggest you create a new branch for this PR (I would create a new branch for each new issue you are working on) and rebase it onto Atrium's master
- git checkout -b \#267-map-entry-new-infix-api
- git remote add upstream https://github.com/robstoll/atrium.git
- git rebase upstream/master

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/MapEntryAssertionsSpec.kt
npswedberg: Oh this makes so much more sense

robstoll:
Closing this PR as it contains changes from an older PR. Going to copy paste the things you have already done and mention you as author in the commit

PR #277, opened 2019-12-08
Add workarounds and changes to CONTRIBUTING.md

lukebiddell:
My first pull request, let me know if there's any issues with it.

----

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -82,6 +82,12 @@ Following a small guidance how we suggest to start.
    This builds the project (compile, run tests etc) via gradle (notice, we renamed `gradlew` to `gr`, more convenient no?) 
 5. Read up the [Coding Conventions of Atrium](#coding-conventions) (there are only 5 points).
 
+Note: IntelliJ requires a few workarounds.
+- IntelliJ may warn you about duplicate content roots.
+  Simply ignore it (IntelliJ doesn't support it).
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -76,12 +76,18 @@ Following a small guidance how we suggest to start.
    We suggest [IntelliJ Community Edition](https://www.jetbrains.com/idea/download/)
    and install the following plugins:
    - EditorConfig by JetBrains (might be already activated)
-   - Spek Framework by Spek Team 
+   - Spek by Ranie Jade Ramiso
```
robstoll: That's the plugin for spek 1.x and usually new contributors should only use spek 2.x 
Thus I am not sure if it makes sense to mention it as well but if you think yes, then please add both as follows:
```suggestion
   - Spek Framework by Spek Team (for Spek 2.x)
   - Spek by Ranie Jade Ramiso (for Spek 1.x)
```

lukebiddell: Strangely I couldn't find Spek Framework in the plugin market when I originally searched for it, but I see it now. So I think we should leave it as it was before?

robstoll: Most likely because spek wasn't yet (or still isn't) available for intellij 2019.3

robstoll:
@lukebiddell thanks for the improvement :+1:

PR #296, opened 2020-01-09
Improve CONTRIBUTION.md

johnGachihi:
How is this?

----

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contributing to Atrium
 
-Thank you very much that you take your time to contribute to Atrium :smiley:
+Thank you very much for taking your time to contribute to Atrium :smiley:
 
 Following a few guidelines so that others can quickly benefit from your contribution.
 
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -92,10 +92,22 @@ Perfect, you are setup and ready to go.
...
 where [good first issues](https://github.com/robstoll/atrium/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22)
 are easier to start with.
-Please write a comment such like `I am working on this` into the issue, 
+Please write a comment such as `I am working on this` in the issue,
 this way we can assign the task to you (so that others know there is already someone working on the issue)
 and it gives us the chance to have a look at the description again and revise if necessary.
 
+When working on an issue, we recommend you do the following on git:
+- create a new branch for the issue using `git checkout -b <branch-name>` (preferrably, the branch name
+  should be descriptive of the issue or the change being made, e.g `#200-add-feature`.) Working
```
jGleitz: Maybe `#200-add-feature` is a bad example because somebody could take it literally. We could change it to something like `#108-path-exists`.

robstoll:
@jGleitz the reason for this was that @johnGachihi had troubles with git during his first PR. I agree that a link to a good tutorial would maybe be better for newcomers. Do you know one by chance? On the other hand it is quite a small list, so maybe we keep it and add a link in addition. I would also suggest that we add a remark -- something like `Dealing the first time with git?` -- so that people knowing git can skip these recommendations

jGleitz:
I like https://git-scm.com/book/. It is clear and comprehensive. The downside may be that is intended to give in-depth understanding. There are certainly other tutorials that will get you going faster.

I like the solution with giving a very rough overview like it’s done in this PR and additionally adding a link to a manual. Using a heading like “Dealing the first time with git?” is also a good idea.

robstoll:
@johnGachihi thanks, also for the textual improvements. 
I have sent you an invitation as collaborator as you are currently contribution a lot :tada: 
This way you can assign tasks to you yourself without the need to write `I'll work on it`. Feel free to do it nonetheless as I normally have a look at the description if it is still correct in such cases.

PR #301, opened 2020-01-10
show only java's qualified name if kotlin type differ from java type

Hubisco:
----

### DIFF
--- a/core/robstoll-lib/atrium-core-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/DetailedObjectFormatter.kt
+++ b/core/robstoll-lib/atrium-core-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/DetailedObjectFormatter.kt
@@ -44,7 +44,7 @@ actual class DetailedObjectFormatter actual constructor(
      * - [StringBasedRawString] is represented as [StringBasedRawString.string]
      * - [TranslatableBasedRawString] is represented as result of its translation (by [translator])
      * - [Class] is represented as "[Class.getSimpleName] ([Class.getName])"
-     * - [KClass] is represented as "[Class.getSimpleName] ([Class.getName])"
+     * - [KClass] is represented as "[Class.getName]"
      * - [Enum] is represented as "[toString] ([Class.getName])
      * - [Throwable] is represented as "[Class.getName]"
      * - All other objects are represented as "[toString] ([Class.getName] <[System.identityHashCode]>)"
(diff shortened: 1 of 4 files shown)


Hubisco:
Hi, any idea why test coverage is decreased by 4.39% ?

core/robstoll-lib/atrium-core-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/DetailedObjectFormatter.kt
```diff
@@ -44,7 +44,7 @@ actual class DetailedObjectFormatter actual constructor(
      * - [StringBasedRawString] is represented as [StringBasedRawString.string]
      * - [TranslatableBasedRawString] is represented as result of its translation (by [translator])
      * - [Class] is represented as "[Class.getSimpleName] ([Class.getName])"
-     * - [KClass] is represented as "[Class.getSimpleName] ([Class.getName])"
+     * - [KClass] is represented as "[Class.getName]"
```
robstoll: That's not correct but the previous is neither. Change it to:
```suggestion
     * - [KClass] is represented as "[KClass.simpleName] ([KClass.qualifiedName])" unless the [KClass.qualifiedName] 
     *   differs from [Class.getName] in which case, "-- Class: [Class.getName]" is appended in addition
```

robstoll:
@Hubisco Let's ignore codecov, sometimes it is wrong; the PR looks good besides the small suggestion I wrote.

gradle/package-lock.json
```diff
@@ -122,9 +122,9 @@
       },
       "dependencies": {
         "glob": {
-          "version": "7.1.4",
-          "resolved": "https://registry.npmjs.org/glob/-/glob-7.1.4.tgz",
-          "integrity": "sha512-hkLPepehmnKk41pUGm3sYxoFs/umurYfYJCerbXEyFIWcAzvpipAgVkBqqT9RBKMGjnq6kMuyYwha6csxbiM1A==",
+          "version": "7.1.6",
```
robstoll: please revert this change

Hubisco: done but are You sure about that ? this is generated automatically after 'gr build'

Hubisco:
@robstoll Hi, I run `./gr build` previously and nothing relevant happened, there were some change in paths (I am using windows)  so I ignore them. One line is missing in README.md after `./gr build` so I added it in this PR.

robstoll:
@Hubisco does `gr :readme-examples:build` succeed or fail?

gradle/package-lock.json
robstoll: does not happen to me, I guess this change was caused by your (or mine) local npm/node setup

robstoll:
seems like `String` and `Integer` is also a primitive in the eyes of Kotlin :slightly_frowning_face:  and that's the reason why the readme does not change. Do you want to have a look into it? its one line above your change in DetailedObjectFormatter

robstoll:
@Hubisco thanks for your contribution :slightly_smiling_face: :+1: 

This one was quite a small issue, fancy another one?
https://github.com/robstoll/atrium/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22help+wanted%22

(and don't forget to star Atrium :wink:)

Hubisco:
@robstoll No problem, sure, I will try to take another one :)

PR #311, opened 2020-01-13
Change toBe and notToBe translations for en_GB

Hubisco:
----


translations/en_GB/atrium-translations-en_GB-common/src/main/kotlin/ch/tutteli/atrium/translations/DescriptionBasic.kt
```diff
@@ -10,8 +10,8 @@ import ch.tutteli.atrium.reporting.translating.StringBasedTranslatable
 enum class DescriptionBasic(override val value: String) : StringBasedTranslatable {
     TO("to"),
     NOT_TO("not to"),
-    TO_BE("to be"),
-    NOT_TO_BE("not to be"),
+    TO_BE("is equal to"),
```
robstoll: Sorry, I just realised that there are functions on some types which are called `isEqual` and this text would be misleading in such cases as one could no longer distinguish between `equals` and `isEqual`. Can you please change it to `equals` and `does not equal`

Hubisco: sure no problem

robstoll:
thanks for the contribution :+1:

PR #322, opened 2020-01-17
Add Optional assertion

slalu:
This is a first draft and I am unsure about my implementation, but I would need some help to be able to continue. I am particularly unsure about the Spec part. I started an OptionalFeatureAssertionsSpec, but then I noticed there was already OptionalAssertionsSpec, so I modified thos classes instead. Now I'm stuck with OptionalAssertionsSpec that doesn't compile (still learning Kotlin syntax as well as this project).

Thanks! 

----

### DIFF
--- a/apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/optionalAssertions.kt
+++ b/apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/optionalAssertions.kt
@@ -19,3 +19,29 @@ import java.util.*
  * @since 0.9.0
  */
 fun <T : Optional<*>> Expect<T>.isEmpty(): Expect<T> = addAssertion(ExpectImpl.optional.isEmpty(this))
+
+/**
+ * Expects that the subject of the assertion (an [Optional]) is present
+ * and returns an [Expect] for the inner type [E].
+ *
+ * Shortcut for more or less something like `feature(Optional<T>::get)` but with error handling; yet it
+ * depends on the underlying implementation though.
+ *
+ * @return The newly created [Expect] if the given assertion is success
+ * @throws AssertionError Might throw an [AssertionError] if the given assertion is not a success.
+ *
+ * @since 0.9.0
+ */
+fun <E, T : Optional<E>> Expect<T>.isPresent(): Expect<E> = ExpectImpl.optional.isPresent(this).getExpectOfFeature()
+
+/**
+ * Expects that the subject of the assertion (an [Optional]) is present and
+ * that it holds all assertions the given [assertionCreator] creates.
+ *
+ * @return This assertion container to support a fluent API.
+ * @throws AssertionError Might throw an [AssertionError] if the given assertions are not success.
+ *
+ * @since 0.9.0
+ */
+fun <E, T : Optional<E>> Expect<T>.isPresent(assertionCreator: Expect<E>.() -> Unit): Expect<T> =
+    ExpectImpl.optional.isPresent(this).addToInitial(assertionCreator)
--- a/apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/OptionalAssertionsSpec.kt
+++ b/apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/OptionalAssertionsSpec.kt
@@ -1,13 +1,14 @@
 package ch.tutteli.atrium.api.fluent.en_GB.jdk8
 
 import ch.tutteli.atrium.creating.Expect
-import ch.tutteli.atrium.specs.fun0
+import ch.tutteli.atrium.specs.*
 import ch.tutteli.atrium.specs.integration.OptionalAssertionsSpec
-import ch.tutteli.atrium.specs.notImplemented
 import java.util.*
 
 class OptionalAssertionsSpec : OptionalAssertionsSpec(
-    isEmpty = fun0(Expect<Optional<Any>>::isEmpty)
+    isEmpty = fun0(Expect<Optional<Int>>::isEmpty),
+    isPresentFeature = feature0<Optional<Int>, Int>(Expect<Optional<Int>>::isPresent),
+    isPresent = fun1<Optional<Int>, Expect<Int>.() -> Unit>(Expect<Optional<Int>>::isPresent)
 ) {
     @Suppress("unused", "UNUSED_VALUE")
     private fun ambiguityTest() {
--- a/domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/optionalAssertions.kt
+++ b/domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/optionalAssertions.kt
@@ -3,12 +3,24 @@
 package ch.tutteli.atrium.domain.robstoll.lib.creating
 
 import ch.tutteli.atrium.assertions.Assertion
+import ch.tutteli.atrium.core.Option
 import ch.tutteli.atrium.creating.Expect
 import ch.tutteli.atrium.domain.builders.ExpectImpl
+import ch.tutteli.atrium.domain.creating.changers.ExtractedFeaturePostStep
 import ch.tutteli.atrium.reporting.RawString
 import ch.tutteli.atrium.translations.DescriptionBasic.IS
-import ch.tutteli.atrium.translations.DescriptionOptionalAssertion.EMPTY
+import ch.tutteli.atrium.translations.DescriptionOptionalAssertion
 import java.util.*
 
 fun <T : Optional<*>> _isEmpty(expect: Expect<T>): Assertion =
-    ExpectImpl.builder.createDescriptive(expect, IS, RawString.create(EMPTY)) { !it.isPresent }
+    ExpectImpl.builder.createDescriptive(expect, IS, RawString.create(DescriptionOptionalAssertion.EMPTY)) { !it.isPresent }
+
+fun <E, T : Optional<E>> _isPresent(expect: Expect<T>): ExtractedFeaturePostStep<T, E> =
+    ExpectImpl.feature.extractor(expect)
+        .withDescription(DescriptionOptionalAssertion.GET)
+        .withRepresentationForFailure(DescriptionOptionalAssertion.IS_NOT_PRESENT)
+        .withFeatureExtraction {
+            Option.someIf(it.isPresent) { it.get() }
+        }
+        .withoutOptions()
+        .build()
--- a/misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
+++ b/misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
@@ -1,43 +1,66 @@
 package ch.tutteli.atrium.specs.integration
 
 import ch.tutteli.atrium.api.fluent.en_GB.messageContains
+import ch.tutteli.atrium.api.fluent.en_GB.toBe
 import ch.tutteli.atrium.api.fluent.en_GB.toThrow
 import ch.tutteli.atrium.api.verbs.internal.expect
+import ch.tutteli.atrium.creating.Expect
 import ch.tutteli.atrium.specs.*
-import ch.tutteli.atrium.translations.DescriptionOptionalAssertion.EMPTY
+import ch.tutteli.atrium.translations.DescriptionOptionalAssertion
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.Suite
 import java.lang.AssertionError
 import java.util.Optional
 
-abstract class OptionalAssertionsSpec (
-    isEmpty: Fun0<Optional<Any>>,
+abstract class OptionalAssertionsSpec(
+    isEmpty: Fun0<Optional<Int>>,
+    isPresentFeature: Feature0<Optional<Int>, Int>,
+    isPresent: Fun1<Optional<Int>, Expect<Int>.() -> Unit>,
     describePrefix: String = "[Atrium] "
 ) : Spek({
 
-    include(object : SubjectLessSpec<Optional<Any>>(
+    include(object : SubjectLessSpec<Optional<Int>>(
         "$describePrefix[Path] ",
-        isEmpty.forSubjectLess()
+        isEmpty.forSubjectLess(),
+        isPresentFeature.forSubjectLess().adjustName { "$it feature" },
+        isPresent.forSubjectLess { toBe(1) }
+    ) {})
+    include(object : AssertionCreatorSpec<Optional<Int>>(
+        describePrefix, Optional.of(2),
+        isPresent.forAssertionCreatorSpec("$toBeDescr: 2") { toBe(2) }
     ) {})
 
     fun describeFun(vararg funName: String, body: Suite.() -> Unit) =
         describeFunTemplate(describePrefix, funName, body = body)
 
-    describeFun(isEmpty.name) {
+    describeFun(isEmpty.name, isPresent.name) {
         val isEmptyFun = isEmpty.lambda
+        val isPresentFun = isPresentFeature.lambda
 
-        context("is empty") {
-            it ("does not throw") {
-                val emptyValue = Optional.empty<Any>()
+        val emptyValue = Optional.empty<Int>()
+        context("$emptyValue") {
+            it("${isEmpty.name} - does not throw") {
                 expect(emptyValue).isEmptyFun()
             }
-
-            it ("throws an AssertionError") {
-                val nonEmptyValue: Optional<Any> = Optional.of(2)
+            it("${isPresent.name} - throws an AssertionError") {
                 expect {
-                    expect(nonEmptyValue).isEmptyFun()
+                    expect(emptyValue).isPresentFun()
                 }.toThrow<AssertionError> {
-                    messageContains("$isDescr: ${EMPTY.getDefault()}")
+                    messageContains(DescriptionOptionalAssertion.IS_NOT_PRESENT.getDefault())
+                }
+            }
+        }
+
+        val presentValue: Optional<Int> = Optional.of(2)
+        context("$presentValue") {
+            it("${isPresent.name} - can perform sub-assertion which holds") {
+                expect(presentValue).isPresentFun()
+            }
+            it("${isEmpty.name} - throws an AssertionError") {
+                expect {
+                    expect(presentValue).isEmptyFun()
+                }.toThrow<AssertionError> {
+                    messageContains("$isDescr: ${DescriptionOptionalAssertion.EMPTY.getDefault()}")
                 }
             }
         }
--- a/translations/en_GB/atrium-translations-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/translations/DescriptionOptionalAssertion.kt
+++ b/translations/en_GB/atrium-translations-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/translations/DescriptionOptionalAssertion.kt
@@ -6,4 +6,6 @@ import ch.tutteli.atrium.reporting.translating.StringBasedTranslatable
 
 enum class DescriptionOptionalAssertion(override val value: String) : StringBasedTranslatable {
     EMPTY("empty"),
+    GET("get()"),
+    IS_NOT_PRESENT("!! is not present")
 }
--- a/domain/api/atrium-domain-api-jvm/src/main/kotlin/ch/tutteli/atrium/domain/creating/OptionalAssertions.kt
+++ b/domain/api/atrium-domain-api-jvm/src/main/kotlin/ch/tutteli/atrium/domain/creating/OptionalAssertions.kt
@@ -5,6 +5,7 @@ package ch.tutteli.atrium.domain.creating
 import ch.tutteli.atrium.assertions.Assertion
 import ch.tutteli.atrium.core.polyfills.loadSingleService
 import ch.tutteli.atrium.creating.Expect
+import ch.tutteli.atrium.domain.creating.changers.ExtractedFeaturePostStep
 import java.util.*
 
 /**
@@ -20,4 +21,5 @@ val optionalAssertions by lazy { loadSingleService(OptionalAssertions::class) }
  */
 interface OptionalAssertions {
     fun <T : Optional<*>> isEmpty(expect: Expect<T>): Assertion
+    fun <E, T : Optional<E>> isPresent(expect: Expect<T>): ExtractedFeaturePostStep<T, E>
 }
--- a/domain/builders/atrium-domain-builders-jvm/src/main/kotlin/ch/tutteli/atrium/domain/builders/creating/OptionalAssertionsBuilder.kt
+++ b/domain/builders/atrium-domain-builders-jvm/src/main/kotlin/ch/tutteli/atrium/domain/builders/creating/OptionalAssertionsBuilder.kt
@@ -9,6 +9,7 @@ package ch.tutteli.atrium.domain.builders.creating
 import ch.tutteli.atrium.core.polyfills.loadSingleService
 import ch.tutteli.atrium.creating.Expect
 import ch.tutteli.atrium.domain.creating.OptionalAssertions
+import ch.tutteli.atrium.domain.creating.changers.ExtractedFeaturePostStep
 import ch.tutteli.atrium.domain.creating.optionalAssertions
 import java.util.*
 
@@ -21,4 +22,7 @@ object OptionalAssertionsBuilder : OptionalAssertions {
 
     override inline fun <T : Optional<*>> isEmpty(expect: Expect<T>) =
         optionalAssertions.isEmpty(expect)
+
+    override inline fun <E, T : Optional<E>> isPresent(expect: Expect<T>): ExtractedFeaturePostStep<T, E> =
+        optionalAssertions.isPresent(expect)
 }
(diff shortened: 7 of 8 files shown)


robstoll:
Thanks, I'll have a look later on

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/optionalAssertions.kt
```diff
@@ -19,3 +19,17 @@ import java.util.*
  * @since 0.9.0
  */
 fun <T : Optional<*>> Expect<T>.isEmpty(): Expect<T> = addAssertion(ExpectImpl.optional.isEmpty(this))
+
+/**
+ * Expects that the subject of the assertion (an [Optional]) is present.
+ * and returns an [Expect] for the inner type [E].
+ *
+ * Shortcut for more or less something like `feature(Optional<T>::get) { toBe(true) }`
```
robstoll:
```suggestion
 * Shortcut for more or less something like `feature(Optional<T>::get)` but with error handling; yet it
```

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/optionalAssertions.kt
```diff
@@ -19,3 +19,17 @@ import java.util.*
...
+ * depends on the underlying implementation though.
+ *
+ * @return The newly created [Expect] if the given assertion is success
+ * @throws AssertionError Might throw an [AssertionError] if the given assertion is not a success.
+ *
+ * @since 0.9.0
+ */
+fun <E, T : Optional<E>> Expect<T>.isPresent(): Expect<E> = ExpectImpl.optional.isPresent(this).getExpectOfFeature()
```
robstoll: don't forget too add an overload which expects an assertionCreator (see resultAssertions in  api.fluent.en_GB.kotlin_1_3)

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/OptionalAssertionsSpec.kt
```diff
@@ -1,13 +1,13 @@
...
-import ch.tutteli.atrium.specs.fun0
+import ch.tutteli.atrium.specs.*
 import ch.tutteli.atrium.specs.integration.OptionalAssertionsSpec
-import ch.tutteli.atrium.specs.notImplemented
 import java.util.*
 
 class OptionalAssertionsSpec : OptionalAssertionsSpec(
-    isEmpty = fun0(Expect<Optional<Any>>::isEmpty)
+    isEmpty = fun0(Expect<Optional<Any>>::isEmpty),
+    isPresent =fun1<Optional<Any>, Any>(Expect<Optional<Any>>::isPresent)
```
robstoll: I suggest you use `Int` instead of `Any` so that you can make assertions about an `Int` in your specs

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/optionalAssertions.kt
```diff
@@ -3,12 +3,25 @@
...
 import ch.tutteli.atrium.translations.DescriptionBasic.IS
 import ch.tutteli.atrium.translations.DescriptionOptionalAssertion.EMPTY
+import ch.tutteli.atrium.translations.DescriptionResultAssertion
 import java.util.*
 
 fun <T : Optional<*>> _isEmpty(expect: Expect<T>): Assertion =
     ExpectImpl.builder.createDescriptive(expect, IS, RawString.create(EMPTY)) { !it.isPresent }
+
+fun <E, T : Optional<E>> _isPresent(expect: Expect<T>): ExtractedFeaturePostStep<T, E> =
+    ExpectImpl.feature.extractor(expect)
```
robstoll: The code is fine but the corresponding report in case of an error is not yet correct. See hint in the Spec

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
```diff
@@ -27,12 +28,12 @@ abstract class OptionalAssertionsSpec (
         val isEmptyFun = isEmpty.lambda
```
robstoll: I suggest you add the isPresent.name to this `describeFun` (as second argument).

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
```diff
@@ -27,12 +28,12 @@ abstract class OptionalAssertionsSpec (
...
 
         context("is empty") {
-            it ("does not throw") {
+            it("does not throw") {
                 val emptyValue = Optional.empty<Any>()
                 expect(emptyValue).isEmptyFun()
             }
 
-            it ("throws an AssertionError") {
+            it("throws an AssertionError") {
```
robstoll: That's not your fault and I must have overlooked it in the review when isEmpty was added. This belongs into an own `context("is present and 2")`. Please change that. Once you have done this, you can add your use case to this context and to the context above (no need to repeat this again). See ResultFeatureAssertionsSpec line 76, both functions isSucessFeature and isFailureFeature are handled in the same context.

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
```diff
@@ -42,4 +43,24 @@ abstract class OptionalAssertionsSpec (
             }
         }
     }
+
+    describeFun(isPresent.name) {
+        val isPresentFun = isPresent.lambda
+
+        context("is present") {
+            it("does not throw") {
+                val isPresentValue: Optional<Any> = Optional.of(2)
```
robstoll: I prefer your name `isPresentValue` over the existing `nonEmptyValue` so I suggest you use your name when you move this above.

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
```diff
@@ -42,4 +43,24 @@ abstract class OptionalAssertionsSpec (
...
+                expect(isPresentValue).isPresentFun(2)
+            }
+
+            it("throws an AssertionError") {
+                expect {
```
robstoll: I suggest you comment this line and line 60-62 and let the spec run. This way you can see how the reporting looks like.
I am sure you will see why the current implementation is not yet as it should be.

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
```diff
@@ -42,4 +43,24 @@ abstract class OptionalAssertionsSpec (
             }
         }
     }
+
+    describeFun(isPresent.name) {
+        val isPresentFun = isPresent.lambda
+
+        context("is present") {
+            it("does not throw") {
```
robstoll: see ResultFeatureAssertionsSpec line 77, I suggest you do here the same, so change it to:
```suggestion
            it("${isPresent.name} - can perform sub-assertion which holds") {
```
moreover, I would add a second use case for 
```
it("${isPresent.name} - can perform sub-assertion which fails, throws AssertionError") {

File: misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt [hunk 10]
```diff
@@ -10,8 +10,9 @@ import org.spekframework.spek2.style.specification.Suite
 import java.lang.AssertionError
 import java.util.Optional
 
-abstract class OptionalAssertionsSpec (
+abstract class OptionalAssertionsSpec(
     isEmpty: Fun0<Optional<Any>>,
+    isPresent: Fun1<Optional<Any>, Any>,
```
robstoll: That's probably one of the reasons why your spec does not compile. You want a `Feature0` and not a `Fun1`. This is Atrium specific and thus a little explanation, `FunX` means simplified that you want to test an assertion function which expects X arguments (`Fun1` expects one argument). Your function expects 0, so that's one mistake. Moreover, a `FunX` has the same return type as receiver type (`Optional<Any>` in your example). That's wrong for your case as well, your return type changes to `Any`. Therefore you need to use Feature0 => see ResultFeatureAssertionsSpec

Moreover, you should rename isPresent to `isPresentFeature` (corresponds to `isSuccessFeature` in ResultFeatureAssertionsSpec) and `isPresent` analogous to `isSuccess` in ResultFeatureAssertionsSpec

robstoll:
I have left a a few review comments and hope that helps (and can give more hints if needed but thought too much is maybe not good if you intend to learn something).

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
```diff
@@ -10,8 +10,9 @@ import org.spekframework.spek2.style.specification.Suite
...
     describePrefix: String = "[Atrium] "
 ) : Spek({
```
robstoll: Please add your functions to SubjectLessSpec as well (see ResultFeatureAssertionsSpec) + add AssertionCreatorSpec (I think you can copy most of the code from ResultFeatureAssertionsSpec)

slalu:
> I see that you have unresolved a few conversations but I cannot see the corresponding changes. Have you forgotten to push them?

Sorry I marked as resolved while I was applying the changes locally. I want to complete the implementation before pushing it. It compiles now I just need to fix the implementation and the tests.

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
slalu: I'm not sure to understand this one. My understanding by looking at ResultFeatureAssertionsSpec is that each  function should have their own context. This one was to validate that isEmpty on a nonempty value throws an assertion error. I will remove the duplicate describeFun in my next push. Let me know if it makes more sense once they have their own context.

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
slalu: I think both val make sense depending on the test. Let me know if you still want to use the same in the upcoming code.

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/optionalAssertions.kt
slalu: I think I'm getting closer. The report produced by the test made more sense now. I'm still unsure about the Description

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
```diff
@@ -1,45 +1,85 @@
...
+        describePrefix,
+        isPresentNullableFeature.forSubjectLess().adjustName { "$it feature" },
+        isPresentNullable.forSubjectLess { toBe(1) }
+    ) {})
+    include(object : AssertionCreatorSpec<Optional<Int>>(
+        describePrefix, Optional.of(2),
+        isPresent.forAssertionCreatorSpec("$toBeDescr: 2") { toBe(2) }
+    ) {})
+    include(object : AssertionCreatorSpec<Optional<Int?>>(
+        describePrefix, Optional.of(2) as Optional<Int?>,
```
slalu: I am getting a warning here. Is it the right way to have an Optional of a nullable Int?

misc/verbs/atrium-verbs-common/src/main/kotlin/ch/tutteli/atrium/api.verbs/AssertionVerb.kt
```diff
@@ -7,7 +7,7 @@ import ch.tutteli.atrium.reporting.translating.Translatable
  * The [Translatable]s for the assertion functions [assert], [assertThat] and [expect].
  */
 enum class AssertionVerb(override val value: String) : StringBasedTranslatable {
-    ASSERT("assert"),
-    ASSERT_THAT("assert that"),
-    EXPECT("expect"),
+    ASSERT("asserted that subject"),
+    ASSERT_THAT("asserted that subject"),
+    EXPECT("expected that subject"),
```
slalu: There are things I didn't modify. Is it changes coming from the master?

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
```diff
@@ -1,45 +1,85 @@
...
+                    val emptyValue = Optional.empty<Int>()
+                    expect(emptyValue).isPresentFun()
+                }.toThrow<AssertionError> {
+                    messageContains(DescriptionOptionalAssertion.PRESENT.getDefault())
                 }
             }
         }
     }
+
+    //TODO working on the remaining tests for features and nullables
```
slalu: I will work on those once you confirm I have the right approach for the tests.

misc/verbs/atrium-verbs-common/src/main/kotlin/ch/tutteli/atrium/api.verbs/AssertionVerb.kt
robstoll: Yes, those are changes from master which you have re-applied on your branch and are thus seen as changes. You don't get this if you rebase on master (for a guideline for your next PR see https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#git). You can do the following to get rid of the commits not belonging to you (please commit all your local changes first):
- `git checkout -b t` to have a backup branch pointing at your current state in case you mess something up. This way you can always to a `git reset --hard t`. 
- `git checkout 103-get-ispresent`  back to your branch
- `git reset --soft 49638d85` kind of un-commit all changes (expect the first) to the point where you started with your first commit
- `git commit --amend` add all changes to your first commit
- `git remote add upstream https://github.com/robstoll/atrium.git`
- `git fetch upstream`
- `git rebase upstream/master` put your changes on top of Atrium's master
- `git push -f`  force push your commit (will update the PR)
- `git branch -D t` delete the tmp branch as you don't need it any more

I hope that helps

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/optionalAssertions.kt
robstoll: Use `.withDescription("get()")` for the description

translations/en_GB/atrium-translations-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/translations/DescriptionOptionalAssertion.kt
```diff
@@ -6,4 +6,6 @@ import ch.tutteli.atrium.reporting.translating.StringBasedTranslatable
 
 enum class DescriptionOptionalAssertion(override val value: String) : StringBasedTranslatable {
     EMPTY("empty"),
+    VALUE("value"),
+    PRESENT("present")
```
robstoll: see reporting in the issue, this should be used if the Optional is not present. Hence rename it to something like the following with the corresponding text: `IS_NOT_PRESENT("❗❗ is not present")`

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
robstoll: Usually `Optional.of(2)` or `Optional.of<Int?>(2)` should be good enough. But I see my mistake here. I was in the wrong language, the value inside of Optional can never be null as Java has a requireNonNull check and Kotlin know it and thus turns even `Optional.of<Int?>(2)` into an `Optional<Int>`. 
Sorry for that but you can remove isPresentNullableFeature / isPresentNullable entirely

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
robstoll: no need for nullable as outlined above

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
```diff
@@ -1,45 +1,85 @@
...
+            }
+        }
+
+        context("is Present") {
+            it("${isPresent.name} - can perform sub-assertion which holds") {
+                val isPresentValue: Optional<Int> = Optional.of(2)
+                expect(isPresentValue).isPresentFun()
+            }
+
+            it("throws an AssertionError") {
```
robstoll: this test can be moved into the `context("is empty")`

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
robstoll: the context is related to the subject of the assertion and not the function. This was done wrong here and is correct in ResultFeatureAssertionsSpec. Thus my suggestion to wrap the `it("throws an AssertionError") {` into an own `context("is present and 2")` (outside of the `is empty` context). Now that you have an `context("is Preent")` you can move this test down into it

misc/verbs/atrium-verbs-common/src/main/kotlin/ch/tutteli/atrium/api.verbs/AssertionVerb.kt
slalu: Got it. I already configured the upstream. I think I didn't fetch from upstream though. Do I need to explicitly fetch from upstream every time or a regular fetch alredy? Thanks!

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
```diff
@@ -1,43 +1,66 @@
...
+    describeFun(isEmpty.name, isPresent.name) {
         val isEmptyFun = isEmpty.lambda
+        val isPresentFun = isPresentFeature.lambda
 
-        context("is empty") {
-            it ("does not throw") {
-                val emptyValue = Optional.empty<Any>()
+        val emptyValue = Optional.empty<Int>()
+        context("$emptyValue") {
+            it("does not throw") {
```
robstoll:
```suggestion
            it("${isEmpty.name} - does not throw") {
```
(note, you can also apply suggestions directly from the GitHub website)

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/optionalAssertions.kt
```diff
@@ -19,3 +19,29 @@ import java.util.*
...
+
+/**
+ * Expects that the subject of the assertion (an [Optional]) is present.
+ * and returns an [Expect] for the inner type [E].
+ *
+ * Shortcut for more or less something like `feature(Optional<T>::get)` but with error handling; yet it
+ * depends on the underlying implementation though.
+ *
+ * @return The newly created [Expect] if the given assertion is success
+ * @throws AssertionError Might throw an [AssertionError] if the given assertion is not a success.
```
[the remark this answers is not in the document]
robstoll: It's less precise as it depends not on the assertion function but on the context (group or not and what reporter was chosen) if an AssertionError is thrown. 
But maybe we could remove the `@throws AssertionError` all together. Yet in a different PR

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/optionalAssertions.kt
```diff
@@ -19,3 +19,29 @@ import java.util.*
...
 fun <T : Optional<*>> Expect<T>.isEmpty(): Expect<T> = addAssertion(ExpectImpl.optional.isEmpty(this))
+
+/**
+ * Expects that the subject of the assertion (an [Optional]) is present.
+ * and returns an [Expect] for the inner type [E].
+ *
+ * Shortcut for more or less something like `feature(Optional<T>::get)` but with error handling; yet it
+ * depends on the underlying implementation though.
+ *
+ * @return The newly created [Expect] if the given assertion is success
```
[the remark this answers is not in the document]
robstoll: `The newly created [Expect] if the given assertion is success` is correct, the subject was narrowed.

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/optionalAssertions.kt
```diff
@@ -19,3 +19,29 @@ import java.util.*
  * @since 0.9.0
  */
 fun <T : Optional<*>> Expect<T>.isEmpty(): Expect<T> = addAssertion(ExpectImpl.optional.isEmpty(this))
+
+/**
+ * Expects that the subject of the assertion (an [Optional]) is present.
+ * and returns an [Expect] for the inner type [E].
+ *
+ * Shortcut for more or less something like `feature(Optional<T>::get)` but with error handling; yet it
+ * depends on the underlying implementation though.
```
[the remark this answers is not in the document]
robstoll: I agree that we could remove `it depends on the underlying implementation though` Something we should do at other places as well. Something for a different PR, @jGleitz do you fancy to do the cleanup?

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/optionalAssertions.kt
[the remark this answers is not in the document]
jGleitz: How so? I would like to get a consistent description and would also do the PR. But I don’t understand under what circumstances atrium does _not_ throw an `AssertionError` if the function got called and the assertion does not hold.

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/optionalAssertions.kt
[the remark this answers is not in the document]
jGleitz: https://github.com/robstoll/atrium/issues/333

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/optionalAssertions.kt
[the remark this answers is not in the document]
robstoll: if you collect assertions for instance, then no error is thrown

slalu:
I think it's complete. I assumed no changes were required for Joshua's comment already marked as resolved.

robstoll:
@slalu I cannot merge your PR as it has conflicts. The reason is again that you merged instead of rebasing, please use the approach described in https://github.com/robstoll/atrium/pull/322#discussion_r369400521 again

slalu:
> @slalu I cannot merge your PR as it has conflicts. The reason is again that you merged instead of rebasing, please use the approach described in [#322 (comment)](https://github.com/robstoll/atrium/pull/322#discussion_r369400521) again

Sorry it should be ok now!

robstoll:
@slalu would you mind to change the commit message, `first draft...` does not sound so good. You can do this with `git commit --amend -m "add Optional.isPresent"

slalu:
@robstoll done!

robstoll:
Thanks and thanks for the contribution 🙂👍

PR #326, opened 2020-01-21
Change assertion verb in reporting to 'expected that subject'

Jak-Sidious:
This pull request caters to the change inside `ch.tutteli.atrium.api.verbs.AssertionVerb` where `asserted that subject` is used for `ASSERT` and `ASSERT_THAT` and `expected that subject` is used for `EXPECT`

----


robstoll:
LGTM thanks for the contribution :+1:

PR #352, opened 2020-02-05
rename `path toNot exist` to ... notTo ...

robstoll:
this way it resembles notToBe

----


jGleitz:
Oh. Sorry for my bad English! :speak_no_evil:

robstoll:
I think `expect path to not exist` would work as well, `expect path not to exist` is not really much nicer in terms of an English sentence i think but it is more consistent with the other methods we have. So no worries

Issue #319, opened 2020-01-14
zonedDateTimeAssertions for thew new infix API


robstoll:
*Platform* (jvm, js, android): jvm
*Extension* (none, kotlin 1.3, jdk8): jdk8

## Code related feature

We want to introduce ZonedDateTimeAssertions for the new infix API `api-infix-en_GB`.

Following the things you need to do:

- copy the file `zonedDateTimeAssertions.kt` from atrium-api-fluent-en_GB-jdk8-jvm to atrium-api-infix-en_GB-jdk8-jvm
  - adjust the signatures of the assertion functions (only `fun`) so that they are infix functions (see other files or files in atrium-api-infix-en_GB-common for a guideline)
  - open module-info.java in atrium-api-infix-en_GB-jdk8-jvm and make sure `exports ch.tutteli.atrium.api.infix.en_GB.jdk8;` is not commented out (might be someone else already did that by now)  
- copy the `ZonedDateTimeAssertionsSpec.kt` from atrium-api-fluent-en_GB-jdk8-jvm to atrium-api-infix-en_GB-jdk8-jvm
  - adjust the ambiguityTests so that we see the infix API in action (see for LocalDateTimeAssertionsSpec)

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime,
  we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

tfesenko:
I'll work on this

PR #368, opened 2020-02-10
#319: zonedDateTimeAssertions for thew new infix API

tfesenko:
Followed intructions from the issue:

- [X] copy and adjust the file zonedDateTimeAssertions.kt from atrium-api-fluent-en_GB-jdk8-jvm to atrium-api-infix-en_GB-jdk8-jvm
- [X] uncomment `exports ch.tutteli.atrium.api.infix.en_GB.jdk8` in module-info.java
- [X] copy and adjust `ZonedDateTimeAssertionsSpec.kt`
- [X] (not mentioned in the issue): changed the `@sinse` from 0.9.0 to 0.10.0 as this change will only be available in 0.10.0

----


robstoll:
@tfesenko thanks for your first contribution 🙂👍
How about a second one? The following should be an easy one for you:
https://github.com/robstoll/atrium/issues/315

PR #370, opened 2020-02-11
fix priority issues when it comes to `o` inside a block

robstoll:
----


apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/AnyAssertionsSpec.kt
```diff
@@ -76,12 +76,12 @@ class AnyAssertionsSpec : ch.tutteli.atrium.specs.integration.AnyAssertionsSpec(
         a1b notToBeNull {}
 
         a1 and o toBe 1
```
jGleitz: Let me guess, we cannot make this one also `it`?

robstoll: that's exactly the issue

robstoll:
I just realised that I am not yet happy with the workaround, going to close this one and think more about it

Issue #63, opened 2019-01-08
infix `to contain in any order` should be `contains inAnyOrder`


robstoll:
*Platform* (JVM and/or JS): all

## Code related feature
```
assert(listOf("hello", "world")) contains "hello" // is short for
assert(listOf("hello", "world")) to contain inAny order atLeast 1 value "hello"
```
I think I have chosen to go with `to contain` since otherwise `contains` is overloaded with `T` and with `inAnyOrder/inOrder` etc. However, thinking about it now I don't think that's too bad (maybe there were other reasons to). Anyway, it does not seem very intuitive to me and I think it would be better if we would start with `contains` as well as in the following example:
```
assert(listOf("hello", "world")) contains inAnyOrder atLeast 1 value "hello"
```
This way one starts to type `contains` and sees all options.

robstoll:
I see the other reason and that was probably the main point why I used `to contain`; it returns `IterableContains.Builder<E, T, NoOpSearchBehaviour>` as do the other APIs (e.g. cc-en_GB when using the `assert(x).contains.`). Nevertheless, from an extension point of view we can skip this step and add the extensions to `Assert<Iterable<T>>` directly. We don't lose anything here but we need to create the builder in each function (a bit of boilerplate but the trade off is worth it IMO).

robstoll:
I guess it would make sense if we provide only this signature for Expect

robstoll:
having `contains inAnyOrder` is nice but what about `Expect<Charsequence>.to contain`. We would still need a filler there. 
I was thinking about using `o` as filler as we are using it inside lambdas already. But it is not that nice even though one quickly skips it in the IDE while reading if appropriate syntax highlighting is active (as `contains` is e.g. in yellow and `o` in white) but I am not so sure
```
expect(text) to contain exactly 2 value "world"
// vs
expect(text) contains o exactly 2 value "world"
```
The benefit of the filler is of course that one can write `contains` and then use auto-completion where `to contain` is kind of another entry point. 

Thoughts?

robstoll:
I'll go with `contains o`

robstoll:
ffa95c1

PR #371, opened 2020-02-11
WIP feature assertion for the new infix API

robstoll:
----


apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/creating/feature/FeatureStepImpl.kt
```diff
@@ -0,0 +1,69 @@
...
+    @ExperimentalWithOptions
+    infix fun withOptions(
+        configuration: FeatureExtractorBuilder.OptionsChooser<R>.() -> Unit
+    ): FeatureStep<T, R> = withOptions(FeatureExtractorBuilder.OptionsChooser.createAndBuild(configuration))
+
+    @ExperimentalWithOptions
+    infix fun withOptions(options: FeatureOptions<R>): FeatureStep<T, R> = copy(options = options)
+
+    @Suppress("NOTHING_TO_INLINE")
+    inline infix fun it(@Suppress("UNUSED_PARAMETER") o: Expect<*>): Expect<R> = it(o)
```
robstoll: another workaround for the priority issue

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/featureAssertions.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    val l: Expect<List<Int>> = notImplemented()
+
+    l get 1 it o toBe 1
+
+    val p: Expect<Pair<String, Int>> = notImplemented()
+
+
+    p and {
+        first { o feature String::length it o toBe 1 }
```
robstoll: second `o` now actually refers to `this` and not to the filler object `o`. This is possible due to hack shown in the previous comment

robstoll:
@jGleitz the priority issues mentioned in #370 was detected in context of https://github.com/robstoll/atrium/pull/371/files#diff-b0e95edad2a7b545135758e6cddf328fR123
Please don't review the PR, there is a lot of garbage in it. Only have a look at the corresponding two comments I made and state your opinion about the workaround. This way one can always use `o` but auto-completion might be a bit confusing.

jGleitz:
Aaaargh, stuff like this is why I don’t like infix APIs. It always gets to the point where you have to use workarounds because you don’t have the full power of a normal programming language.

For my understanding, can you explain for `o feature String::length it o toBe 1` why we need the `it o`? I get the reason for the first `o`, but since `o feature String::length` is an `Expect<Int>`, I assumed we could just write `o feature String::length toBe 1`

robstoll:
I tried to make a consistent API for `feature`:
https://github.com/robstoll/atrium/pull/371/files#diff-b0e95edad2a7b545135758e6cddf328fR131

The problem is that you need a way to distinguish the two overloads one which does not expect an assertion creator and the other which expects one. Having the `it` we can do it. Without it we would need something like:
```
e feature Person::name toBe "a"
e feature of(Person::name) it { o toBe "a" }
```
But here we run into Kotlin type inference bugs quite quickly :(

robstoll:
Maybe looking at it form another way, why do we need the assertionCreator version?
This version does not narrow the subject. Following an example on how I would use the infix API
```
expect(person) {
 o feature { f(it::name) } toBe "robert" // subject now of type String in case a fluent happens
 // subject here person again as it is not a fluent call but within a group
 o feature { f(it::age) } /* narrowed to Int */ it { // it as an alias for and/addAssertionsCreatedBy
   o isLessThan 20
   o isGreaterThan 10
 }
}
```
Here it does not even matter that the subject was narrowed (there are cases where the assertionCreator-version would yield a better error report in addition but I think we can neglect that)_
However, if one writes in a fluent style, then it looks as follows
```
expect(person) it  // it as an alias for and/addAssertionsCreatedBy
  o feature { f(it::name) } toBe "robert" and   // subject narrowed to String, the next line would not compile  
  o feature { f(it::age) } it ...
```
And here it makes a difference as one wants to use the non-narrowing version.

One way we could proceed with this is to provide currently only the version without assertionCreator. the type inference bugs involving `of` should be fixed with Kotlin 1.4 i.e we can read it then. On the other hand, I like that the shown API here is almost identical for assertionCreator lambda or not, just `o` or `{...}`. So maybe we wait with the infix API as such until https://youtrack.jetbrains.com/issue/KT-36624 is fixed :laughing:

robstoll:
No, this is of course not an option. I think we should work around it, one is by using the hack shown in this PR, another would be to use a different identifier for the extension val; however, not `it` as proposed in the other PR, I would go with something nonsensical like `o` maybe `v` or `u`.
What is not so nice with `v` or `u` is that one cannot simply wrap the it into `{ }` as in the current version. i.e. going from 
```
e feature Person::name it o toBe "a"
```
to 
```
e feature Person::bar it { o toBe "a" }
```
Thus the hack might actually be better and we remove it once https://youtrack.jetbrains.com/issue/KT-36624 is fixed

jGleitz:
Okay. Am I getting this right that the whole problem stems from the fact that you try to allow users to specify multiple assertions without using a lambda?

In that case, I would like to challenge the whole use case. A syntax like

```
expect(person) it  
  o feature { f(it::name) } toBe "robert" and   
  o feature { f(it::age) } toBe 6
```

is _very_ fragile and confusing, from my point of view. On the other hand,

```
expect(person) {
  o feature { f(it::name) } toBe "robert" 
  o feature { f(it::age) } toBe 6
}
```

is not worse in any way, from my point of view. I would argue that it is even easier to read and I am not sure that even the strongest infix API supported would find the second worse than the first.

jGleitz:
Also, can we rename `o` to `the`? I think 

```
expect(person) {
  the feature { f(it::name) } toBe "robert" 
  the feature { f(it::age) } toBe 6
}
```

is easier to read.

Sure, once you understand that `o` is meant to be a bullet point, you get the hang of it. But I needed some readings until I understood it and found it quite confusing at first.

robstoll:
I intend to use `o` as filler object next to a bullet point. See https://github.com/robstoll/atrium/issues/63#issuecomment-576452230

`the` might make more sense for `feature` as shown in your example but only at the start. `it` makes more sense for things like
```
expect(int) {
  it isLessThan 20
  it isGreaterThan 4
}
```

IMO we can add `the` and `it` as alternatives in addition but I would keep `o` as well because
```
the feature { f(it::age) } it { 
  it isLessThan 20
  it isGreaterThan 4
}
```
Looks ugly as well no? Or do you have a different idea for `it` in ` feature { f(it::age) } it { ` ?

robstoll:
> In that case, I would like to challenge the whole use case.

I challenged myself as well. I first figured we could start with a reduced feature set for the new infix API. I.e. not support an overload which expects an assertionCreator-lambda, that's basically what I meant with:
> One way we could proceed with this is to provide currently only the version without assertionCreator.

I thought that we should think about how we would support it in the future without having a strange API but maybe we should just say YAGNI and wait until someone ask for it. What do you think?

robstoll:
> the whole problem stems from the fact that you try to allow users to specify multiple assertions without using a lambda?

That's one part, the other is:
> The problem is that you need a way to distinguish the two overloads one which does not expect an assertion creator

jGleitz:
> we can add `the` and `it` as alternatives

I like your principle of not having more than one way to specify a thing. So I am against alternatives here.

> What do you think?

I think when we publish the infix API, there should be a clear way how to do feature assertions. I think it is an important part and if it is missing, people may likely silently just not use the infix API. 

However, I also think that we should not create an API that is awkward to use. For example, I don’t understand why we need the second `it` in  `the feature { f(it::age) } it {`. And I can guarantee you that it would continue to irritate me whenever I needed to use the API. So let’s try and find solutions for all of them.

> The problem is that you need a way to distinguish the two overloads one which does not expect an assertion creator

If you want my opinion on this, you really need to walk me through this. I still don’t get most parts.

What I understand until now, is that we want these two syntaxes:

 1. `expect(subject) <one feature assertion without braces>`, e.g. `expect(person) feature Person::name toBe "peter"
 2. `expect(subject) { <multiple feature assertion statements>}`, e.g.

```
   expect(person) {
      the feature Person::name toBe "peter"
      the feature Person::age toBe 30
   }
```

As I said, I don’t think having multiple feature assertions without braces is a good use case. If you disagree, let us discuss!

What is keeping us from building those two?

robstoll:
> As I said, I don’t think having multiple feature assertions without braces is a good use case. If you disagree, let us discuss!

I agree, let us only cover the case where one uses braces for multiple assertions. However, something like: `expect(pair) feature { f(it::age) } toBe 1` should be possible without the need to wrap it into `{ .. }` after `expect(pair)` IMO => this is already covered, so nothing to do in this regard

> If you want my opinion on this, you really need to walk me through this. I still don’t get most parts.

`the feature { f(it::age) } { ... }` => feature would be a function with two arguments. That's not allowed for an infix call, that's why we need an extra step.
That's basically one of the main restrictions which I try to workaround smoothly (by additional pseudo keywords etc.). In the old infix API I did it with helper objects, for instance `Index` for List.get: https://github.com/robstoll/atrium/blob/master/apis/differences.md#list-get 
But they only work out well if they don't involve type parameters because otherwise Kotlin's type inference system is not good enough.

IMO we could start with the principle that we support assertionCreator lambdas only for 0-arity features such as `LocalDateTime.year` but do not support it for all other (e.g. List.get) and one can use `it` instead.

jGleitz:
> `the feature { f(it::age) } { ... }` => feature would be a function with two arguments. 

oookay, I see the problem there. Can we do that with extension functions? As a _very_ rough prototype, the following compiles:

```kotlin
class Expect

operator fun (() -> Expect).invoke(block: Expect.() -> Unit): () -> Expect = TODO()

infix fun String.feature(other: () -> Expect): Expect = TODO()

fun test() {
    "a" feature { Expect() } {

    }
}
```

Of course, I don’t know into which kind of type inference problems we will run wtih this.

robstoll:
Unfortunately your solution does not address
> The problem is that you need a way to distinguish the two overloads one which does not expect an assertion creator

You cannot call `"a" feature { Expect() } toBe 1`

jGleitz:
> You cannot call `"a" feature { Expect() } toBe 1` 
 

Why not? 

`"a" feature { Expect() }` returns an `Expect` on which we can call `toBe`. What am I missing here?

robstoll:
ah sorry, I misread the syntax, thought that `feature` returns `() -> Expect` respectively
I'll take a look at the idea, it could actually work out nicely :+1:

robstoll:
I think it could work out nicely for `feature` but I was not so sure if it is a good idea in other cases as we need to extend all kind of types with an invoke operator. 

However, as this operator does not show up in the code completion, I think it is good.
The function signature might be a bit harder to comprehend for the user and the operator function clutters the overall API but I would say that's fine for the infix API as the code completion support by Intellij isn't so good anyway.

following a prototype how it would look like for List.get

```
infix fun <E, T: List<E>> Expect<T>.get(index: Int): Expect<E> = TODO()
infix fun <E, T: List<E>> Expect<T>.get(indexWithAssertionCreator: Pair<Int, Expect<E>.() -> Unit>): Expect<T> = TODO()
operator fun <T, R> T.invoke(assertionCreator: Expect<R>.() -> Unit): Pair<T, Expect<R>.() -> Unit> = TODO()
fun foo() {
    expect(listOf(1,2)) get 1 toBe 1
    expect(listOf(1,2)) get 1 {
        o isLessThan 2
        o isGreaterThan 39
    }
}
```
To my surprise, the inference system did not fail :)

jGleitz:
Hey, now I feel we are getting somewhere! :+1:

robstoll:
Me too, glad we talked about it

robstoll:
Unfortunately it does not work so well as soon as type parameters are involved
```
e feature Person::name toBe "a"
e feature Person::bar { o toBe "a" }

e feature withArgs(Person::foo, 1) toBe "a"
e feature withArgs(Person::foo, 1) { o toBe "a" }           // fails
e feature withArgs(Person::foo, 1) invoke { o toBe "a" }    // fails
e feature (withArgs(Person::foo, 1) invoke { o toBe "a" })  

e feature { f(it::age) } toBe 1
e feature { f(it::age) } { o toBe 1 }                       // fails
e feature ({ f(it::age) } { o toBe 1 })                     // fails
e feature ({ f(it::age) } invoke { o toBe 1 })              // fails as well, there is kind of no immediate workaround

e feature { manual("age") { it.age } } isGreaterThan 2
```

Due to this I think we should use a different solution. I am going to use it for List.get as well as the class-reference version but we need something else for `withArgs` and `{ f(...) }`

jGleitz:
Do you have a branch where I can see it not working? I’d like to understand what the problem is. This branch currently still has `e feature Person::name it o toBe "a"` in `featureAssertions.kt`

robstoll:
I'll push it later on

robstoll:
I have pushed it to this branch, commit 442255a9
and I have created a bug: https://github.com/robstoll/atrium/pull/371#issuecomment-585747890

robstoll:
And another bug which comes into our way: https://youtrack.jetbrains.com/issue/KT-37135
I'll have to create more than one workaround I guess

robstoll:
@jGleitz I had a further look at your idea. Unfortunately it is very fragile. As soon as one defines an `operator invoke` for the type we would like to wrap, it fails. (see https://github.com/robstoll/atrium/commit/e967714425b907085db91ba2a24a53951891c842#diff-b0e95edad2a7b545135758e6cddf328fR238) the most stable way is still the one which uses an explicit wrapper class/function. I think using a function could help here to simplify things a bit, I'll do further work in this direction

robstoll:
superseded by #381

PR #372, opened 2020-02-12
Add path resolve shortcut

slalu:
I'm kind of blocked and I want to make sure I'm going somewhere with this. So, before I go further, I would like to:
- Confirm that the usage of ExpectImpl.feature.f1 is correct in pathAssertions.kt 
- Maybe have an hint on how to make PathAssertionsImpl compile. I suspect something wrong with the generics...

Thanks!

----

### DIFF
--- a/apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/pathAssertions.kt
+++ b/apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/pathAssertions.kt
@@ -156,6 +156,29 @@ val <T : Path> Expect<T>.parent: Expect<Path>
 fun <T : Path> Expect<T>.parent(assertionCreator: Expect<Path>.() -> Unit): Expect<T> =
     ExpectImpl.path.parent(this).addToInitial(assertionCreator)
 
+/**
+ * Expects that [other] resolve against this [Path] and creates an [Expect] for the resolved [Path]
+ * so that further fluent calls are assertions about it.
+ *
+ * @return The newly created [Expect].
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ *
+ * @since 0.10.0
+ */
+fun <T : Path> Expect<T>.resolve(other: String): Expect<Path> =
+    ExpectImpl.path.resolve(this, other).getExpectOfFeature()
+
+/**
+ * Expect that [other] resolve against this [Path] and that the resolved [Path] holds all assertions the
+ * given [assertionCreator] creates for it and returns this assertion container.
+ *
+ * @return This assertion container to support a fluent API.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ *
+ * @since 0.10.0
+ */
+fun <T : Path> Expect<T>.resolve(other: String, assertionCreator: Expect<Path>.() -> Unit): Expect<T> =
+    ExpectImpl.path.resolve(this, other).addToInitial(assertionCreator)
 
 /**
  * Expects that the subject of the assertion (a [Path]) is readable;
--- a/apis/infix-en_GB/extensions/jdk8/atrium-api-infix-en_GB-jdk8-jvm/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/jdk8/PathFeatureAssertionsSpec.kt
+++ b/apis/infix-en_GB/extensions/jdk8/atrium-api-infix-en_GB-jdk8-jvm/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/jdk8/PathFeatureAssertionsSpec.kt
@@ -1,15 +1,15 @@
 package ch.tutteli.atrium.api.infix.en_GB.jdk8
 
 import ch.tutteli.atrium.creating.Expect
-import ch.tutteli.atrium.specs.fun1
-import ch.tutteli.atrium.specs.notImplemented
-import ch.tutteli.atrium.specs.property
+import ch.tutteli.atrium.specs.*
 import ch.tutteli.atrium.specs.testutils.WithAsciiReporter
 import java.nio.file.Path
 
 class PathFeatureAssertionsSpec : ch.tutteli.atrium.specs.integration.PathFeatureAssertionsSpec(
     property<Path, Path>(Expect<Path>::parent),
     fun1<Path, Expect<Path>.() -> Unit>(Expect<Path>::parent),
+    feature1<Path, String, Path>(Expect<Path>::resolve),
+    fun2<Path, String,  Expect<Path>.() -> Unit>(Expect<Path>::resolve), //resolve with assertionCreator not implemented in this API
     property<Path, String>(Expect<Path>::fileName),
     fun1<Path, Expect<String>.() -> Unit>(Expect<Path>::fileName),
     property<Path, String>(Expect<Path>::fileNameWithoutExtension),
@@ -34,5 +34,10 @@ class PathFeatureAssertionsSpec : ch.tutteli.atrium.specs.integration.PathFeatur
 
         a1.parent
         a1 parent {}
+
+        a1 resolve "test"
     }
 }
+
+private fun <T: Path> Expect<T>.resolve(other: String, assertionCreator: Expect<Path>.() -> Unit): Expect<Path> =
+    (this resolve other).addAssertionsCreatedBy(assertionCreator)
--- a/domain/builders/atrium-domain-builders-jvm/src/main/kotlin/ch/tutteli/atrium/domain/builders/creating/PathAssertionsBuilder.kt
+++ b/domain/builders/atrium-domain-builders-jvm/src/main/kotlin/ch/tutteli/atrium/domain/builders/creating/PathAssertionsBuilder.kt
@@ -48,6 +48,9 @@ object PathAssertionsBuilder : PathAssertions {
     override inline fun <T : Path> parent(expect: Expect<T>) =
         pathAssertions.parent(expect)
 
+    override inline fun <T : Path> resolve(expect: Expect<T>, other: String) =
+        pathAssertions.resolve<T>(expect, other)
+
     override inline fun <T : Path> isReadable(expect: Expect<T>) =
         pathAssertions.isReadable(expect)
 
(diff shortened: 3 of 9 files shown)


apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/pathAssertions.kt
```diff
@@ -156,6 +156,9 @@ val <T : Path> Expect<T>.parent: Expect<Path>
 fun <T : Path> Expect<T>.parent(assertionCreator: Expect<Path>.() -> Unit): Expect<T> =
     ExpectImpl.path.parent(this).addToInitial(assertionCreator)
 
+//TODO to be completed.
+fun <T : Path> Expect<T>.resolve(assertionCreator: Expect<Path>.() -> Unit, other: String): Expect<T> =
```
robstoll: switch arguments, assertionCreator-lambda should always be the last argument
```suggestion
fun <T : Path> Expect<T>.resolve(other: String, assertionCreator: Expect<Path>.() -> Unit): Expect<T> =
```

domain/builders/atrium-domain-builders-jvm/src/main/kotlin/ch/tutteli/atrium/domain/builders/creating/PathAssertionsBuilder.kt
```diff
@@ -48,6 +48,9 @@ object PathAssertionsBuilder : PathAssertions {
     override inline fun <T : Path> parent(expect: Expect<T>) =
         pathAssertions.parent(expect)
 
+    override fun <T : Path> resolve(expect: Expect<T>, other: String) =
```
robstoll: please inline
```suggestion
    override inline fun <T : Path> resolve(expect: Expect<T>, other: String) =
```

domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/pathAssertions.kt
```diff
@@ -395,5 +396,8 @@ fun <T : Path> _parent(expect: Expect<T>): ExtractedFeaturePostStep<T, Path> =
         .withoutOptions()
         .build()
 
+fun <T : Path, String, R: Path> _resolve(expect: Expect<T>, f: KFunction2<T, String, R>, other: String): ExtractedFeaturePostStep<T, Path> =
```
robstoll: don't expect the KFunction here, using Path.resolve should be internal to this function.
```suggestion
fun <T : Path> _resolve(expect: Expect<T>, other: String): ExtractedFeaturePostStep<T, Path> =
```
Also your definition `fun <T: Path, String, R: Path>` actually defines a type parameter named `String`. That's certainly not what you want :slightly_smiling_face:  Most likely the reason why you run into type inference problems. Note also, that you don't require `R` either as you don't use it afterwards.

apis/infix-en_GB/extensions/jdk8/atrium-api-infix-en_GB-jdk8-jvm/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/jdk8/PathFeatureAssertionsSpec.kt
```diff
@@ -1,15 +1,15 @@
...
-import ch.tutteli.atrium.specs.property
+import ch.tutteli.atrium.specs.*
 import ch.tutteli.atrium.specs.testutils.WithAsciiReporter
 import java.nio.file.Path
 
 class PathFeatureAssertionsSpec : ch.tutteli.atrium.specs.integration.PathFeatureAssertionsSpec(
     property<Path, Path>(Expect<Path>::parent),
     fun1<Path, Expect<Path>.() -> Unit>(Expect<Path>::parent),
+    feature1<Path, String, Path>(Expect<Path>::resolve),
+    notImplemented(),
```
slalu: I cannot implement the infix function with both parameters (String and lambda), so I am using notImplemented here. Not sure if it's the right thing to do.

slalu:
At least it compiles now. I'm still working on the tests. You can let me know if you have comments on what's completed in the meantime

apis/infix-en_GB/extensions/jdk8/atrium-api-infix-en_GB-jdk8-jvm/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/jdk8/PathFeatureAssertionsSpec.kt
```diff
@@ -34,5 +34,7 @@ class PathFeatureAssertionsSpec : ch.tutteli.atrium.specs.integration.PathFeatur
 
         a1.parent
         a1 parent {}
+
+        a1.resolve("test")
```
robstoll: don't use parentheses here. That's the main difference between the fluent and the infix api
```suggestion
        a1 resolve "test"
```

apis/infix-en_GB/extensions/jdk8/atrium-api-infix-en_GB-jdk8-jvm/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/jdk8/PathFeatureAssertionsSpec.kt
robstoll: It's ok that you cannot implement the infix for two parameters (which is not possible). I am not yet sure how I am going to support this in the new infix API. However, you cannot use notImplemented() as this will fail the test.
Instead, create the following function in the same file
```
private fun <T: Path> resolve(expect: Expect<T>, other: String): Expect<T> = and {
  expect resolve other
}
```  
and use it
```suggestion
    "resolve with assertionCreator not implemented in this API" to ::resolve,
```

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/pathAssertions.kt
```diff
@@ -156,6 +156,29 @@ val <T : Path> Expect<T>.parent: Expect<Path>
 fun <T : Path> Expect<T>.parent(assertionCreator: Expect<Path>.() -> Unit): Expect<T> =
     ExpectImpl.path.parent(this).addToInitial(assertionCreator)
 
+/**
+ * Converts the given [other] to a path and expects it to resolve against this [Path]. Creates an [Expect] for it,
```
robstoll: This is slightly wrong as you don't convert `other` to a path first. Also I would stick to other docs and not split it up in two sentences. I suggest:
```suggestion
 * Expects that [other] resolve against this [Path] and creates an [Expect] for the resolved [Path]
```

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/pathAssertions.kt
```diff
@@ -156,6 +156,29 @@ val <T : Path> Expect<T>.parent: Expect<Path>
...
+ * @return The newly created [Expect].
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ *
+ * @since 0.10.0
+ */
+fun <T : Path> Expect<T>.resolve(other: String): Expect<Path> =
+    ExpectImpl.path.resolve(this, other).getExpectOfFeature()
+
+/**
+ * Converts the given [other] to a path and expects it to resolve against this [Path]. Also expects that the parent holds
```
robstoll: adapt this as well as KDoc in infix API accordingly to the above input

apis/infix-en_GB/extensions/jdk8/atrium-api-infix-en_GB-jdk8-jvm/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/jdk8/PathFeatureAssertionsSpec.kt
```diff
@@ -34,5 +34,10 @@ class PathFeatureAssertionsSpec : ch.tutteli.atrium.specs.integration.PathFeatur
 
         a1.parent
         a1 parent {}
+
+        a1 resolve "test"
     }
 }
+
+private fun <T: Path> Expect<T>.resolve(other: String, assertionCreator: Expect<Path>.() -> Unit): Expect<Path> =
```
slalu: Wasn't able to make the suggestion compile. Hope it's ok this way.

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathFeatureAssertionsSpec.kt
```diff
@@ -105,6 +109,49 @@ abstract class PathFeatureAssertionsSpec(
         }
     }
 
+    describeFun("val ${resolveFeature.name}") {
```
slalu: This class seemed to use multiples describeFun as opposed at what I had to do in OptionalAssertionsSpec. I followed the pattern already used.

robstoll: that's fair enough

apis/infix-en_GB/extensions/jdk8/atrium-api-infix-en_GB-jdk8-jvm/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/jdk8/PathFeatureAssertionsSpec.kt
robstoll: I wrote it off the top of my head without checking if it actually compiles. It's good enough for this PR but I am going to rewrite it after the merge. For my taste it resembles too much like a function which is implemented in this API. I'll send you a link to the comment later on

robstoll: https://github.com/robstoll/atrium/commit/8ff3a7495ed5839587da5f35d976184a4a7a4d5e

PR #376, opened 2020-02-18
print error message only once for IntelliJ runners

robstoll:
JUnit
```
assert: 1        (kotlin.Int <462882660>)
◆ equals: 2        (kotlin.Int <143144795>)

ch.tutteli.atrium.reporting.AtriumError
	at ch.tutteli.atrium.reporting.AtriumErrorKt.createAtriumError(AtriumError.kt:26)
        ....
```

Spek
```
assert: ch.tutteli.atrium.reporting.AtriumError
◆ ▶ cause: null
    ◾ is type or sub-type of: Throwable (kotlin.Throwable) -- Class: java.lang.Throwable

ch.tutteli.atrium.reporting.AtriumError
	at ch.tutteli.atrium.reporting.AtriumErrorKt.createAtriumError(AtriumError.kt:26)
        ....
```

gradle
```
ch.tutteli.atrium.reporting.AtriumErrorSpec > has `null` as cause FAILED
    assert: ch.tutteli.atrium.reporting.AtriumError
    ◆ ▶ cause: null
        ◾ is type or sub-type of: Throwable (kotlin.Throwable) -- Class: java.lang.Throwable

    ch.tutteli.atrium.reporting.AtriumError
        at ch.tutteli.atrium.reporting.AtriumErrorKt.createAtriumError(AtriumError.kt:26)
        ....
```

----

### DIFF
--- a/core/api/atrium-core-api-jvm/src/main/kotlin/ch/tutteli/atrium/reporting/AtriumError.kt
+++ b/core/api/atrium-core-api-jvm/src/main/kotlin/ch/tutteli/atrium/reporting/AtriumError.kt
@@ -1,5 +1,6 @@
 package ch.tutteli.atrium.reporting
 
+import ch.tutteli.atrium.core.polyfills.fullName
 import ch.tutteli.atrium.reporting.AtriumError.Companion
 
 /**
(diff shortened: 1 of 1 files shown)


robstoll:
@jGleitz the solution is inspired by your idea. it's still a hack and as you can see from the failing tests, it has some unwanted side effects.

jGleitz:
I am generally in favour of merging this.

But I wonder: The runners’ behaviour regarding this is surprisingly consistent, given that the behaviour seems like a bug to me. Is there any convention regarding this? Is there any place we could file a bug?

robstoll:
I had a closer look and I could see that the behaviour of showing the message + printStracktrace is specific to spek (https://github.com/raniejade/spek-idea-plugin/blob/8853f044fcf451ac799c375af15e203cbd9296f1/tooling/src/main/kotlin/org/jetbrains/spek/tooling/adapter/sm/ServiceMessageAdapter.kt#L51)

When running a JUnit test, then ExceptionPlaceholder from gradle is used:
https://github.com/gradle/gradle/blob/master/subprojects/messaging/src/main/java/org/gradle/internal/serialize/ExceptionPlaceholder.java#L64
Depending on the configured reporting it will show only that a test failed, it shows the message or it shows message + printStacktrace. Or in other words, the above hack probably won't work if one only shows the message. I have to check though.

> The runners’ behaviour regarding this is surprisingly consistent, 

Not sure what you are referring to; that they use message and not localizedMessage? I guess most people are not aware of it and in most cases localizedMessage is the same as message.

robstoll:
gradle with showExceptions = true but showStacktrace= false:
```
Test > atrium() FAILED
    expect: 1        (kotlin.Int <956500077>)
    ◆ equals: 2        (kotlin.Int <1637453566>)

    ch.tutteli.atrium.reporting.AtriumError
```
gradle with exceptionFormat  = TestExceptionFormat.SHORT 
```
Test > atrium() FAILED
    ch.tutteli.atrium.reporting.AtriumError at SmokeSpec.kt:23
```
:fearful: 

However, without the hack:
```
Test > atrium() FAILED
    ch.tutteli.atrium.reporting.AtriumError at SmokeSpec.kt:23
```
:relieved: 

I think we can merge this. I am going to rebase the PR.

robstoll:
Oh... right, the tests fail now because they use `messageContains` a lot. Hm... it stays a hack, leaves me with a bad after taste. I'll think about this again... not sure if the hack is worth the outcome.

jGleitz:
I agree that the hack is an improvement or at least not worse in all scenarios you have shown.

robstoll:
I restricted the hack to spek and junit (or rather `org.gradle.internal.serialize.ExceptionPlaceholder`, might affect other test runners as well). This way `message` behaves normal for others (at the price of checking the caller stack) which is better IMO.

PR #396, opened 2020-03-16
#116 migrate spek

assaflei:
Started working on DetailedObjectFormatterSpec class
Unfortunately I'm getting compile errors but cant find any references online or examples to proceed with resolution

First error related to usage of include keyword: "Expecting a class body".

I tried this command: include(object: AtriumsObjectFormatterSpec() {}) but got a different error: "Type mismatch: inferred type is <no name provided> but Spek was expected"

@robstoll would appreciate your feedback and advice on how to proceed

----

### DIFF
--- a/core/robstoll-lib/atrium-core-robstoll-lib-jvm/src/test/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/DetailedObjectFormatterSpec.kt
+++ b/core/robstoll-lib/atrium-core-robstoll-lib-jvm/src/test/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/DetailedObjectFormatterSpec.kt
@@ -11,18 +11,16 @@ import ch.tutteli.atrium.reporting.translating.UsingDefaultTranslator
 import ch.tutteli.atrium.specs.reporting.ObjectFormatterSpec
 import ch.tutteli.atrium.api.verbs.internal.AssertionVerb
 import com.nhaarman.mockitokotlin2.doReturn
-import com.nhaarman.mockitokotlin2.mock
-import org.jetbrains.spek.api.Spek
-import org.jetbrains.spek.api.dsl.context
-import org.jetbrains.spek.api.dsl.describe
-import org.jetbrains.spek.api.dsl.it
-import org.jetbrains.spek.api.dsl.on
-import org.jetbrains.spek.api.include
+import io.mockk.*
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+//import org.jetbrains.spek.api.dsl.on
+//import ch.tutteli.atrium.specs.include
 import kotlin.reflect.KClass
 
 //TODO #116 migrate spek1 to spek2 - move to common module
 object DetailedObjectFormatterSpec : Spek({
-    include(AtriumsObjectFormatterSpec)
+    include(object: AtriumsObjectFormatterSpec)
 
     val testee = DetailedObjectFormatter(UsingDefaultTranslator())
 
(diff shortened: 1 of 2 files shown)


core/robstoll-lib/atrium-core-robstoll-lib-jvm/src/test/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/DetailedObjectFormatterSpec.kt
```diff
@@ -11,18 +11,16 @@ import ch.tutteli.atrium.reporting.translating.UsingDefaultTranslator
...
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+//import org.jetbrains.spek.api.dsl.on
+//import ch.tutteli.atrium.specs.include
 import kotlin.reflect.KClass
 
 //TODO #116 migrate spek1 to spek2 - move to common module
 object DetailedObjectFormatterSpec : Spek({
-    include(AtriumsObjectFormatterSpec)
+    include(object: AtriumsObjectFormatterSpec)
```
robstoll: Migrate ch.tutteli.atrium.specs.reporting.ObjectFormatterSpec first (misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/ObjectFormatterSpec.kt). Afterwards you can leave this line as it was

assaflei:
Work on ObjectFormatterSpec done, should it be copied to a new folder?
Regarding DetailedObjectFormatterSpec I started working on it and have a progress - switched from describe, on, it to Feature, Scenario, When (Given), Then due to instructions from spek migration guide
@robstoll can you review the updated code in DetailedObjectFormatterSpec and approve the concept before I proceed?

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/ObjectFormatterSpec.kt
```diff
@@ -9,32 +9,27 @@ import ch.tutteli.atrium.reporting.StringBasedRawString
...
+import org.spekframework.spek2.style.specification.Suite
+
 abstract class ObjectFormatterSpec(
     testeeFactory: (Translator) -> ObjectFormatter,
     describePrefix: String = "[Atrium] "
 ) : Spek({
 
-    fun describeFun(vararg funName: String, body: SpecBody.() -> Unit) =
-        describeFun(describePrefix, funName, body = body)
+    fun describeFunTemplate(vararg funName: String, body: Suite.() -> Unit) =
```
robstoll:
```suggestion
    fun describeFun(vararg funName: String, body: Suite.() -> Unit) =
```

assaflei: Hi, thanks for your feedback
Originally there were 'on' blocks which need to be changed to 'context' - we might want to log this as a rule of thumb for later changes.

Another issue I face is nesting 'group' block under 'describe'
compilation fails, which can be resolved in one of two ways:
- change group to context
- move group block outside of containing describe

what do you think?

Compilation error is: 'fun group(description: String, skip: Skip = ..., defaultCachingMode: CachingMode = ..., preserveExecutionOrder: Boolean = ..., failFast: Boolean = ..., body: GroupBody.() -> Unit): Unit' can't 
be called in this context by implicit receiver. Use the explicit one if necessary

robstoll: > Originally there were 'on' blocks which need to be changed to 'context' - we might want to log this as a rule of thumb for later changes.

Good idea, I have updated the description of the issue (also the fact that `context` should be used instead of `group`)

assaflei:
Latest changed uploaded
Just to verify, successful gr build command means that task is valid and ready for merge, right?

robstoll:
> Just to verify, successful gr build command means that task is valid and ready for merge, right?

Usually yes but CI actually does a bit more than just `gr bulid` it also executes backward compatibility tests. In the end you need green checks here and an approval of a reviewer.

robstoll:
@assaflei thanks for the contribution :+1: 
I would migrate OnlyFailureReporterSpec as next step. But up to you :slightly_smiling_face:

PR #398, opened 2020-03-18
working on OnlyFailureReporterSpec

assaflei:
Used mockk just Runs command for verify action
Please approve if all is valid and I will finish this task

----

### DIFF
--- a/misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/OnlyFailureReporterSpec.kt
+++ b/misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/OnlyFailureReporterSpec.kt
@@ -20,16 +20,7 @@ import ch.tutteli.atrium.translations.DescriptionBasic.TO_BE
 import io.mockk.*
 import org.spekframework.spek2.Spek
 import org.spekframework.spek2.style.specification.Suite
-// import com.nhaarman.mockitokotlin2.any
-// import com.nhaarman.mockitokotlin2.eq
-// import com.nhaarman.mockitokotlin2.mock
-// import com.nhaarman.mockitokotlin2.verify
-// import org.jetbrains.spek.api.Spek
-// import org.jetbrains.spek.api.dsl.SpecBody
-// import org.jetbrains.spek.api.dsl.context
-// import org.jetbrains.spek.api.dsl.it
 
-//TODO #116 migrate spek1 to spek2 - move to specs-common
 abstract class OnlyFailureReporterSpec(
     testeeFactory: (AssertionFormatterFacade, AtriumErrorAdjuster) -> Reporter,
     describePrefix: String = "[Atrium] "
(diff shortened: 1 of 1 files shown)


robstoll:
Looks good to me

assaflei:
Done
Whats next?

robstoll:
Thanks :slightly_smiling_face: 
I would continue with the `Text...Specs` in the package `ch.tutteli.atrium.specs.reporting`. Might well be that you have to migrate all at once. 
Note, in case I should not reply quickly you can basically migrate any spec which still uses Spek 1.x Search for `import org.jetbrains.spek.api.Spek` and you will find them

assaflei:
Ok I'll try
The class TextIndentBasedAssertionGroupFormatterSpec is marked as not needed for spek migration but it breaks build since other classes are dependent of it
What do you say?

robstoll:
Just delete it, it's not worth to migrate it.

assaflei:
delete the file? (misc\specs\atrium-specs-jvm\src\main\kotlin\ch\tutteli\atrium\specs\reporting\TextIndentBasedAssertionGroupFormatterSpec.kt) and remove the include blocks using it?

robstoll:
wait, that's not the one I was thinking of :laughing: give me a minute

robstoll:
Delete ch.tutteli.atrium.core.robstoll.lib.reporting.TextIndentAssertionGroupFormatterSpec
but keep and migrate ch.tutteli.atrium.specs.reporting.TextIndentAssertionGroupFormatterSpec -- the `//TODO remove with 1.0.0 - no need to migrate to spek2` was wrong, you can remove that TODO as well

assaflei:
see this class: misc\specs\atrium-specs-jvm\src\main\kotlin\ch\tutteli\atrium\specs\reporting\TextExplanatoryBasedAssertionGroupFormatterSpec.kt
when upgrading to new spek, build fails due to the other class

robstoll:
btw. I just realised that the specs you migrated are still under specs-jvm. Can you please move them to atrium-specs-common. Thanks

assaflei:
you mean reporting folder in specs-jvm to similar folder in specs-common?

robstoll:
for instance, move 
misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/ObjectFormatterSpec.kt
to
/misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/reporting/ObjectFormatterSpec.kt

robstoll:
> see this class: misc\specs\atrium-specs-jvm\src\main\kotlin\ch\tutteli\atrium\specs\reporting\TextExplanatoryBasedAssertionGroupFormatterSpec.kt
when upgrading to new spek, build fails due to the other class

with other class you mean the subclasses right? you need to migrate them as well. That's what I meant with you probably need to migrate all of them at once

assaflei:
I started moving classes from specs-jvm to specs-common
Than I get these errors: "Unresolved reference: System" and "Unresolved reference: java"
It looks like a gradle related issue, can you advice which configuration should be used?

robstoll:
As I don't know which class you moved I can only guess that you tried to move a spec which is actually jvm specific. I think you should be able to move ObjectFormatterSpec and OnlyFailureReporterSpec

assaflei:
So in case of jvm dependencies, leave the classes there?

robstoll:
normally yes, but sometimes it could also be that we used the wrong type. For instance java.util.Collection instead of kotlin.Collection. I need the class to be able to judge

assaflei:
In OnlyFailureReporterSpec I get Unresolved reference: java for this command assertionFormatterFacade::class.java.simpleName
So same jvm dependency here?

robstoll:
you can use `::class.simpleName` instead

PR #405, opened 2020-03-21
[WIP] #116 migrate spek

assaflei:
Managed to push all misc jvm reporting classes
Few points
AssertionFormatterSpecBase use of v1 spek wrap function which doesnt exist in v2, I opened an issue in spek framework to add similar method

TextIndentBasedAssertionGroupFormatterSpec marked to delete but updated as well

Update if all is fine and I will remove WIP status

----


misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/TextExplanatoryAssertionGroupFormatterSpec.kt
```diff
@@ -5,16 +5,18 @@ import ch.tutteli.atrium.assertions.ExplanatoryAssertionGroupType
 import ch.tutteli.atrium.domain.builders.ExpectImpl
 import ch.tutteli.atrium.reporting.AssertionFormatter
 import ch.tutteli.atrium.reporting.AssertionFormatterController
+import org.spekframework.spek2.Spek
 import kotlin.reflect.KClass
 
 //TODO #116 migrate spek1 to spek2 - move to specs-common
```
robstoll: don't forget to remove this TODO as well

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/TextExplanatoryAssertionGroupFormatterSpec.kt
```diff
@@ -5,16 +5,18 @@ import ch.tutteli.atrium.assertions.ExplanatoryAssertionGroupType
...
 abstract class TextExplanatoryAssertionGroupFormatterSpec(
     testeeFactory: (Map<KClass<out BulletPointIdentifier>, String>, AssertionFormatterController) -> AssertionFormatter,
     describePrefix: String = "[Atrium] "
-) : TextExplanatoryBasedAssertionGroupFormatterSpec<ExplanatoryAssertionGroupType>(
+) : Spek({
+    include(object : TextExplanatoryBasedAssertionGroupFormatterSpec<ExplanatoryAssertionGroupType>(
```
robstoll: what's the reason that you have changed it to an include?  there is no additional spec going on in this class so IMO it is enough if it just extends TextExplanatoryBasedAssertionGroupFormatterSpec

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/TextExplanatoryAssertionGroupFormatterSpec.kt
assaflei: done

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/TextExplanatoryAssertionGroupFormatterSpec.kt
assaflei: indeed it was not necessary (was required during development), I reverted back to original version

robstoll:
@assaflei looks good to me, do you want to do the move to the common module in this PR or in the next?

assaflei:
I placed ObjectFormatterSpec and OnlyFailureReporterSpec in common folder do we have anymore?

robstoll:
You should be able to move the Text...Spec you just migrated as well

robstoll:
@assaflei I wait for your feedback here. Remove the WIP prefix in the title in case I shall merge

assaflei:
I already started working on the rest of the classes to convert and I'm handling it
I will commit all the changes up to now later
Is this ok?

robstoll:
sure, just wanted to be sure that I don't misunderstand you and you wait until I merge :)

assaflei:
No its ok
You'll be surprised but there are still new issues I encounter :)

assaflei:
I finished working on all classes marked as #116

Lots of specs-jvm classes couldnt be moved due to dependency in java classes, I managed to move only one

Details on the commit:

TextMethodCallFormatterSpec moved to specs-common
misc\specs\atrium-specs-jvm\src\main\kotlin\ch\tutteli\atrium\specs\SpecBodyExtensions.kt - can be deleted according to comment in it so can I delete?
There are still files using old spek but not marked as #116, what do you want to do with them?
I noticed that there could be classes that should be moved to common in other packages than misc\specs - in robstoll-lib classes in atrium-core-robstoll-lib-jvm can be moved to atrium-core-robstoll-lib-common - is this included in the scope as well? (core\robstoll-lib\atrium-core-robstoll-lib-jvm\src\test\kotlin\ch\tutteli\atrium\core\robstoll\lib\checking\ThrowingAssertionCheckerSpec.kt)

robstoll:
In the end we want to get rid of spek 1 so please migrate everything which is located in one of the mentioned modules in the description of the assertion (please write here which one you spotted not that you do work for nothing).

Yes please move classes in jvm which are not jvm specific to common. but let's do this in another PR so that we can already merge this one and thus lower the chances for conflicts.

I'll review the changes later on

assaflei:
Regarding the move from jvm to common I'll put it aside for now

From the two modules in the scope of the issue there are:
domain\robstoll-lib\atrium-domain-robstoll-lib-jvm\src\test\kotlin\ch\tutteli\atrium\assertions\LazyThreadUnsafeAssertionGroupSpec.kt

domain\robstoll-lib\atrium-domain-robstoll-lib-jvm\src\test\kotlin\ch\tutteli\atrium\assertions\LazyThreadUnsafeBasicAssertionSpec.kt

domain\robstoll-lib\atrium-domain-robstoll-lib-jvm\src\test\kotlin\ch\tutteli\atrium\creating\charsequence\contains\searchers\CharSequenceContainsIndexSearcherSpec.kt

domain\robstoll-lib\atrium-domain-robstoll-lib-jvm\src\test\kotlin\ch\tutteli\atrium\creating\charsequence\contains\searchers\CharSequenceContainsRegexSearcherSpec.kt

Waiting for you approval to proceed

robstoll:
Yes, they can be moved :+1:

assaflei:
Migrated another class which contains issues with order of execution and the lazy loading class
hope changes I made are correct and fit your expectations
Will proceed during the week

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/SpecBodyExtensions.kt
```diff
@@ -1,26 +1,29 @@
-//TODO remove file as soon as all specs have been transformed to spek2
-package ch.tutteli.atrium.specs
+// //TODO remove file as soon as all specs have been transformed to spek2
+// package ch.tutteli.atrium.specs
```
robstoll: You commented the code out and the build is still green, ergo you can remove the file :slightly_smiling_face:

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/translating/TranslationSupplierBasedTranslatorSpec.kt
```diff
@@ -4,30 +4,33 @@ import ch.tutteli.atrium.api.fluent.en_GB.toBe
...
-    fun mockTranslationProvider(locale: Locale, translatable: Translatable, translation: String): TranslationSupplier {
-        return mock {
-            on { get(translatable, locale) }.doReturn(translation)
+    fun mockTranslationProvider(locale: Locale, translatable: Translatable, translation: String?): TranslationSupplier {
+        val localeSlot = slot<Locale>()
+        val translatableSlot = slot<Translatable>()
+        return mockk {
+            every { get(capture(translatableSlot), capture(localeSlot)) } answers {
+                // println("translatableSlot.captured.value:" + (translatableSlot.captured as StringBasedTranslatable).value);
+                // println("translatable.value:" + (translatable as StringBasedTranslatable).value);
```
robstoll: remove commented out code

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/translating/TranslationSupplierBasedTranslatorSpec.kt
```diff
@@ -46,27 +49,29 @@ abstract class TranslationSupplierBasedTranslatorSpec(
     }
 
     @Suppress("unused")
-    fun SpecBody.checkUsesDefaultOfTranslatable(testee: Translator) {
-        it("uses ${Translatable::class.simpleName}'s ${Translatable::getDefault.name}") {
-            val result = testee.translate(translatableHello)
-            expect(result).toBe(translatableHello.value)
+    fun ScopeBody.checkUsesDefaultOfTranslatable(testee: Translator) {
+        describe("checkUsesDefaultOfTranslatable") {
+            it("uses ${Translatable::class.simpleName}'s 111 ${Translatable::getDefault.name}") {
```
robstoll:
```suggestion
            it("uses ${Translatable::class.simpleName}'s ${Translatable::getDefault.name}") {
```

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/translating/TranslationSupplierBasedTranslatorSpec.kt
```diff
@@ -46,27 +49,29 @@ abstract class TranslationSupplierBasedTranslatorSpec(
     }
 
     @Suppress("unused")
-    fun SpecBody.checkUsesDefaultOfTranslatable(testee: Translator) {
-        it("uses ${Translatable::class.simpleName}'s ${Translatable::getDefault.name}") {
-            val result = testee.translate(translatableHello)
-            expect(result).toBe(translatableHello.value)
+    fun ScopeBody.checkUsesDefaultOfTranslatable(testee: Translator) {
+        describe("checkUsesDefaultOfTranslatable") {
```
robstoll: no need for the describe, we are stating the same in the `it`

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/translating/TranslationSupplierBasedTranslatorSpec.kt
```diff
@@ -91,7 +96,9 @@ abstract class TranslationSupplierBasedTranslatorSpec(
         describe("translating a ${Translatable::class.simpleName} to $greatBritain without fallbacks") {
 
             context("no translations provided at all") {
-                val testee = testeeFactory(mock(), greatBritain)
+                val testee = testeeFactory(mockk {
+                    every { get(any(), any()) } returns null
```
robstoll: that does not seem to be right, intended?

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/translating/TranslationSupplierBasedTranslatorSpec.kt
```diff
@@ -4,30 +4,33 @@ import ch.tutteli.atrium.api.fluent.en_GB.toBe
...
-        return mock {
-            on { get(translatable, locale) }.doReturn(translation)
+    fun mockTranslationProvider(locale: Locale, translatable: Translatable, translation: String?): TranslationSupplier {
+        val localeSlot = slot<Locale>()
+        val translatableSlot = slot<Translatable>()
+        return mockk {
+            every { get(capture(translatableSlot), capture(localeSlot)) } answers {
+                // println("translatableSlot.captured.value:" + (translatableSlot.captured as StringBasedTranslatable).value);
+                // println("translatable.value:" + (translatable as StringBasedTranslatable).value);
+                if ((localeSlot.captured != locale) || ((translatableSlot.captured as StringBasedTranslatable).value != (translatable as StringBasedTranslatable).value)) null else translation
```
robstoll: that's a change in behaviour. Could you briefly explain why?

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/translating/TranslationSupplierBasedTranslatorSpec.kt
assaflei: According to spek2 rules, _it_ block should be wrapped inside a describe block

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/translating/TranslationSupplierBasedTranslatorSpec.kt
assaflei: It actually is intended
mockk is more strict than the older mock
What it means is that we have to supply a specification to methods used by the object,
I provided the general implementation of get method in order for it to pass the tests.

Leaving it only by using empty mockk object will fail tests execution

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/translating/TranslationSupplierBasedTranslatorSpec.kt
assaflei: Also related to strictness of the framework
There is no implementation for a single call to _get_ method which was the requirement here
So I had to come up with a way of separating the single scenario required in this scope which is calling _get_ with specific parameters

Hope it explains my point better, if you would like we can have discuss about it

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/translating/TranslationSupplierBasedTranslatorSpec.kt
robstoll: That's true `it` is not even available otherwise. However checkUsesDefaultOfTranslatable is used within a describe and therefore no additional `describe` is necessary. Maybe you have to change the receiver `ScopeBody` in case `it` is not available.

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/reporting/translating/TranslationSupplierBasedTranslatorSpec.kt
robstoll: I see the problem of the strictness and I obviously did not have a good look at your implementation, it is still the same behaviour but I think written in a complicated way. I think the following would be easier to understand:
```
return mockk {
  every { get(any(), any()) } returns null
  every { get(translatable, locale) } returns translation
}
```

assaflei:
I managed to find the right objects to use and the mockk solution was an elegant idea
Thanks for your feedbacks
I'm waiting for further updates

robstoll:
@assaflei thanks a lot for all the migrations :+1: 

The following specs should be migrated in addition (all in atrium-domain-robstoll-lib-jvm):
- LazyThreadUnsafeBasicAssertionSpec
- CharSequenceContainsIndexSearcherSpec
- CharSequenceContainsRegexSearcherSpec

Issue #226, opened 2019-10-18
arrayAssertions for the new infix API


robstoll:
*Platform* (jvm, js, android): jvm
*Extension* (none, kotlin 1.3, jdk8): none

## Code related feature

We want to migrate the `api-cc-infix-en_GB...` modules which are based on `Assert` to `Expect`. We have already created the corresponding
modules `api-infix-en_GB...`. Also there are already `fluent-en_GB...` modules where we have already done the transition from `Assert` to `Expect`

Following the things you need to do:

- copy the file `arrayAssertions.kt` from atrium-api-fluent-en_GB-common to atrium-api-infix-en_GB-common
  - adjust the signatures of the assertion functions so that they are infix functions (see arrayAssertions.kt in atrium-api-cc-infix-en_GB-common for a guideline)
  - change since 0.9.0 to 0.11.0
- copy the `ArrayAsListAssertionsSpec.kt` from atrium-api-fluent-en_GB-common to atrium-api-infix-en_GB-common

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime,
  we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

Kshitij09:
I'll work on this.

robstoll:
@Kshitij09 do you need help?

Kshitij09:
No. Was busy in exams.:sweat_smile: Will work on it now.

robstoll:
@Kshitij09 do you need help now?

Kshitij09:
Actually I'm getting busy in some other stuff. You may unassign me if it has to be implemented urgently. Otherwise, I'll work on it when get the time.

robstoll:
Un-assigning you, so that others can take over. You are welcome to come back and contribute any time (to this or another issue)

Kshitij09:
Sure :+1:

isfedorov:
@robstoll I'll work on this

robstoll:
[...]

PR #418, opened 2020-04-01
add arrayAssertions for the new infix API

isfedorov:
This PR addresses issue #226. @robstoll I have a couple of questions: 1) `change since 0.9.0 to 0.11.0` but currently there is no any `since` version. Should I just add `since 0.11.0` to every function added to `arrayAssertions`? 2) I'm not sure that I've properly updated infix methods to use filler object `o`.  Could you please check it? Also I've left description as is from fluent arrayAssertions. Please notify me if descriptions should be updated.
_______________

### DIFF
--- a/apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/ArrayAsListAssertionsSpec.kt
+++ b/apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/ArrayAsListAssertionsSpec.kt
@@ -27,34 +27,34 @@ class ArrayAsListAssertionsSpec : ch.tutteli.atrium.specs.integration.ArrayAsLis
 ) {
 
     companion object {
-        fun arrayInt(plant: Expect<Array<Int>>) = plant asList o
-        fun byteArray(plant: Expect<ByteArray>) = plant asList o
-        fun charArray(plant: Expect<CharArray>) = plant asList o
-        fun shortArray(plant: Expect<ShortArray>) = plant asList o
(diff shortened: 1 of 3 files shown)


apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/ArrayAsListAssertionsSpec.kt
```diff
@@ -0,0 +1,84 @@
...
+        fun charArray(plant: Expect<CharArray>) = plant asList o
+        fun shortArray(plant: Expect<ShortArray>) = plant asList o
+        fun intArray(plant: Expect<IntArray>) = plant asList o
+        fun longArray(plant: Expect<LongArray>) = plant asList o
+        fun floatArray(plant: Expect<FloatArray>) = plant asList o
+        fun doubleArray(plant: Expect<DoubleArray>) = plant asList o
+        fun booleanArray(plant: Expect<BooleanArray>) = plant asList o
+
+        fun arrayIntWithCreator(plant: Expect<Array<Int>>, assertionCreator: Expect<List<Int>>.() -> Unit) =
+            plant asList assertionCreator
```
robstoll, line 41: please change this line to (not that yours is wrong, I just like to see the syntax as it is used afterwards)
```suggestion
            plant asList { assertionCreator() }
```

robstoll:
about the `@since 0.11.0`: please add it to every function. Would you mind to add `@since 0.9.0` to the fluent API? Seems like we missed it there.

isfedorov:
@robstoll seems I've resolved all the comments you left. So I'm removing [WIP] from the title but if you'll find anything else that should be fixed please ping me

robstoll:
@isfedorov perfect, thanks :+1:

PR #428, opened 2020-04-04
moved from jvm translating to common

assaflei:
I think I moved all possible classes from specs jvm to common


robstoll:
Everything in the reporting.translating package I agree, but not in the reporting package. You don't have to do it in the same PR though. I can also merge this and you can do the rest in a second PR. As you like. In any case, please remove the WIP in case you wish that it is merged (otherwise I assume you are still working on it).

assaflei:
Hi, moving the classes in the domain folder fails during build due to missing spek definitions (Unresolved reference: spekframework), probably related to gradle definition
Can you advice on how to connect the project properly?

robstoll:
It's always easier for me to give answers when I see what was already done. So please push what you already have. 
And I think it would make sense if we split the work into several PRs. Otherwise I will review over and over again the same files. The ones here are already good, so I am going to merge them now.

robstoll:
@assaflei thinks for the first step :+1: 

Please create a second PR with the changes in specs-jvm and specs-common. This one should be easy to review and quick to merge.
Create a third PR with the changes concerning domain and I'll take a look why it fails.

assaflei:
I think I am done with the specs-jvm to specs-common, do you see a class for example to move?
I opened a PR for the domain related classes

robstoll:
you missed classes in ch.tutteli.atrium.specs.reporting

assaflei:
I checked the reporting package and noticed that this dependency will not allow the move:
ToStringObjectFormatter used by AssertionFormatterControllerSpec which uses System.getProperty which requires jvm based projects
Am I correct here?

robstoll:
You should be able to move ToStringObjectFormatter without problems, it does not use any jvm specifics.

Replace separator with lineSeparator

https://github.com/robstoll/atrium/blob/master/misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/testUtils.kt#L216

assaflei:
by replacing seperator with lineSeperator you mean simply rename the variable?

robstoll:
Either assign lineSeparator to separator or delete the variable and use lineSeparator directly

Issue #416, opened 2020-03-31
move non-jvm specific specs to common


robstoll:
*Platform* (jvm, js, android):
*Extension* (none, kotlin 1.3, jdk8): none

## Code related feature
This is a follow up issue for #116 

One additional step in cleaning up specs is to move the non-jvm specific specs to the common module.

Following the things which need to be done:

move from specs-jvm to specs-common (1. PR)
- [ ] everything in ch.tutteli.atrium.specs.reporting and ch.tutteli.atrium.specs.reporting.translating with the exception of TranslatorIntSpec
     - [ ] check the implementation of each of those specs and move to common as well. For instance, once you moved LocaleOrderDeciderSpec to common then you can also move LocaleOrderDeciderSpec in core/robstoll-lib/atrium-core-robstoll-lib-jvm/src/test/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/translating/LocaleOrderDeciderSpec.kt to trium-core-robstoll-lib-common

move from atrium-core-robstoll-lib-jvm to atrium-core-robstoll-lib-common
- [ ] move remaining specs in package checking and reporting (don't move remaining specs in reporting.translating)

move from atrium-domain-robstoll-lib-jvm to atrium-domain-robstoll-lib-common
- [ ] move remaining specs in package creating.charsequence.contains.searchers and in package assertions (without the sub-package filesystem)

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime, we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

assaflei:
Hi, I would like to take this

assaflei:
Hi, moving the classes in the domain folder fails during build due to missing spek definitions (Unresolved reference: spekframework), probably related to gradle definition
Can you advice on how to connect the project properly?

robstoll:
all specs moved which we wanted to move. thanks @assaflei for all the work :+1:

PR #433, opened 2020-04-09
#416 move jvm common

assaflei:
started work on domain classes, currently build fails with unresolved references


robstoll:
atrium/domain/robstoll-lib/atrium-domain-robstoll-lib-js/build.gradle => change 
```
testImplementation prefixedProject('verbs-internal-js')
```
to 
```
testImplementation prefixedProject('specs-js')
```

assaflei:
done with domain, will check specs
btw can you explain shortly why did the change to js gradle did the trick? (if its too complicated to explain its ok)

robstoll:
specs-js defines a dependency on spek as well as verbs-internal-js.

robstoll:
Hard to know without knowing the project. However, most of the time it helps to look how the other platform specific modules are defined. For instance jvm defined a dependency on specs-jvm. That would have been a hint

assaflei:
I understand that connection between jvm to specs-jvm, there is a usage of specs-common in common project which seems to relate to this.
Since building of common project failed, I dont understand why a change in js project resolved the problem

robstoll:
No idea what error you had in a common module, the build failed in js:
https://travis-ci.com/github/robstoll/atrium/jobs/317930541#L1214

So maybe bad luck.

assaflei:
I actually didnt notice that, just saw the file where the error occurred and thought it was related to this project... thanks for pointing that out

Issue #430, opened 2020-04-08
deprecate api-cc-infix add replacements for api-infix


robstoll:
*Platform* (jvm, js, android): all
*Extension* (none, kotlin 1.3, jdk8): all

## Code related feature
The new infix API (api-infix-...) which is based on `Expect` shall replace the current infix api (api-cc-infix-...) which is still based on `Assert`.

Too smooth the migration we want to deprecate all functions in cc-infix and add `ReplaceWith` so that users can use those to switch to the new API easily.

TODO (no need to do everything in one PR, file per file (e.g. PR for anyAssertions, then PR for arrayAssertions.kt etc.) is good as well):
- [ ] add `@Deprecated` annotation with a corresponding `ReplaceWith` -- copy the annotation from api-cc-en_GB and adopt to infix. 
- [ ] replace the usage in the tests.:
  1. For instance, if you have deprecated anyAssertions then go to AnyAssertionsSpec
  2. remove `@file:Suppress("DEPRECATION" /* will be removed with 1.0.0 */)` 
  3. replace the usages with the help of your ReplaceWith (this way we see if it works) -- in IntelliJ usually AL+ENTER ->  `Replace with...` 
   => note, in case you don't see the deprecation due to an error saying something like `symbol is defined in unnamed module` then add the following at the beginning of the file: 
       ```
       @file:Suppress("JAVA_MODULE_DOES_NOT_READ_UNNAMED_MODULE" /* TODO remove once https://youtrack.jetbrains.com/issue/KT-35343 is fixed */)
       ```
       merge with an existing `@file:Suppress` in case there is already one.
  4. re-add `@file:Suppress("DEPRECATION" /* will be removed with 1.0.0 */)` in the end

Miftahunajat:
I would like to work on this

Miftahunajat:
When I merge the 
`@file:Suppress("JAVA_MODULE_DOES_NOT_READ_UNNAMED_MODULE" /* TODO remove once https://youtrack.jetbrains.com/issue/KT-35343 is fixed */)`
with the
`@file:Suppress("DEPRECATION" /* will be removed with 1.0.0 */)`
The deprecation didn't show up.
[...]

PR #434, opened 2020-04-10
#430 Deprecate api-cc-infix add replacements for api-infix anyInsertions

Miftahunajat:
This PR add Deprecate api-cc-infix and add `Replacewith` for the expect in anyInsertions

### DIFF
--- a/apis/cc-infix-en_GB/atrium-api-cc-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/cc/infix/en_GB/anyAssertions.kt
+++ b/apis/cc-infix-en_GB/atrium-api-cc-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/cc/infix/en_GB/anyAssertions.kt
@@ -168,15 +168,6 @@ inline infix fun <reified T : Any> AssertionPlantNullable<T?>.toBeNullIfNullGive
  *
  * @return This plant to support a fluent API.
  */
-@Deprecated(
-    "Switch from Assert to Expect; will be removed with 1.0.0",
-    ReplaceWith(
-        "this.asExpect().and(assertionCreator).asAssert()",
-        "ch.tutteli.atrium.domain.builders.migration.asExpect",
-        "ch.tutteli.atrium.domain.builders.migration.asAssert",
-        "ch.tutteli.atrium.api.infix.en_GB.and"
-    )
-)
 infix fun <T : Any> AssertionPlant<T>.and(assertionCreator: Assert<T>.() -> Unit)
     = addAssertionsCreatedBy(assertionCreator)
 
--- a/apis/cc-infix-en_GB/atrium-api-cc-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/cc/infix/en_GB/anyAssertions.kt
+++ b/apis/cc-infix-en_GB/atrium-api-cc-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/cc/infix/en_GB/anyAssertions.kt
@@ -24,6 +24,15 @@ import kotlin.jvm.JvmName
  * @return This plant to support a fluent API.
  * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
  */
+@Deprecated(
+    "Switch from Assert to Expect; will be removed with 1.0.0",
+    ReplaceWith(
+        "this.asExpect().toBe(expected).asAssert()",
+        "ch.tutteli.atrium.domain.builders.migration.asExpect",
+        "ch.tutteli.atrium.domain.builders.migration.asAssert",
+        "ch.tutteli.atrium.api.infix.en_GB.toBe"
+    )
+)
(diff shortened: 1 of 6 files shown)


Miftahunajat:
The previous build have errors when I add the _deprecated_ annotation.
so I have to try to add `@file:Suppress("DEPRECATION" /* will be removed with 1.0.0 */)` in some Test files.
What is your opinion? @robstoll

robstoll:
Thanks for the PR, I see that I should have prepared a few things beforehand - like suppressing deprecations in the tests of core-common. We will remove those deprecations once we have deprecated cc-infix and migrate the core modules to the new infix API. But that will be an issue we tackle after this one.

I have made the adjustments now. Please rebase on master. Last but not least I would like to see the replacement in action. My description in the issue was not good enough, I have improved it. The goal is to use it in the tests so that we can be sure that they actually work. I should have done that for cc-en_GB but did not have the time. So it could be that we will reveal a few bugs and can fix those. For a starter, I realised that the deprecation of `and` does not really make sense, so I removed it from cc-en_GB, please do the same in your PR.

Miftahunajat:
Okay I will try it out

Miftahunajat:
Hello @robstoll 
Sorry for asking again
Is this replaceWith expression is correct?
`"this.asExpect().toBe(expected).asAssert()"`
Because when I replace some of spec with this. The `.asAssert()` part is deprecated also.
and I'm not sure about to replace it.

robstoll:
Just ask, no problem. `asAssert` is deprecated and you don't have to replace it. That's also the reason why you need to suppress deprecations in the end (re-add `@file: Suppress (...)`). The reason why we have `asAssert` is that one can use own assertion functions which are based on Assert. If one uses such a function in a fluent call chain but still wants to migrate Atrium's functions to the new API which is based on Expect. Unfortunately it's not possible to detect in a ReplaceWith if there is a method call following or not.

Miftahunajat:
When I use the ALT+ENTER shortcut to replace the usages in test.
for example in `toBe` it imports `ch.tutteli.atrium.api.fluent.en_GB.toBe` instead of `ch.tutteli.atrium.api.infix.en_GB.toBe`.
When I try to write the asExpect manually it doesn't have shortcut for toBe in `infix` packages in anyAssertion spec.
am I did something wrong in the steps?

robstoll:
I'll have a look tomorrow. The `import` part sounds like an awkward Kotlin bug - though I have in mind I tried it out as well and it worked. What Intellij version are you using?

robstoll:
I have a wild guess, maybe it's due to the new type inference system. Check in your Intellij preference, search for kotlin compiler and uncheck `enable new...`. see if it works then. If it does then it's good enough. Let me know about the result

Miftahunajat:
I already uncheck the `Enable new type inference algorithm for IDE analysis` but the error still.
When I manually import the `ch.tutteli.atrium.api.infix.en_GB` in anyAssertionsSpec it gives me error Unsolved reference: infix
image

robstoll:
Sorry, totally my fault. I forgot to add the dependency to the new infix API in tests of cc-infix

Miftahunajat:
Oh I wish I knew earlier that the problem is with the dependency.
I will try to add the dependency on this PR then.

robstoll:
Sorry about that, I fixed it for core but forgot the API. You can copy from this commit: https://github.com/robstoll/atrium/commit/47facb86ebfa0340ae7d0a304f05a10738dad1d7

Wait... I the commit already adds it; and I thought I was day dreaming 😅
The problem must be elsewhere. Have you rebased on master?

Miftahunajat:
Thank you
The problem gone as I rebuild it after rebasing.

robstoll:
Ah good. 
Note, in case you rebase again: you need to update intellij to 2020.1

robstoll:
Oho... more things I forgot as it seems. Please add `@file:Suppress("DEPRECATION" /* will be removed with 1.0.0 */)` to:
- bundles/cc-infix-en_GB-robstoll/atrium-cc-infix-en_GB-robstoll-common/src/test/kotlin/SmokeTest.kt
- bundles/cc-infix-en_GB-robstoll/atrium-cc-infix-en_GB-robstoll-jvm/src/test/kotlin/custom/SmokeSpec.kt

robstoll:
@Miftahunajat thanks for the first step in deprecating cc-infix. Looking forward to the next PR :slightly_smiling_face:

Miftahunajat:
Okay, since there are still many files left. May I will create several PR with several files.

Miftahunajat:
Hello @robstoll sorry for late response
I have a small problem with deprecating arrayAssertion on atrium-api-cc-infix-en_GB.
When I replace the usage of `asIterable` with `this.asExpect().asList().asAssert()` (I copied from atrium-api-infix-en_GB) it gives error message, because there is no such function on `asList()` without any argument on it.
Is there a function for an `asList()` call without any argument on it?
or should I replace the call with `asList(o)`? because there is function for that.

robstoll:
No worries. Please use `asList(o)` and don't forget to add the import for `o` to the replacement

Issue #245, opened 2019-10-22
Build samples as composite build


robstoll:
## Dev-op related feature
We have currently two js projects in /samples. It would be nice if we could create the projects as independent projects, like we did for samples/jvm/spek and samples/jvm/junit5. Meaning with an own settings.gradle, build.gradle and gradlew so that one can really just copy the project and immediately start.

Those projects have to use a published version of Atrium and cannot use the current development version.
However, it would be nice to know if the current development version still works for those sample projects. 

Following the things you need to do:

* [ ] setup a module in misc/tools/atrium-samples-test (add folder + include it in settings.gradle in root)
  * [ ] see https://docs.gradle.org/current/userguide/composite_builds.html for how to setup a composite build and test it with the existing independent projects: samples/jvm/spek, samples/jvm/junit5 or samples/multiplatform
* [ ] turn samples/js/jasmine into an independent project
  - [ ] remove it from settings.gradle in root 
  - [ ] add own settings.gradle, gradle.build, gradlew etc. in this project
  - [ ] include it in the composite build
* [ ] turn samples/js/mocha into an independent project
  - [ ] remove it from settings.gradle in root 
  - [ ] add own settings.gradle, gradle.build, gradlew etc. in this project
  - [ ] include it in the composite build
* [ ] search for `TODO #245` and resolve them

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime,
  we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

iamshubh:
Hey, assign it to me

robstoll:
@iamshubh do you need help?

iamshubh:
> @iamshubh do you need help?

Did not get time to loom into it, will check this out!

robstoll:
@iamshubh are you still working on it?

assaflei:
I'll would like to take this

PR #435, opened 2020-04-10
#416 move jvm common

assaflei:
I switched to lineSeperator and moved ToStringObjectFormatter
Tried moving other classes but encountered errors due to using System or KClass reflection


robstoll:
I guess you would like to tackle the problem, so here my hints:
As far as I can see TextListBasedAssertionGroupFormatterSpec is using `Class` instead of `KClass`, that's a problem for common. You could rewrite TextListBasedAssertionGroupFormatterSpec to use KClass instead. However, it depends on TranslatorIntSpec which needs a bit more in order that it can be moved, so I would do this in another PR. Which means you can keep TextListBasedAssertionGroupFormatterSpec and TextListAssertionGroupFormatterSpec in specs-jvm (and TextListAssertionGroupFormatterSpec in atrium-core-robstoll-lib-jvm). Same applies to TextFeatureAssertionGroupFormatterSpec, it also depends on TranslatorIntSpec. However, you should be able to move all other classes.

The problem with KClass reflection can be solved by using `fullName` instead of `qualifiedName!!`

I am not sure what you mean with 
> errors due to using System

assaflei:
Moved all other classes, I will now try to handle TranslatorIntSpec and related classes.
Regarding System namespace, there are classes such as core\robstoll-lib\atrium-core-robstoll-lib-jvm\src\test\kotlin\ch\tutteli\atrium\core\robstoll\lib\reporting\DetailedObjectFormatterSpec.kt using commands such as `System.identityHashCode(string)` which requires them to stay at jvm project. We discussed about similar commands before (System.getProperty) and I left these classes in place.

robstoll:
:+1: merged this already so that we don't run into conflicts.

robstoll:
> Regarding System namespace, there are classes such as core\robstoll-lib\atrium-core-robstoll-lib-jvm\src\test\kotlin\ch\tutteli\atrium\core\robstoll\lib\reporting\DetailedObjectFormatterSpec.kt using commands such as System.identityHashCode(string) which requires them to stay at jvm project. 

I see, keep DetailedObjectFormatterSpec in the jvm module then.

robstoll:
Note, you need to update IntelliJ to 2020.1 when you base your next PR on master.

robstoll:
> I will now try to handle TranslatorIntSpec and related classes.

IMO thos os not a must. We need to rework translation anyway (it is too much JVM based) and therefore it would be totally fine to leave it and dependent classes in JVM.

assaflei:
> Note, you need to update IntelliJ to 2020.1 when you base your next PR on master.

I'm using VS code so not relevant

> > I will now try to handle TranslatorIntSpec and related classes.
> 
> IMO this is not a must. We need to rework translation anyway (it is too much JVM based) and therefore it would be totally fine to leave it and dependent classes in JVM.

So we can close the issue don't we?

robstoll:
> I'm using VS code so not relevant

Interesting, what's your experience so far?

> So we can close the issue don't we?

Might well be, I'll check later if we have covered everything.

robstoll:
all done, I closed the issue. I could use your help for other issues, for instance: 
https://github.com/robstoll/atrium/issues/245 would be great, or one which is maybe a bit tougher but actually more valuable https://github.com/robstoll/atrium/issues/167
But there are more:

assaflei:
Regarding the VS code, the experience is good since its working and not only for java related project.
I love the idea of one app for all types of devs.
Although there is the issue of not knowing what I'm missing by not using specialized tool for this eco system.

Happy to help with closing of the issue, I'll try to upgrade the level step by step so will check 245 first

PR #440, opened 2020-04-13
[WIP] Samples composite build

assaflei:
Started following instructions
How do I test the setup?

In misc\tools\atrium-samples-test\build.gradle I simply copied the file from atrium-bc-test and renamed jvm to js hope its enough
I think that this comment in this file is redundand, is it?
`            //TODO tests are now mostly in the common module. However, we cannot unzip this to the jvm project but
            // require a common module for it. Probably easiest to introduce a second bc-test which is an MPP project
//            add(testJarSources, "$groupId:atrium-api-$apiName-common:$oldVersion:testsources") {
//                exclude group: '*'
//            }`


robstoll:
I see, the first point in the issue lacks a possibility to test the setup. I have updated the description; merged the first and the last point.
atrium-bc-test was only meant as guideline but maybe it is more confusing than helping as it does very specific things -- e.g. copy old versions of Atrium into src/old, you won't do anything like that in your module. Maybe it is better to do it from scratch, i.e. only follow the guideline of gradle.

assaflei:
got back to this after some time...

tried starting it
it seems like the js modules are dependent on fluent-en_GB-js build
it looks like a tight coupling that breaks the standalone requirement
am I correct here?
currently the build containing jasmine fails with the message:   Project with path 'ch.tutteli.atrium:atrium-fluent-en_GB-js:0.11.0' could not be found in project ':jasmine'.

what do you think?

assaflei:
I removed the dependency in atrium-fluent-en_GB-js project and the execution works
meaning running gradlew run succeed

regarding the multiplatform project, which task should be executed there?

assaflei:
I think I managed to execute all tests eventually and its successful

misc/tools/atrium-samples-test/build.gradle
```diff
@@ -0,0 +1,7 @@
+task run {
+    dependsOn gradle.includedBuild('junit5').task(':test')
```
robstoll, line 2: I am afraid but I made a wrong assumption. I did not have a look at how composite builds work in gradle but they work the opposite way. I.e. the project which depends on another needs to do the includeBuild. Which means, we don't need a `atrium-samples-test` at all because not atrium has a dependency but the opposite. Thus remove the whole `atrium-samples-test` again. Instead it is enough if we add the `--include-build` to the CI-definitions (.travis.yml and java-windows.yml). So basically duplicate the existing build-steps/jobs and add --include-build. for instance, for mpp in java-windows.yml:
```
            -   name: Build sample mpp with current Atrium
                run: samples\multiplatform\gradlew -p samples\multiplatform build --include-build ../../../
```

robstoll: Wait wait... maybe gradle is wrong (let's hope). Maybe this approach works but you surely need to include Atrium as well (and make atrium-samples-test an independent project entirely, i.e. do not add it to settings.gradle in the root of the project). In order to test if it works, I advice you to change apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt use the following for toBe
```
fun <T> Expect<T>.toBe(expected: T) = addAssertion(ExpectImpl.any.notToBe(this, expected))
```
if the composite build works, then the tests should be executed and fail.

assaflei: what is the meaning of ../../../?
where are we pointing to?
the include build was required in case we wanted a composite build, if we wont use it than why use --include-build?

assaflei:
Lets go for the next steps as you require, I'll learn whats required
Update me when the env is ready

robstoll:
> Lets go for the next steps as you require, I'll learn whats required
Cool, I'll look a bit more into composite builds and let you know once pre-work is done.

misc/tools/atrium-samples-test/build.gradle
assaflei: > Wait wait... maybe gradle is wrong (let's hope). Maybe this approach works but you surely need to include Atrium as well (and make atrium-samples-test an independent project entirely, i.e. do not add it to settings.gradle in the root of the project). In order to test if it works, I advice you to change apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt use the following for toBe
> 
> ```
> fun <T> Expect<T>.toBe(expected: T) = addAssertion(ExpectImpl.any.notToBe(this, expected))
> ```
> 
> if the composite build works, then the tests should be executed and fail.

 I missed this comment...
Should we revert?

robstoll: > the include build was required in case we wanted a composite build, if we wont use it than why use --include-build?

Sorry for the confusion, that comment was before I saw the restriction of composite builds.

But to answer your other question

> what is the meaning of ../../../?

note the parameter -p in 
```
samples\multiplatform\gradlew -p samples\multiplatform build --include-build ../../../
```
so `../../../` is pointing back to the root of the project.

assaflei: arent we running from the root already? 
samples\... looks like reference from the root

robstoll: `-p` specifies the working directory for gradle and --include-build is relative to this working directory.

assaflei: understood
so leave the yml files without change or should I try and add the include-build param?

robstoll: I'll see if I can build up on your current work, tweaking gradle a bit so that we still can use composite build. But this would require `atrium-samples-test`. So for the moment you can wait until I get back to you.

robstoll: What you can do already in the meantime is:
- squash your commits, rebase on master and force push

robstoll: Nice, I have a working workaround, so you can proceed:
- add the following to settings.gradle:
   ```
    includeBuild('../../../') {
        dependencySubstitution {
            gradle.includedBuild('atrium').getAvailableModules().collect { pair -> pair.right.projectName }
                .findAll { it.endsWith("-jvm") }
                .forEach { projectName ->
                    def withoutSuffix = projectName.substring(0, projectName.length() - 4)
                    substitute module("ch.tutteli.atrium:$withoutSuffix") with project(":$projectName")
                }
        }
    }
    ```
- in build.gradle, change the dependsOn and always depend on `:build` (instead of test etc.)

robstoll: I can also merge your changes into a separate branch and push my changes, if you prefer (and then you can take up the work from there)

robstoll:
Nice, I have a working workaround, so you can proceed:
- add the following to settings.gradle:
   ```
    includeBuild('../../../') {
        dependencySubstitution {
            gradle.includedBuild('atrium').getAvailableModules().collect { pair -> pair.right.projectName }
                .findAll { it.endsWith("-jvm") }
                .forEach { projectName ->
                    def withoutSuffix = projectName.substring(0, projectName.length() - 4)
                    substitute module("ch.tutteli.atrium:$withoutSuffix") with project(":$projectName")
                }
        }
    }
    ```
- in build.gradle, change the dependsOn and always depend on `:build` (instead of test etc.)
- change .travis.yml and java-windows.yml remove steps/jobs for samples and step/job for `atrium-samples-test` instead.

I can also merge your changes into a separate branch and push my changes, if you prefer (and then you can take up the work from there)

assaflei:
which settings.gradle do you mean? at the root? or of the js and jvm projects we address?
also just point me to the branch you used so i could merge changes into my own branch

robstoll:
I meant settingsgradle of atrium-samples-test and build.gradle of atrium-samples-test

robstoll:
You can see my changes here (note that it is not based on your latest commits):
https://github.com/robstoll/atrium/tree/pull/440
https://github.com/robstoll/atrium/commit/1ba255a93a7f820e2a30876b5441fd4c915e3f92

assaflei:
updated as required, whats next?

robstoll:
squash your commits, rebase on master and force push
I'll review in the meantime

robstoll:
Looks good, the only thing left is that you have merge conflicts you need to resolve

assaflei:
> squash your commits, rebase on master and force push
> I'll review in the meantime

trying to understand this first... never done before

robstoll:
I usually do it as follows (from top of my head, errors included :wink:)
get first commit hash (e.g. from here https://github.com/robstoll/atrium/pull/440/commits)
then:
```
git reset --soft f6e09c69a6cc4b3561cc23d3ba7d8fd2e18b989d
```
takes you back to f6e09c69a6cc4b3561cc23d3ba7d8fd2e18b989d and adds all modification to the staged changes (do a `git status` to see that)
next we want to include those changes in the current commit
```
git commit --amend -m "composite build, make mocha/jasmine independent"
```
now you have one single commit (see `git log`)
```
git rebase upstream/master
```
Here you will get a conflict in settings.gradle.
Open settings.gradle and fix it
```
git add settings.gradle
git rebase --continue
```
now your commit should be on master
```
git push -f
```
force push your branch, i.e. let the branch-pointer switch from the previous commit hash to the new one

assaflei:
Followed your instructions and its all good, thank you!
What is required next?

settings.gradle
```diff
@@ -121,10 +121,7 @@ include {
     }
 
     samples {
-        js {
-            project('mocha')
-            project('jasmine')
-        }
+        project('readme-examples')
```
robstoll, line 124: This was moved to tools, you can remove the whole `samples` block

robstoll:
I think it should be good know but let's wait to see if the build is green.

robstoll:
you need to give `gradlew` in `atrium-samples-tests` execution rights, the build failed with
```
/home/travis/.travis/functions: line 109: misc/tools/atrium-samples-test/gradlew: Permission denied
```
this can be done as follows:
```
cd misc/tools/atrium-samples-test
chmod +x gradlew
```
commit and push afterwards

assaflei:
add this to this file gradlew?

robstoll:
No worries, I guess you are a Windows user and it might be not trivial to do it there. I'll do it.

assaflei:
So is this issue closed?

robstoll:
I created a new PR: https://github.com/robstoll/atrium/pull/471 let's see if the build passes now. If it does then everything is done :+1:

assaflei:
> @assaflei you need to give `gradlew` in `atrium-samples-tests` execution rights, the build failed with
> 
> ```
> /home/travis/.travis/functions: line 109: misc/tools/atrium-samples-test/gradlew: Permission denied
> ```
> 
> this can be done as follows:
> 
> ```
> cd misc/tools/atrium-samples-test
> chmod +x gradlew
> ```
> 
> commit and push afterwards

just to understand the context, you wanted to add the commands to gradlew file in order for them to run as part of the travis build?

robstoll:
travis tries to execute gradlew as part of the build (that's what you have set up: `misc/tools/atrium-samples-test/gradlew -p ./misc/tools/atrium-samples-test build`), In order that it can do that, the file needs execution rights; otherwise travis fails because it is not allowed to execute it. 
It's a bit unfortunate, but you cannot see the change I did here:
https://github.com/robstoll/atrium/pull/471/commits/81881bc6329f521d509d2be886f1beadac4d8398
but its basically setting the execution flag with `chmod +x gradlew`

robstoll:
But you can see it when you do: `git show 81881bc`
```
commit 81881bc6329f521d509d2be886f1beadac4d8398 (origin/#245-samples-composite-build, #245-samples-composite-build)
Author: Robert Stoll <rstoll@tutteli.ch>
Date:   Sat Apr 25 21:25:13 2020 +0200

    add execution rights to misc/tools/atrium-samples-test/gradlew

diff --git a/misc/tools/atrium-samples-test/gradlew b/misc/tools/atrium-samples-test/gradlew
old mode 100644
new mode 100755
```
See, the mode of the file changed from 644 to 755. For more information about the modes: https://en.wikipedia.org/wiki/Modes_(Unix)

assaflei:
I didnt see any line of code related to this change
You just changed the setting on your own machine and the file was marked and committed with the new permission?

robstoll:
Exactly, I changed the file mode, committed that change as you would commit any other change and pushed it. GitHub (and many IDEs) are lacking the feature of showing file mode changes. That's unfortunate but in the end `git` is the only truth :wink:

robstoll:
btw. the other PR passed, thanks for your contribution :+1: 

There is #470 which deals with a sample project, so kind of in the same context. Would you like to tackle this one? Or perhaps more Kotlin related? For instance 460 would be a good start.

assaflei:
throw whatever is more important :)
this is a great learning experience for me

robstoll:
In this case, it would be great if you could tackle https://github.com/robstoll/atrium/issues/167

Issue #425, opened 2020-04-04
shortcut Iterable.containsElementsOf


robstoll:
*Platform* (jvm, js, android): all
*Extension* (none, kotlin 1.3, jdk8): none

## Code related feature
```
expect(listOf(1, 2)).containsElementsOf(anotherList)

//instead of

expect(listOf(1, 2)).contains.inAnyOrder.atLeast(1).elementsOf(anotherList)
```

Please react with :+1:  in case you would like to see this shortcut implemented. The more upvotes the more likely I will implement it myself. You are of course welcome to implement it yourself and create a PR. Write `I'll work on it` as comment so that we can assign the task to you.

tkech17:
i'll work on it

robstoll:
Since it is your first contributions, I recommend that you have a look at [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution).

And a few pointers what needs to be done:
*fluent api*
- [ ] add the shortcut to fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/iterableAssertions.kt
   - [ ] add `@since 0.11.0` to KDoc
- [ ] add a ShortcutSpec to IterableContainsInOrderOnlyElementsOfAssertionSpec
  - [ ] also add a test for passing an empty iterable

*infix api* same same but different

robstoll:
And please rebase on master in case you already forked, I made a cleanup specific for this task.

tkech17:
I have question about adding ShortcutSpec. 
why do i need to add ShortcutSpec to IterableContainsInOrderOnlyElementsOfAssertionSpec,
as the method test inAnyOrder contains.
or maybe i did not understand what "add a ShortcutSpec to IterableContainsInOrderOnlyElementsOfAssertionSpec"

robstoll:
Thanks for the question, the example in the description was wrong. This issue want to replace `expect(listOf(1, 2)).contains.inOrder.only.elementsOf(anotherList)`

robstoll:
Ah.. no sorry, my fault, wrong issue, the example was correct. In this case you need to place the the ShortcutSpec into IterableContainsInAnyOrderAtLeast1ElementsOfAssertionsSpec

Issue #131, opened 2019-08-27
shortcut Iterable.containsExactlyElementsOf


robstoll:
*Platform* (jvm, jdk8, js, android): all

## Code related feature
```
expect(listOf(1, 2)).containsExactlyElementsOf(anotherList)

//instead of

expect(listOf(1, 2)).contains.inOrder.only.elementsOf(anotherList)
```

Please react with :+1:  in case you would like to see this shortcut implemented. The more upvotes the more likely I will implement it myself. You are of course welcome to implement it yourself and create a PR. Write `I'll work on it` as comment so that we can assign the task to you.

robstoll:
@audkar do you mind sharing your use case for this functionality?

audkar:
For me this is probably one of the most often used checks on collections. I have tests which assert that all UI states changed in right order after some events were emitted. e.g.
```
expect(recordedStates).containsExactlyElementsOf(listOf(Idle, IsLoading, ErrorShown))
```

robstoll:
Are you aware of that you can write the above as:
```
expect(recordedStates).containsExactly(Idle, IsLoading, ErrorShown)
```
I guess the list is somewhat dynamic right?

audkar:
Actually no. Was using Google's `Truth` before and was not expecting that `containsExactly` is checking order :)
So for my usecase `ElementsOf` part is not necessary then :)))

robstoll:
I see, Atrium follows AssertJ's behaviour here (see [api diff - iterable contains](https://github.com/robstoll/atrium/blob/master/apis/differences.md#iterable-contains) for a compact view of Iterable contains features).

 Good to have a Truth user here, let me know if you miss a feature in Atrium by opening an issue

tkech17:
i saw this topic, and i tought is was not active anymore(from comments).
if it's still active assign me

PR #442, opened 2020-04-14
[#425] shortcut Iterable.containsElementsOf implemented

tkech17:
Iterable.containsElementsOf implemented


robstoll:
@tkech17 thanks for your first contribution :slightly_smiling_face: :+1:
how about a second one, https://github.com/robstoll/atrium/issues/131 would be very similar

PR #446, opened 2020-04-15
unused parameter warning suppressed

tkech17:


misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/OptionalAssertionsSpec.kt
```diff
@@ -64,7 +64,7 @@ abstract class OptionalAssertionsSpec(
                 }
             }
 
-            isPresentFunctions.forEach { (name, isPresentFun, hasExtraHint) ->
+            isPresentFunctions.forEach { (name, isPresentFun, @Suppress("UNUSED_DESTRUCTURED_PARAMETER_ENTRY") hasExtraHint) ->
```
robstoll, line 67: please change to 
```suggestion
            isPresentFunctions.forEach { (name, isPresentFun, _) ->
```

PR #448, opened 2020-04-15
shortcut_containsElementsOf code cleanup

tkech17:


apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/IterableContainsInAnyOrderAtLeast1ElementsOfAssertionsSpec.kt
```diff
@@ -59,16 +59,16 @@ class IterableContainsInAnyOrderAtLeast1ElementsOfAssertionsSpec : Spek({
...
+            Expect<Iterable<Double>>::containsElementsOf
 
         private fun getContainsShortcutPair() = containsElementsOfShortcutFun.name to Companion::containsInAnyOrderShortcut
 
         private fun containsInAnyOrderShortcut(
             expect: Expect<Iterable<Double>>,
             a: Double,
             aX: Array<out Double>
-        ): Expect<Iterable<Double>> = expect.containsElementsOf(a glue aX)
+        ): Expect<Iterable<Double>> = expect.containsElementsOf(listOf(a, *aX))
```
robstoll, line 71: sorry, I only meant in the infix API, here could have kept it but it is also OK this way

Issue #235, opened 2019-10-19
setup code coverage for windows


robstoll:
## Dev-Ops related feature

We have some Windows specific code due to Path assertions => It would be nice to see them covered in the code coverage report as well. Thus:

- adjust .github/workflows/java-windows.yml 
  => see https://github.com/codecov/codecov-action for the setup of codecov (the secrret CODECOV_TOKEN is already setup)
  - please use `flags: current-windows` 
  - coverage files are created per subproject and if codecov works the same way as it does for travis-ci, then it should detect them automatically (no need to specify `file: ...`)

sanatik:
Read about GitHub actions recently, sounds interesting for me.
I'll work on this.

robstoll:
@sanatik do you need any help?

robstoll:
@sanatik are you still working on this one? I am un-assigning you so that others can take over. let me know if you want to give it a go again and I re-assign you.

name213:
Hello, If its up for grabs I would like to analyse and do this

robstoll:
It is, go ahead 🙂👍

robstoll:
@ivanmiklec let me know in case you need help

PR #452, opened 2020-04-19
setup codecov upload for windows workflow #235

ivanmiklec:
Hey, I think this should do for windows code coverage #235 . I didn't specify file but apparently codecov github actions only support .xml and not .codecov type of file, so if its the latter one it shouldn't work, but let's see.

Also added `fail_ci_if_error: true` so the whole CI fails if codecov upload fails, rather than blocking PR because coverage dropped, if you think that is not needed I will remove it :).

EDIT: just checked, looks like jacoco reports are .xml

### DIFF
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -24,7 +24,7 @@ jobs:
                 uses: codecov/codecov-action@v1
                 with:
                     token: ${{ secrets.CODECOV_TOKEN }}
-                    flags: current-windows
+                    flags: current_windows
                     fail_ci_if_error: true
 
             -   name: build sample atrium+spek project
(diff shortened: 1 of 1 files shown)


ivanmiklec:
Looks like flag needs to match regex` ^[\w\,]+$` so i changed current-windows to current_windows

robstoll:
Perfect, thanks :+1: 

https://codecov.io/gh/robstoll/atrium/src/f24291c811cae2b5af42d99d093439392e71577e/domain/robstoll-lib/atrium-domain-robstoll-lib-jvm/src/main/kotlin/ch/tutteli/atrium/domain/robstoll/lib/creating/pathAssertions.kt

is now as it should be

PR #468, opened 2020-04-23
#466 Added validation on TranslatableWithArgs

### LINKED ISSUE #466: fail if a TranslatableWithArgs has not exact arguments
*Platform* (jvm, js, android): all
*Extension* (none, kotlin 1.3, jdk8): none

## Code related feature
We currently do not check if the number of given arguments corresponds to the number of expected placeholders in the translatable. It would be nice if we throw an IllegalArgumentException if they do not match. This way it is more likely that we detect wrong translations.

tkech17:
resolves #466


robstoll:
Please create a new branch based on current master and cherry-pick your changes. Your PR contains commits which are already in master.

tkech17:
i make reset on this branch and pushed only this commit. I think for this task it's not problem, right?

robstoll:
That's fine as well, yes, please also fix the French translation and other places (search for DATE_KNOWN)

PR #487, opened 2020-04-29
android dex compilation

assaflei:
Started working on the task
@robstoll I noticed that getAndroidProjects() returns an empty list, is this correct?

### DIFF
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -27,7 +27,7 @@ jobs:
                     if ($file_sha256 -Ne "cdc1846376a14b0370cc63454a129606b4a52cc50ada75ef0d4cf956b1ad2daa") {
                         exit -1
                     }
-                    ./gradlew buildAndroidProjects checkDexer
+                    ./gradlew checkDexer
                 env:
                     CI: true
                     ATRIUM_ANDROID_JAR: android.jar
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -16,6 +16,10 @@ jobs:
                 uses: actions/setup-java@v1
                 with:
                     java-version: ${{ matrix.java_version }}
+            -   name: buildNonDeprecatedJvm
+                run: ./gradlew buildNonDeprecatedJvm
+                env:
+                  CI: true
             -   name: check dexer
                 run: |
                     curl -o android.jar -L "https://github.com/Sable/android-platforms/blob/master/android-26/android.jar?raw=true"
@@ -27,10 +31,6 @@ jobs:
                 env:
                     CI: true
                     ATRIUM_ANDROID_JAR: android.jar
-            -   name: buildNonDeprecatedJvm
-                run: ./gradlew buildNonDeprecatedJvm
-                env:
-                  CI: true
             -   name: Upload windows build code coverage
                 uses: codecov/codecov-action@v1
                 with:
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -28,7 +28,7 @@ jobs:
                     ./gradlew checkDexer
                 env:
                     CI: true
-                    ATRIUM_ANDROID_JAR: ./android.jar
+                    ATRIUM_ANDROID_JAR: android.jar
                 shell: powershell
             -   name: buildNonDeprecatedJvm
                 run: ./gradlew buildNonDeprecatedJvm
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -27,7 +27,7 @@ jobs:
                     if ($file_sha256 -Ne "cdc1846376a14b0370cc63454a129606b4a52cc50ada75ef0d4cf956b1ad2daa") {
                         exit -1
                     }
-                    ./gradlew build checkDexer
+                    ./gradlew buildAndroidProjects checkDexer
                 env:
                     CI: true
                     ATRIUM_ANDROID_JAR: android.jar
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -23,6 +23,8 @@ jobs:
                     if ($file_sha256 -Ne "4c438bf1b3d4e6104faccc62e1ffe95ef197b88055497b540a871349e93c8685") {
                         exit -1
                     }
+                    Write-Host "android.jar downloaded"
+                    ls
                     ./gradlew checkDexer
                 env:
                     CI: true
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -20,6 +20,7 @@ jobs:
                 run: ./gradlew buildNonDeprecatedJvm
                 env:
                   CI: true
+                  ATRIUM_ANDROID_JAR: ./android.jar
             -   name: Upload windows build code coverage
                 uses: codecov/codecov-action@v1
                 with:
@@ -45,3 +46,13 @@ jobs:
 
             -   name: composite build atrium-samples
                 run: misc\tools\atrium-samples-test\gradlew -p misc\tools\atrium-samples-test build
+
+            -   name: check dexer
+                run: |
+                    wget "https://github.com/Sable/android-platforms/blob/master/android-28/android.jar?raw=true" -P "./android.jar"
+                    ./gradlew checkDexer
+                shell: bash
+                env:
+                    CI: true
+                    ATRIUM_ANDROID_JAR: ./android.jar
+
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -18,18 +18,15 @@ jobs:
                     java-version: ${{ matrix.java_version }}
             -   name: check dexer
                 run: |
-                    Invoke-WebRequest -Uri "https://github.com/Sable/android-platforms/blob/master/android-24/android.jar?raw=true" -OutFile "./android.jar"
+                    curl -o android.jar -L "https://github.com/Sable/android-platforms/blob/master/android-24/android.jar?raw=true"
                     $file_sha256 = $(certutil -hashfile android.jar sha256)[1] -replace " ",""
                     if ($file_sha256 -Ne "4c438bf1b3d4e6104faccc62e1ffe95ef197b88055497b540a871349e93c8685") {
                         exit -1
                     }
-                    Write-Host "android.jar downloaded"
-                    ls
                     ./gradlew checkDexer
                 env:
                     CI: true
                     ATRIUM_ANDROID_JAR: android.jar
-                shell: powershell
             -   name: buildNonDeprecatedJvm
                 run: ./gradlew buildNonDeprecatedJvm
                 env:
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -16,6 +16,14 @@ jobs:
                 uses: actions/setup-java@v1
                 with:
                     java-version: ${{ matrix.java_version }}
+            -   name: check dexer
+                run: |
+                    $client = new-object System.Net.WebClient
+                    $client.DownloadFile("https://github.com/Sable/android-platforms/blob/master/android-28/android.jar?raw=true", "./android.jar")
+                    ./gradlew checkDexer
+                env:
+                    CI: true
+                    ATRIUM_ANDROID_JAR: ./android.jar
             -   name: buildNonDeprecatedJvm
                 run: ./gradlew buildNonDeprecatedJvm
                 env:
@@ -28,15 +36,6 @@ jobs:
                     flags: current_windows
                     # too many timeout errors, let's not fail at the moment
                     #fail_ci_if_error: true
-
-            -   name: check dexer
-                run: |
-                    $client = new-object System.Net.WebClient
-                    $client.DownloadFile("https://github.com/Sable/android-platforms/blob/master/android-28/android.jar?raw=true", "android.jar")
-                    ./gradlew checkDexer
-                env:
-                    CI: true
-                    ATRIUM_ANDROID_JAR: ./android.jar
 
             -   name: build jasmine sample
                 run: samples\js\jasmine\gradlew -p samples\js\jasmine build
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -17,10 +17,8 @@ jobs:
                 with:
                     java-version: ${{ matrix.java_version }}
             -   name: check dexer
-                run: |
-                    $client = new-object System.Net.WebClient
-                    $client.DownloadFile("https://github.com/Sable/android-platforms/blob/master/android-28/android.jar?raw=true", "./android.jar")
-                    ./gradlew checkDexer
+                run: Invoke-WebRequest -Uri "https://github.com/Sable/android-platforms/blob/master/android-28/android.jar?raw=true" -OutFile "./android.jar"
+                run: ./gradlew checkDexer
                 env:
                     CI: true
                     ATRIUM_ANDROID_JAR: ./android.jar
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -18,9 +18,9 @@ jobs:
                     java-version: ${{ matrix.java_version }}
             -   name: check dexer
                 run: |
-                    curl -o android.jar -L "https://github.com/Sable/android-platforms/blob/master/android-24sha/android.jar?raw=true"
+                    Invoke-WebRequest -Uri "https://github.com/Sable/android-platforms/blob/master/android-24/android.jar?raw=true" -OutFile "./android.jar"
                     $file_sha256 = $(certutil -hashfile android.jar sha256)[1] -replace " ",""
-                    if ($file_sha256 -Eq "4c438bf1b3d4e6104faccc62e1ffe95ef197b88055497b540a871349e93c8685") {
+                    if ($file_sha256 -Ne "4c438bf1b3d4e6104faccc62e1ffe95ef197b88055497b540a871349e93c8685") {
                         exit -1
                     }
                     ./gradlew checkDexer
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -17,8 +17,13 @@ jobs:
                 with:
                     java-version: ${{ matrix.java_version }}
             -   name: check dexer
-                run: Invoke-WebRequest -Uri "https://github.com/Sable/android-platforms/blob/master/android-28/android.jar?raw=true" -OutFile "./android.jar"
-                run: ./gradlew checkDexer
+                run: |
+                    curl -o android.jar -L "https://github.com/Sable/android-platforms/blob/master/android-24sha/android.jar?raw=true"
+                    $file_sha256 = $(certutil -hashfile e:\temp\android.jar sha256)[1] -replace " ",""
+                    if ($file_sha256 -Eq "4c438bf1b3d4e6104faccc62e1ffe95ef197b88055497b540a871349e93c8685") {
+                        exit -1
+                    }
+                    ./gradlew checkDexer
                 env:
                     CI: true
                     ATRIUM_ANDROID_JAR: ./android.jar
@@ -26,7 +31,6 @@ jobs:
                 run: ./gradlew buildNonDeprecatedJvm
                 env:
                   CI: true
-                  ATRIUM_ANDROID_JAR: ./android.jar
             -   name: Upload windows build code coverage
                 uses: codecov/codecov-action@v1
                 with:
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -19,7 +19,7 @@ jobs:
             -   name: check dexer
                 run: |
                     curl -o android.jar -L "https://github.com/Sable/android-platforms/blob/master/android-24sha/android.jar?raw=true"
-                    $file_sha256 = $(certutil -hashfile e:\temp\android.jar sha256)[1] -replace " ",""
+                    $file_sha256 = $(certutil -hashfile android.jar sha256)[1] -replace " ",""
                     if ($file_sha256 -Eq "4c438bf1b3d4e6104faccc62e1ffe95ef197b88055497b540a871349e93c8685") {
                         exit -1
                     }
@@ -27,6 +27,7 @@ jobs:
                 env:
                     CI: true
                     ATRIUM_ANDROID_JAR: ./android.jar
+                shell: powershell
             -   name: buildNonDeprecatedJvm
                 run: ./gradlew buildNonDeprecatedJvm
                 env:
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -18,9 +18,9 @@ jobs:
                     java-version: ${{ matrix.java_version }}
             -   name: check dexer
                 run: |
-                    curl -o android.jar -L "https://github.com/Sable/android-platforms/blob/master/android-24/android.jar?raw=true"
+                    curl -o android.jar -L "https://github.com/Sable/android-platforms/blob/master/android-26/android.jar?raw=true"
                     $file_sha256 = $(certutil -hashfile android.jar sha256)[1] -replace " ",""
-                    if ($file_sha256 -Ne "4c438bf1b3d4e6104faccc62e1ffe95ef197b88055497b540a871349e93c8685") {
+                    if ($file_sha256 -Ne "cdc1846376a14b0370cc63454a129606b4a52cc50ada75ef0d4cf956b1ad2daa") {
                         exit -1
                     }
                     ./gradlew checkDexer
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -29,6 +29,15 @@ jobs:
                     # too many timeout errors, let's not fail at the moment
                     #fail_ci_if_error: true
 
+            -   name: check dexer
+                run: |
+                    $client = new-object System.Net.WebClient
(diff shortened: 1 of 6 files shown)


robstoll:
> @robstoll I noticed that getAndroidProjects() returns an empty list, is this correct?

Good that you ask, android projects are only included in CI due to intellij problems. 
You have to set an environment variable CI = true (alternatively adjust settings.gradle line 17 to `if (true || System.getenv('CI')) {` for local testing.)

assaflei:
I managed to get a process running, there are few issues requiring attention:
- Currently I created a folder on the root called **d8out** that will contain resulting dex files
- Running build for android projects fails with this error
`
Execution failed for task ':atrium-cc-de_CH-robstoll-android:test'.
> No tests executed, most likely the discovery failed.
`

- Running dex process fails as well with the message on the jar atrium-specs-0.12.0-SNAPSHOT.jar 
`
Class or interface java.lang.UnsupportedOperationException required for desugaring of try-with-resources is not found.
`

What do you think?

build.gradle
```diff
@@ -136,10 +137,12 @@ kotlinutils.kotlinVersion = kotlin_version
 repositories {
     mavenCentral()
     jcenter()
+    maven { url "https://maven.google.com/" }
 }
 
 configurations {
     dependabot
+    r8
```
robstoll, line 145: :+1:

build.gradle
```diff
@@ -535,6 +541,24 @@ configure(subprojects - getJsProjects()) { subproject ->
     }
 }
 
+def checkDexer = task('checkDexer', description: 'compiles android class files into dex bytecode', dependsOn: compileKotlin) {
+    getAndroidProjects().each { p->
+        doLast {
+            javaexec {
+                def nameWithoutSuffix = p.name.substring(p.name.indexOf('-') + 1, p.name.lastIndexOf('-'))
+                def jvmProject = prefixedProject(nameWithoutSuffix + '-jvm')
+                def jarToTest = "${jvmProject.projectDir}\\build\\libs\\*.jar"
```
robstoll, line 550: what is the reason that you test the jvm jar? We don't guarantee that the jvm jar is dex compatible but  the android jar should be.

build.gradle
```diff
@@ -535,6 +541,24 @@ configure(subprojects - getJsProjects()) { subproject ->
...
+                mkdir "d8out/atrium-${nameWithoutSuffix}-${rootProject.version}"
```
robstoll, line 551: can you please write it to `${p.buildDir}/kotlin/d8`

build.gradle
```diff
@@ -535,6 +541,24 @@ configure(subprojects - getJsProjects()) { subproject ->
...
+                classpath = configurations.r8
+                main = "com.android.tools.r8.D8"
+                args = ['--output' , "d8out/atrium-${nameWithoutSuffix}-${rootProject.version}", "${jarToTest}"]
+            }
+        }
+    }
+}
+
+// check.dependsOn(checkDexer)
```
robstoll, line 560: don't forget to activate that at the end

build.gradle
```diff
@@ -535,6 +541,24 @@ configure(subprojects - getJsProjects()) { subproject ->
     }
 }
 
+def checkDexer = task('checkDexer', description: 'compiles android class files into dex bytecode', dependsOn: compileKotlin) {
+    getAndroidProjects().each { p->
```
robstoll, line 545: That's a nice idea but in the end I would like to have not only a task on the root of the project but individual tasks for those projects, so that I can do: `gradlew :atrium-core-api-android:checkDexer`

Hence the advice to do:
```
configure(getAndroidProjects()) { p ->
...
}
```

build.gradle
```diff
@@ -535,6 +541,24 @@ configure(subprojects - getJsProjects()) { subproject ->
     }
 }
 
+def checkDexer = task('checkDexer', description: 'compiles android class files into dex bytecode', dependsOn: compileKotlin) {
```
robstoll, line 544: as you test on the jar (which makes more sense than on the class files) you should change the dependsOn accordingly:
```suggestion
def checkDexer = task('checkDexer', description: 'compiles android class files into dex bytecode', dependsOn: jar) {
```

build.gradle
robstoll: Also, please use forward slashes

build.gradle
```diff
@@ -535,6 +541,24 @@ configure(subprojects - getJsProjects()) { subproject ->
...
+    getAndroidProjects().each { p->
+        doLast {
+            javaexec {
+                def nameWithoutSuffix = p.name.substring(p.name.indexOf('-') + 1, p.name.lastIndexOf('-'))
+                def jvmProject = prefixedProject(nameWithoutSuffix + '-jvm')
+                def jarToTest = "${jvmProject.projectDir}\\build\\libs\\*.jar"
+                mkdir "d8out/atrium-${nameWithoutSuffix}-${rootProject.version}"
+                classpath = configurations.r8
```
robstoll, line 552: I guess you need more than just the dependency to r8, you probably need also all jars which the corresponding project depends on

build.gradle
```diff
@@ -535,6 +541,24 @@ configure(subprojects - getJsProjects()) { subproject ->
...
+                main = "com.android.tools.r8.D8"
+                args = ['--output' , "d8out/atrium-${nameWithoutSuffix}-${rootProject.version}", "${jarToTest}"]
```
robstoll, line 554:
```suggestion
                args = ['--output' , "d8out/atrium-${nameWithoutSuffix}-${rootProject.version}", p.jar.archiveFile.get().asFile]
```
maybe better but requires that you wrap the whole thing into an `afterEvaluate {..}` block as the archiveFile does not include the version otherwise

robstoll:
So after I have made all comments above I quickly tried out if the comments make sense. So I came up with the following:
```gradle
configure(getAndroidProjects()) { p ->
    afterEvaluate {
        task('checkDexer', type: JavaExec, description: 'compiles android class files into dex bytecode', dependsOn: jar) {
            doFirst {
                mkdir "${p.buildDir}/kotlin/d8"
            }
            
            classpath = rootProject.configurations.r8
            main = "com.android.tools.r8.D8"
            args = ['--output', "${p.buildDir}/kotlin/d8", p.jar.archiveFile.get().asFile]
        }
    }
}
```

And I saw that my guess was correct, the dexer outputs a lot of warnings (unfortunately not errors, maybe there is a flag to turn warnings into errors?) that it cannot find dependent classes. For instance:
```
 Type `ch.tutteli.atrium.domain.creating.ChronoLocalDateAssertionsKt` was not found, it is required for default or static interface methods desugaring of `ch.tutteli.atrium.creating.Expect ch.tutteli.atrium.api.fluent.en_GB.jdk8.ChronoLocalDateAssertionsKt.isBefore(ch.tutteli.atrium.creating.Expect, java.time.chrono.ChronoLocalDate)`
```

assaflei:
I will handle the remarks later today or tomorrow, in the meanwhile several comments:

> Running build for android projects fails with this error
> Execution failed for task ':atrium-cc-de_CH-robstoll-android:test'.
> No tests executed, most likely the discovery failed.
We need this error to be handled in order to have a build that prepares all jars for the dexer, what exactly is missing? Can you assist or maybe we can ignore it for the scope of this task and only have a working dex process

- Regarding jar files that fail, do we want to try and fix the reasons for the failure? Since dex fails, having it executed as part of the build currently will break the build so we could leave it aside and invoke it on demand

build.gradle
assaflei: I noticed this method on the build.gradle of android projects so I figured we use jvm projects jar
But now I saw that building with android switch create jars so we can use them

Problem is that the build fails for me

build.gradle
assaflei: currently it will fail the build since not all jars dex are successful

robstoll:
The dexer shall give us further assurance that we can deliver our jar in an android setup. This of course make only sense if it is included in CI and the build stays green. Hence we need to handle it. 

However, I don't know why you get the error that discovery failed, this does not happen to me.

 > Running dex process fails as well with the message on the jar atrium-specs-0.12.0-SNAPSHOT.jar

right, specs, this one we can actually exclude. I have update the snippet above.

assaflei:
Actually, executing your suggested code works well on my side without any errors
Trying now to execute full build now

assaflei:
```
Execution failed for task ':atrium-cc-de_CH-robstoll-android:test'.
> No tests executed, most likely the discovery failed.
```
Regarding this error, I checked and there is no src/test folder inside the relevant project so the message makes sense to me.

`check.dependsOn(checkDexer)`
The command is not necessary since the suggested snippet executes as part of the build

Is there a build that executes on the server as part of travis?
We can see if the execution fails only on my side or not

robstoll:
> Regarding this error, I checked and there is no src/test folder inside the relevant project so the message makes sense to me.

That's because android has the same sources like jvm (see srcAndResourcesFromJvmProject in /build.gradle) with the exception of module-info.java which is only defined for jvm.
So the error makes sense at first glance but should not occur because there are tests in the corresponding jvm module. Have you tried to do a `gradlew clean build` ? Does that change something?

>> check.dependsOn(checkDexer)
>
> The command is not necessary since the suggested snippet executes as part of the build

That's right, because I included it for the subprojecs: `subproject.check.dependsOn(checkDexer)`

> Is there a build that executes on the server as part of travis?
We can see if the execution fails only on my side or not

There are multiple builds (see https://github.com/robstoll/atrium/pull/487/checks), some are executed on travis where linux is configured and some are executed on github actions where windows is configured. We can see for instance here:
https://travis-ci.com/github/robstoll/atrium/jobs/327529278#L591

That it is executed. However, with lot of warnings. We would like to:
- turn those into errors if possible (would be nice if you could to some research if d8 supports this)
- fix those warnings and for this I guess you need to do configure r8 so that the corresponding dependencies are also on the classpath. I think you can do this somehow with something along the line of 
   ```
   classpath subproject.configuratinos.implementation.asPath
   ```
   I have already similar things in atrium-bc-test/build.gradle maybe you find hints there

robstoll:
regarding classpath, maybe it would be easier if you move the 
```
configurations {
        r8
    }
```
including `repositories` and `dependencies` into the `configure(dexerProjects) { subproject ->` and add a dependency on `subproject`

robstoll:
Also note that I did not have a look at what `d8` expects, maybe (you need to check that) you need to adjust what you pass as argument to d8. Maybe there is a classpath flag or it is not enough to pass only the jar but requires all dependencies :man_shrugging:

robstoll:
@assaflei I had a quick look at the `--help` command of d8, there is a `--classpath` flag which you can use to pass in the dependencies of the subproject.

build.gradle
```diff
@@ -535,6 +541,59 @@ configure(subprojects - getJsProjects()) { subproject ->
...
+def dexerProjects = getAndroidProjects().findAll {
+    !it.name.contains("-specs-")
+}
+
+configure(dexerProjects) { subproject ->
+    afterEvaluate {
+        def checkDexer = task('checkDexer', type: JavaExec, description: 'compiles android project jar file into dex bytecode', dependsOn: jar) {
+            doFirst {
+                mkdir "${subproject.buildDir}/kotlin/d8"
+            }
```
robstoll, line 553: it just came to my mind that gradle supports this via outputs
```
outputs.dir("${subproject.buildDir}/kotlin/d8")
```

assaflei:
I had some problem when having to use rt.jar from jre
I didn't find it in any repository to include so I assume it exists as part of local jre installation
Is this assumption ok?

build.gradle
assaflei: will try updating

build.gradle
```diff
@@ -552,9 +552,27 @@ configure(dexerProjects) { subproject ->
...
+            subproject.configurations.each { conf ->
+                if (conf.name.endsWith("implementation")) {
+                    def copiedConfiguration = conf.copyRecursive();
+                    copiedConfiguration.setCanBeResolved(true);
+                    copiedConfiguration.asPath.split(';').each { path ->
+                        classPaths += ['--classpath', path]
+                    }
+                }
+            }
+            // classPaths += ['--classpath', "c:/Program Files (x86)/Java/jre1.8.0_241/lib/rt.jar"]
```
robstoll, line 568: I think there is a better solution, I remember having the same issue with atrium-bc-tests and the solution was to define an own configuration. You could do something like:
```
configure(dexerProjects) { subproject ->
    configurations {
        d8Classpath
    }
    dependencies {
        d8Classpath subproject
    }
```
and then somehting along the line of
```
def classPaths = subproject.configurations.d8Classpath.files.collect {
    ["--classpath", it.getAbsolutePath()]
}.flatten()
```

build.gradle
```diff
@@ -552,9 +552,27 @@ configure(dexerProjects) { subproject ->
...
+
+            //rt.jar is required as classpath for java.lang.* classes
+            classPaths += ['--classpath', "${System.getProperty('java.home')}/lib/rt.jar"]
```
robstoll, line 571: I see, lib/rt.jar is only available for < jdk9, i.e. for jdk8 in our case. We need to guard against that otherwise we will still see the warnings in >= jdk9 and I intend to fail the task if there are warnings.

robstoll:
@assaflei bad news, I read a bit about d8 and it seems it produces code which is not runnable under some circumstances in android in the end. Following an excerpt of https://jakewharton.com/androids-java-8-support/

```
If your device runs API 26 or newer you will see a timestamp and the string “Hello!” as expected. But running it on a device with a version earlier than API 26 produces a very different result.

java.lang.NoClassDefFoundError: Failed resolution of: Ljava/time/LocalDateTime;
  at Java8.sayHi(Java8.java:13)
  at Java8.main(Java8.java:9)
```

Which sadly means what we do here will not give us any guarantee at all :disappointed: 
We need to do it in another way. The first thing which comes to my mind is an actual android project where we can define an SDK API Level we want to support. 
But maybe there is another command line tool which checks for compatibility with a certain API level?  I could not find one with a quick search but maybe you can try to research this one.

assaflei:
> regarding classpath, maybe it would be easier if you move the
> 
> ```
> configurations {
>         r8
>     }
> ```
> 
> including `repositories` and `dependencies` into the `configure(dexerProjects) { subproject ->` and add a dependency on `subproject`

done

assaflei:
> @assaflei bad news, I read a bit about d8 and it seems it produces code which is not runnable under some circumstances in android in the end. Following an excerpt of https://jakewharton.com/androids-java-8-support/
> 
> ```
> If your device runs API 26 or newer you will see a timestamp and the string “Hello!” as expected. But running it on a device with a version earlier than API 26 produces a very different result.
> 
> java.lang.NoClassDefFoundError: Failed resolution of: Ljava/time/LocalDateTime;
>   at Java8.sayHi(Java8.java:13)
>   at Java8.main(Java8.java:9)
> ```
> 
> Which sadly means what we do here will not give us any guarantee at all 😞
> We need to do it in another way. The first thing which comes to my mind is an actual android project where we can define an SDK API Level we want to support.
> But maybe there is another command line tool which checks for compatibility with a certain API level? I could not find one with a quick search but maybe you can try to research this one.

Indeed bad news... at least the learning was fun...
So I'll stop my progress for now

I'll google a bit and will update
If you find something before, update as well...

robstoll:
I asked in the android channel @ kotlinlang.slack.com and people could not come up with something different than using an example project. But the good thing is, they gave me an idea how we could get to a result nonetheless.

We can download the `android.jar` of a certain API level we want to support for Atrium. Then we use it instead of the rt.jar of the jdk. This way we get at least warnings if types are not yet supported at that level. I tried that one out and I think it works:
running d8 with android.jar of level 23 gives the following warning for `:atrium-core-api-android:checkDexer`

```
Warning in atrium/core/api/atrium-core-api-jvm/build/libs/atrium-core-api-0.12.0-SNAPSHOT.jar:ch/tutteli/atrium/core/polyfills/LoadServiceKt$useSingleService$1$1.class:
  Type `java.util.function.Consumer` was not found, it is required for default or static interface methods desugaring of `ch.tutteli.atrium.core.polyfills.LoadServiceKt$useSingleService$1$1`
```

running it with level 24 does not as Optional as well as Consumer were both introduce at level 24 (https://developer.android.com/reference/java/util/Optional, https://developer.android.com/reference/java/util/function/Consumer).

So we can progress here :smiley: 

In order that you don't have to do the same again, following my current script:
```
def dexerProjects = getAndroidProjects().findAll {
    !it.name.contains("-specs-")
}
configure(dexerProjects) { subproject ->
    configurations {
        r8
        d8
    }
    repositories{
        maven { url "https://maven.google.com/" }
    }
    dependencies {
        r8 "com.android.tools:r8:$d8_version"
        d8 subproject
    }

    afterEvaluate {
        def checkDexer = task('checkDexer', type: JavaExec, description: 'compiles android class files into dex bytecode', dependsOn: jar) {
            outputs.dir("${subproject.buildDir}/kotlin/d8")

            classpath subproject.configurations.r8.asPath
            main = "com.android.tools.r8.D8"

            def d8Classpath = subproject.configurations.d8.files.collect {
                ["--classpath", it.getAbsolutePath()]
            }.flatten()

            args = d8Classpath + [
                '--lib', '/home/robert/projects/android-sdk-linux/platforms/android-23/android.jar',
                '--output', "${subproject.buildDir}/kotlin/d8",
                jar.archiveFile.get().asFile
            ]
        }
        subproject.check.dependsOn(checkDexer)
    }
}
task('checkDexer',  description: 'executes all checkDexer tasks') {

    doFirst {
        if (dexerProjects.size() == 0) {
            throw new GradleException("env variable CI not set and hence no android projects included. Use e.g. `CI=true gradlew checkDexer`")
        }
    }

    dexerProjects.forEach { subproject ->
        subproject.afterEvaluate {
            dependsOn subproject.tasks.getByName("checkDexer")
        }
    }
}
```
I guess the easiest thing is if we include the jar we want to support in the repo as it is quite a pain to download it. Or what do you think?

assaflei:
> Thus the following things need to be done:
> 
> use the code above
> adjust java-windows.yml include a new step check dexer where you:
> download the android.jar (you can download it e.g. from here https://github.com/Sable/android-platforms) check that the sha256sum is correct
> run checkDexer
> somehow check that checkDexer does not output warnings (otherwise fail the build)

So I should download the jar file as part of the new step inside the yml? 
And set an env var as well "ATRIUM_ANDROID_JAR"?

.github/workflows/java-windows.yml
```diff
@@ -16,10 +16,17 @@ jobs:
                 uses: actions/setup-java@v1
                 with:
                     java-version: ${{ matrix.java_version }}
+            -   name: check dexer
+                run: Invoke-WebRequest -Uri "https://github.com/Sable/android-platforms/blob/master/android-28/android.jar?raw=true" -OutFile "./android.jar"
```
robstoll, line 20: Let's try out if we can support API level 24 first, I think it should be enough for now. We can still increase to 26 in case we want to include java.nio.fie.Files

Btw. I liked the approach with wget, maybe curl works?
Don't forget to include a sha265sum check (sha256 for android-24 is `4c438bf1b3d4e6104faccc62e1ffe95ef197b88055497b540a871349e93c8685`
)

.github/workflows/java-windows.yml
```diff
@@ -16,10 +16,17 @@ jobs:
...
+                run: ./gradlew checkDexer
```
robstoll, line 21: seems like you cannot have multiple `run` in one step.

assaflei: yes I got to this error locally
do you know why is the travis queued for so long?

robstoll: it actually already failed: https://travis-ci.com/github/robstoll/atrium/builds/164298158
I guess the github did not get the notification

.github/workflows/java-windows.yml
robstoll: Nah... let's increase to 26 already. I intend to get rid of the jdk8 extension in the long run and we have java.nio.file.Files in there.

build.gradle
```diff
@@ -535,6 +537,76 @@ configure(subprojects - getJsProjects()) { subproject ->
...
+                ["--classpath", it.getAbsolutePath()]
+            }.flatten()
+
+            args = d8Classpath + [
+                // ATRIUM_ANDROID_JAR has to point to something like android-sdk-linux/platforms/android-XY/android.jar where Xy is the version we want to support
+                '--lib', System.getenv('ATRIUM_ANDROID_JAR'),
+                '--output', "${subproject.buildDir}/kotlin/d8",
+                jar.archiveFile.get().asFile
+            ]
+        }
```
robstoll, line 573: you need to include a check here that we did not have warnings in the output. I think you could do it with setting
```
standardOutput = yourOwnOutputStream
errorOutput = yourOwnOutputStream
```
and then perform a check in `doLast`

build.gradle
```diff
@@ -535,6 +537,62 @@ configure(subprojects - getJsProjects()) { subproject ->
...
+            }.flatten()
+
+            def androidJarPath = System.getenv('ATRIUM_ANDROID_JAR')
+
+            // ATRIUM_ANDROID_JAR has to point to something like android-sdk-linux/platforms/android-XY/android.jar
+            // where Xy is the version we want to support
+            def androidJar = androidJarPath != null ? ['--lib', androidJarPath] : []
+
+            args = d8Classpath + androidJar + [
+                // ATRIUM_ANDROID_JAR has to point to something like android-sdk-linux/platforms/android-XY/android.jar where Xy is the version we want to support
```
robstoll, line 574: remove this line

assaflei:
To improve execution time I'm thinking about adding another task to use in CI:

```
def buildAndroidProjects = task('buildAndroidProjects', group: build, description: 'builds all android projects')
buildAndroidProjects.dependsOn(dexerProjects.collect { it.build })
```

robstoll:
You will not speed up execution time with this. The contrary is the case, as `build`  includes much more than what checkDexer needs (it needs only `jar` of the corresponding task).
I guess you came up with this due to the build error shown here: https://github.com/robstoll/atrium/runs/651057605#step:5:384

We should fix/workaround the problem there as with the current solution we would run into the same problem locally when someone does ` gradlew clean checkDexer`. I guess the problem is, that ` atrium-domain-robstoll is defined as runtime dependency. Which means we need to adjust `subproject.configurations.d8.files.collect` I think there is a way to get the runtimeClasspath. Have a look into bc-tests I am pretty sure I used it there

assaflei:
I saw that the build succeed after I ran the build of the android projects step before check dexer step

We should do it differently although?

robstoll:
Yes, run clean checkDexer locally and you see why I guess

robstoll:
Basically a gradle task should work on its own without the need to execute another task beforehand. This should be configured with the dependsOn.

build.gradle
```diff
@@ -558,7 +558,7 @@ configure(dexerProjects) { subproject ->
     }
 
     afterEvaluate {
```
robstoll, line 560: As said depending on jar should be enough we don't want to execute tests for instance. You need to fix the classpath we pass to dexer. Try `gr checkDexer --dry-run` I guess everything is already correct

build.gradle
```diff
@@ -558,7 +558,7 @@ configure(dexerProjects) { subproject ->
...
-        def checkDexer = task('checkDexer', type: JavaExec, description: 'compiles android class files into dex bytecode', dependsOn: jar) {
+        def checkDexer = task('checkDexer', type: JavaExec, description: 'compiles android class files into dex bytecode', dependsOn: buildAndroidProjects) {
```
robstoll, line 561: As said depending on jar should be enough we don't want to execute tests for instance. You need to fix the classpath we pass to dexer. Try gr checkDexer --dry-run I guess everything is already correct

robstoll: Ah no, it must be the other way round, we pass runtime dependencies to dexer but their dependencies aren't build yet. Could be a gradle bug, we already had to do workarounds for it. search for domain-robstollin build.gradle files. Maybe you find the workaround. In any case, perform the dry run and check if atrium-domain-robstoll-android:jar comes earlier than checkDexer (I assume currently it is executed later)

assaflei:
```
configure(dexerProjects) {
    compileKotlin.dependsOn(
        prefixedProject('domain-robstoll-android').jar
    )
}
```

Is this the type of change you were thinking of?
Its not working for me but I didnt notice any other special usage of domain-robstoll due to circular dependency

robstoll: Yeah something like that, but I would put it into the project's build.gradle which causes the problem. But no worries, you can also keep it like it is now and I improve it once your PR is merged. For me it's more valuable if you can implement the check that d8 did not output warnings and fail otherwise

assaflei: Added the output verification
Since successful build results in no messages at all, in any case of error message I will fail the build

robstoll: 👍

assaflei:
I think I managed to build transient and direct dependencies in the latest commit

assaflei:
I think that this task is complete, @robstoll what do you think?

.github/workflows/java-windows.yml
```diff
@@ -45,3 +52,5 @@ jobs:
 
             -   name: composite build atrium-samples
                 run: misc\tools\atrium-samples-test\gradlew -p misc\tools\atrium-samples-test build
+                env:
+                    ATRIUM_ANDROID_JAR: ./android.jar
```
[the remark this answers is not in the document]
robstoll: Not sure why you have marked this as resolved as it is still here. Did you encounter problems?

robstoll:
@assaflei I think so, I'll just make a few adjustments before merging into master. thanks for your work :+1:

robstoll:
@assaflei this one is the next step, now that we could show that the module-info.class bug is no longer a problem and we are compatible with Android SDK API Level 26: https://github.com/robstoll/atrium/issues/493

I could again use your help :slightly_smiling_face: 

And, in case you prefer to stick to more dev-ops related tasks: https://github.com/robstoll/atrium/issues/495

Issue #470, opened 2020-04-24
create maven sample project


robstoll:
*Platform* (jvm, js, android): jvm
*Extension* (none, kotlin 1.3, jdk8): none

## Code related feature
We currently provide an example-pom.xml and a settings.xml for maven.
It would be nicer if we actually have a full-working sample for maven.

Following the things you need to do:
- [x] setup a maven project in /samples/maven and include some tests -- up to you what runner you are using.
- [ ] build the sample in CI, add corresponding steps/jobs in .travis.yml and .github/workflows/java-windows.yml
- :no_entry_sign:  include the new sample project in the composite build of misc/tools/atrium-samples-test => won't be possible as it is not a gradle project. That's fine though.
- [x] link to the sample project in the README -> installation -> jvm (https://github.com/robstoll/atrium#jvm) -> maven

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime, we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

robstoll:
@binkley I saw that you are using Atrium in a maven context (https://github.com/binkley/spikes/blob/master/scratch/pom.xml). Would you mind creating a sample project so that others have an easier start when using Atrium as well?

binkley:
@robstoll I'd love to!  <3

What model should I follow?  Any specific kinds of tests, etc.?

robstoll:
nice :+1: the idea is to have a minimalistic setup of a maven project using Atrium and some test-runner (e.g. junit + jupiter) including a SampleTest. So that someone wanting to start out a new project and would like to use Atrium can basically copy the sample project.
You can come up with a SampleTest or copy an existing one (e.g. samples/jvm/junit5/src/test/kotlin/org/atriumlib/samples/junit5/SampleJvmTest.kt)

binkley:
The sample looks good.  The only change I'd like to make: I've run into the Kotlin convention of providing "text" test method names, ie:

```
fun `expect an exception occurred`() { ... }
```
[...]

PR #488, opened 2020-04-30
Sample Maven and wrapper setup for JVM JUnit5

binkley:
Identical to Gradle build with 1 needed change:

The PR updates the Maven JUnit to 5.6.1; the Gradle build retains use of 5.3.1.

A key improvement with the 5.6.1 is better/easier integration with Maven, so less `pom.xml` setup is needed.

This PR addresses https://github.com/robstoll/atrium/issues/470


samples/jvm/junit5/.gitignore
```diff
@@ -1,10 +1,11 @@
 # Ignore Gradle project-specific cache directory
-.gradle
+/.gradle
```
robstoll, line 2: out of curiosity: unless I am wrong, this will only ignore .gradle in the root, why do you want to commit it if it is not in the root?

binkley: Ah, I see.  I'll back this out.  I was working from this dir as-if it were project root

robstoll:
Thanks for the update, two points to go (see https://github.com/robstoll/atrium/pull/488#pullrequestreview-403462956) and it's good to merge

robstoll:
@binkley do you plan to do the last to points as well? Let me know in case you need help with them.

robstoll:
@binkley I take over from here, many thanks for your work :slightly_smiling_face: :+1:

PR #492, opened 2020-05-04
#430 [WIP] deprecate iterableassertions file

Miftahunajat:
This PR add Deprecate api-cc-infix and add Replacewith for the expect for these files:
- [x] iterableAssertions.kt
- [x] iterableContainsCheckers.kt
- [x] iterableContainsInAnyOrderCreators.kt
- [x] iterableContainsInOrderOnlyCreators.kt
- [x] iterableContainsInAnyOrderOnlyCreators.kt
- [ ] iterableContainsInOrderOnlyGroupedCreators.kt
- [ ] iterableContainsSearchBehaviours.kt

### DIFF
--- a/apis/cc-infix-en_GB/atrium-api-cc-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/cc/infix/en_GB/iterableAssertions.kt
+++ b/apis/cc-infix-en_GB/atrium-api-cc-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/cc/infix/en_GB/iterableAssertions.kt
@@ -26,6 +26,16 @@ import kotlin.jvm.JvmName
  *
  * @return The newly created builder.
  */
+@Deprecated(
+    "Switch from Assert to Expect; will be removed with 1.0.0 -- see https://github.com/robstoll/atrium/releases/tag/v0.9.0#migration for migration hints and scripts.",
+    ReplaceWith(
+        "this.asExpect().contains(ch.tutteli.atrium.api.infix.en_GB.o)",
+        "ch.tutteli.atrium.domain.builders.migration.asExpect",
+        "ch.tutteli.atrium.domain.builders.migration.asAssert",
+        "ch.tutteli.atrium.api.infix.en_GB.contains",
+        "ch.tutteli.atrium.api.infix.en_GB.o"
+    )
+)
 infix fun <E, T : Iterable<E>> Assert<T>.to(@Suppress("UNUSED_PARAMETER") contain: contain): IterableContains.Builder<E, T, NoOpSearchBehaviour>
     = AssertImpl.iterable.containsBuilder(this)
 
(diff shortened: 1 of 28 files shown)


Miftahunajat:
I am checking with one file first to make sure the build and the test run successfully

apis/cc-infix-en_GB/atrium-api-cc-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/cc/infix/en_GB/iterableAssertions.kt
```diff
@@ -85,6 +104,15 @@ infix fun <E, T : Iterable<E>> Assert<T>.contains(values: Values<E>): AssertionP
  * @return This plant to support a fluent API.
  * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
  */
+@Deprecated(
+    "Switch from Assert to Expect; will be removed with 1.0.0 -- see https://github.com/robstoll/atrium/releases/tag/v0.9.0#migration for migration hints and scripts.",
+    ReplaceWith(
+        "this.asExpect().contains(assertionCreatorOrNull).asAssert()",
```
robstoll, line 110: this won't work as assertionCreatorOrNull is of type `(Assert<E>.() -> Unit)?` where the new infix requires `(Expect<E>.() -> Unit)?` => see cc-en_GB/atrium-api-cc-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/cc/en_GB/iterableAssertions.kt

apis/cc-infix-en_GB/atrium-api-cc-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/cc/infix/en_GB/iterableAssertions.kt
```diff
@@ -46,6 +46,15 @@ infix fun <E, T : Iterable<E>> Assert<T>.notTo(@Suppress("UNUSED_PARAMETER") con
  * @return This plant to support a fluent API.
  * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
  */
+@Deprecated(
+    "Switch from Assert to Expect; will be removed with 1.0.0 -- see https://github.com/robstoll/atrium/releases/tag/v0.9.0#migration for migration hints and scripts.",
```
robstoll, line 50: don't forget `to contain` and `notTo contain` above

Miftahunajat: I'm not sure about the replacement of to contain and notTo contain.
Are there any examples?

robstoll: you can always find the correspondence in the new infix API. The structure of the files should be similar. i.e. have a lookt at iterableAssertions.kt in api-infix

robstoll:
just that you don't search for the reason why the build fails. You need to do the replacements in addition

apis/cc-infix-en_GB/atrium-api-cc-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/cc/infix/en_GB/iterableAssertions.kt
Miftahunajat: Hmm. 
Why I can't find method `to` in `ch.tutteli.atrium.api.cc.infix.en_GB.to` for replacement

robstoll: It's again the same as for CharSequence, it was renamed and is called `contains(o)` instead of `to contain`.
I guess I was not clear enough what I meant with `same structure`. I try to explain it better. If you deprecate `to contain` in api-cc-infix which is the first thing within iterableAssertions.kt then most likely the same functionality is also the first thing you will see in api-infix. I hope you get what I mean

robstoll:
@Miftahunajat you need to suppress deprecations in AtLeastCheckerOptionImpl.kt etc.

robstoll:
I assume you have seen why the build fails (the error message is different). Ping me in case you need help

Miftahunajat:
Hello @robstoll sorry for my late response.
I already updated the deprecation of some file.
But it looks like I can't continue for this issue for a while.
Sorry for inconvenience

robstoll:
No worries, in this case I might finish it and in that case merge it into another branch.
Already thanks for the work you did.

Issue #495, opened 2020-05-08
cache gradle and dependencies in github actions


robstoll:
*Platform* (jvm, js, android): all
*Extension* (none, kotlin 1.3, jdk8): all

## Non-Code related feature
We currently have no cache in place for github actions. I guess we could reduce the build time quite a bit if we integrate it.

Things TODO:
- :no_entry_sign: cache gradle and dependencies: https://github.com/actions/cache/blob/master/examples.md#java---gradle => due to a Kotlin bug (see https://youtrack.jetbrains.com/issue/KT-39109)
- [x] cache also maven dependencies (for maven sample project): https://github.com/actions/cache/blob/master/examples.md#java---maven
- [x] cache android.jar which we download for d8: https://github.com/actions/cache#skipping-steps-based-on-cache-hit - skip downloading in case of a cache-hit

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime, we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

assaflei:
I'll take this

PR #497, opened 2020-05-12
merge jdk8 extensions to jvm

assaflei:
Started working with the api-fluent classes
module-info.java file can be deleted, right?


README.md
```diff
@@ -107,13 +107,12 @@ which provides a pure fluent API (in en_GB) for the JVM platform.
...
 
 We currently provide the following extensions for the JVM platform: 
-- jdk8: assertion functions for JDK 8 specific types (e.g. for `Path`)
 - kotlin_1_3: assertion functions for Kotlin 1.3 specific types (e.g. for [Result](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/index.html)). 
 
 You can enable them as follows:
 ```
 dependencies {
-    testImplementation "ch.tutteli.atrium:atrium-api-fluent-en_GB-jdk8:$atrium_version"
+    testImplementation "ch.tutteli.atrium:atrium-api-fluent-en_GB:$atrium_version"
```
robstoll, line 115: please delete this line as well

README.md
```diff
@@ -140,7 +139,7 @@ dependencies {
 And for the aforementioned extensions:
 ```
 dependencies {
-    testImplementation "ch.tutteli.atrium:atrium-api-infix-en_GB-jdk8:$atrium_version"
+    testImplementation "ch.tutteli.atrium:atrium-api-infix-en_GB:$atrium_version"
```
robstoll, line 142: same here

README.md
```diff
@@ -1356,9 +1355,9 @@ expect(Paths.get("/usr/bin/noprogram")).exists()
 ```
 ↑ <sub>[Example](https://github.com/robstoll/atrium/tree/master/samples/readme-examples/src/main/kotlin/readme/examples/ReadmeSpec.kt#L326)</sub> ↓ <sub>Output</sub>
 ```text
-expected that subject: /usr/bin/noprogram        (sun.nio.fs.UnixPath <1234789>)
+expected that subject: \usr\bin\noprogram        (sun.nio.fs.WindowsPath <1234789>)
 ◆ to: exist
-    » the closest existing parent directory is /usr/bin
+    » the closest existing parent directory is none
```
robstoll, line 1360: please revert this and the other remaining changes further below in this file

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/chronoLocalDateAssertions.kt
```diff
@@ -1,6 +1,6 @@
 @file:Suppress("JAVA_MODULE_DOES_NOT_READ_UNNAMED_MODULE" /* TODO remove once https://youtrack.jetbrains.com/issue/KT-35343 is fixed */)
 
-package ch.tutteli.atrium.api.fluent.en_GB.jdk8
+package ch.tutteli.atrium.api.fluent.en_GB
```
robstoll, line 3: don't change the package here, change it in the copied files (which are in the normal jvm module)

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/module/module-info.java
```diff
@@ -1,10 +1,10 @@
-module ch.tutteli.atrium.api.fluent.en_GB.jdk8 {
-    requires ch.tutteli.atrium.api.fluent.en_GB;
-    requires ch.tutteli.atrium.domain.builders;
-    requires ch.tutteli.kbox;
-    requires ch.tutteli.niok;
-    requires kotlin.stdlib;
-    requires java.base;
+// module ch.tutteli.atrium.api.fluent.en_GB.jdk8 {
```
robstoll, line 1: keep, you basically don't need to modify the jdk8 extension, only deprecate the methods. We want to keep the extension to stay backward compatible. We will remove everything with 1.0.0. The deprecation messages are i.a. here to help users migrate from the jdk8 extension to the normal jvm module

bundles/fluent-en_GB/extensions/atrium-fluent-en_GB-smoke-test-jdk8/build.gradle
```diff
@@ -6,5 +6,5 @@ targetCompatibility = JavaVersion.current()
 dependencies {
     //I don't see how to set up compileTestKotlin with --patch-module, so we have put the module-info.java directly in src/test/kotlin instead
     testImplementation prefixedProject('fluent-en_GB-jvm')
-    testImplementation prefixedProject('api-fluent-en_GB-jdk8-jvm')
+    // testImplementation prefixedProject('api-fluent-en_GB-jdk8-jvm')
```
robstoll, line 9: I suggest we remove atrium-fluent-en_GB-smoke-test-jdk8 entirely

robstoll:
@assaflei I have crossed the points you have already done in the issue. As you can see, there was a mistunderstanding for the first 3 points

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/chronoLocalDateAssertions.kt
assaflei: this is the file in the jvm module folder, not change the package name here?

bundles/fluent-en_GB/extensions/atrium-fluent-en_GB-smoke-test-jdk8/build.gradle
assaflei: delete the whole folder?

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/module/module-info.java
assaflei: the build fails with an error:
```
package is empty or does not exist: ch.tutteli.atrium.api.fluent.en_GB.jdk8        
    exports ch.tutteli.atrium.api.fluent.en_GB.jdk8
```
Should it remain commented?

assaflei:
> there was a mistunderstanding for the first 3 points
I didnt understand the mistunderstanding, can you explain?

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/chronoLocalDateAssertions.kt
robstoll: ah... now I see, you moved it instead of copying it. So here it would be correct to adjust the package name but in this file we don't want to see the deprecations.

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/module/module-info.java
robstoll: will change once you copy the file from the jdk8 extension to the normal jvm module instead of moving.

bundles/fluent-en_GB/extensions/atrium-fluent-en_GB-smoke-test-jdk8/build.gradle
robstoll: I would delete the whole folder. Check in settings.gradle in the root that it is excluded as well.
What would be nice, move:
```
    test("see if `Path.existsNot` can be used") {
        assert(Paths.get("nonExisting")).existsNot()
    }
``` 
from bundles/fluent-en_GB/extensions/atrium-fluent-en_GB-smoke-test-jdk8/src/test/kotlin/custom/SmokeSpec.kt to bundles/fluent-en_GB/atrium-fluent-en_GB-jvm/src/test/kotlin/custom/SmokeSpec.kt

robstoll:
@assaflei see comments, I would fix those remarks first before moving on with the infix API

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/chronoLocalDateAssertions.kt
assaflei: @robstoll ?

robstoll: You need to two two separate things:
1. you need to copy (not move) every file from the jdk8 module to the normal jvm module
2. deprecate methods in the jdk8 extension, taking this file as example, every method in in apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/chronoLocalDateAssertions.kt => here you don't have to adjust the package name


It seems like you first deprecated methods in the jdk8 module and then moved/instead of copying the file to the jvm module. 
Does that make things clearer?

assaflei:
Can you check my feedback to your comments?

robstoll:
Did and wrote back. Is it now clear what needs to be done?

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/module/module-info.java
assaflei: Maybe we can remove the whole folder apis\fluent-en_GB\extensions\jdk8\atrium-api-fluent-en_GB-jdk8-jvm
what do you think?

robstoll: no, we need it for backward compatibility

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/chronoLocalDateAssertions.kt
assaflei: got it, my mistake

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/module/module-info.java
assaflei: got it

assaflei:
I got what I did wrong,
committed updates, can you confirm I can proceed?

README.md
```diff
@@ -1375,9 +1372,8 @@ expected that subject: /root/.ssh/config        (sun.nio.fs.UnixPath <1234789>)
 ◆ equals: writable
     » failure at parent path: /root        (sun.nio.fs.UnixPath <1234789>)
       » access was denied
-      » the owner is root, the group is root
-      » the permissions are u=rwx g= o=
-```
+      » the owner is root, the group is root	
+      » the permissions are u=rwx g= o=```
```
robstoll, line 1376: please revert this

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/ChronoLocalDateAssertionsSpec.kt
```diff
@@ -1,4 +1,4 @@
-package ch.tutteli.atrium.api.fluent.en_GB.jdk8
+package ch.tutteli.atrium.api.fluent.en_GB
```
robstoll, line 1: here (and in all other files in the jdk8 extension) you should keep the package as is

assaflei:
I'm having a problem with the functions defined at apis\infix-en_GB\atrium-api-infix-en_GB-jvm\src\test\kotlin\ch\tutteli\atrium\api\infix\en_GB\PathAssertionsSpec.kt:
```
    fun1<Path, readable>(Expect<Path>::toBe).name to Companion::isReadable,
    fun1<Path, writable>(Expect<Path>::toBe).name to Companion::isWritable,
    fun1<Path, aRegularFile>(Expect<Path>::toBe).name to Companion::isRegularFile,
    fun1<Path, aDirectory>(Expect<Path>::toBe).name to Companion::isDirectory,

```
Compilation fails with this error
```
e: apis\infix-en_GB\atrium-api-infix-en_GB-jvm\src\test\kotlin\ch\tutteli\atrium\api\infix\en_GB\PathAssertionsSpec.kt: (23, 40): Cannot choose among the following candidates without completing type inference:
public infix fun <T : Path> Expect<Path>.toBe(writable: writable): Expect<Path> defined in ch.tutteli.atrium.api.infix.en_GB
public infix fun <T> Expect<Path>.toBe(expected: Path): Expect<Path> defined in ch.tutteli.atrium.api.infix.en_GB
```

Is there a conflict with the function toBe found at apis\infix-en_GB\atrium-api-infix-en_GB-common\src\main\kotlin\ch\tutteli\atrium\api\infix\en_GB\anyAssertions.kt ?

I tried using @JvmName attribute but the build still fails

Do you jave any ideas?

robstoll:
Let me know in case I shall have a look into. I assume you found a solution otherwise

assaflei:
So I thought I have found something but no...
I understand that this line for example
```  
fun1<Path, writable>(Expect<Path>::toBe).name to Companion::isWritable,
```
causes this error
```
e: apis\infix-en_GB\atrium-api-infix-en_GB-jvm\src\test\kotlin\ch\tutteli\atrium\api\infix\en_GB\PathAssertionsSpec.kt: (23, 40): Cannot choose among the following candidates without completing type inference:
public infix fun <T : Path> Expect<Path>.toBe(writable: writable): Expect<Path> defined in ch.tutteli.atrium.api.infix.en_GB
public infix fun <T> Expect<Path>.toBe(expected: Path): Expect<Path> defined in ch.tutteli.atrium.api.infix.en_GB
```

but I couldnt find out why and where the fix should be
tried JvmName but it doesnt work, I thought of playing with the syntax but I didnt find the correct way

@robstoll Would appreciate your assistance

robstoll:
the problem is that Kotlin's type inference system is not good enough and hence cannot figure out which overload it should take. However, I would rewrite the test in any case and use the approach as in apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/CollectionAssertionsSpec.kt

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/module/module-info.java
```diff
@@ -1,6 +1,7 @@
 module ch.tutteli.atrium.api.infix.en_GB {
     requires ch.tutteli.atrium.domain.builders;
     requires kotlin.stdlib;
+    requires java.base;
```
robstoll, line 4: also add
```
    requires ch.tutteli.kbox;
    requires ch.tutteli.niok;
```
unfortunately there is a bug in Kotlin which does not detect this mistake :(

assaflei:
I'll check and update

assaflei:
Well it seems like the suggestion worked, now I get a new error in common module which fails the build of jvm module
It appears only after I added the new classes to jvm and I dont see the relation,
Can you advice?

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/ChronoLocalDateAssertionsSpec.kt
robstoll: @assaflei already revert this (basically revert all files in apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8); 
Add the following to each file at the beginning instead:
```
//TODO remove file with 1.0.0
@file:Suppress("DEPRECATION")
```
I will look into the problem of the build tomorrow.

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/module/module-info.java
robstoll: also apply the same changes in apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/module/module-info.java

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/pathAssertions.kt
```diff
@@ -0,0 +1,297 @@
...
+ *
+ * This matcher _resolves_ symbolic links. Therefore, if a symbolic link exists at the location the subject points to,
+ * then the search will continue at that location.
+ *
+ * @return An [Expect] for the current subject of the assertion.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ *
+ * @since 0.12.0
+ */
+infix fun <T : Path> Expect<T>.to(@Suppress("UNUSED_PARAMETER") exist: exist): Expect<T> =
```
robstoll, line 68: the problem of the build is this function, I always figured we might run into problems and tada, now we do. 
The good thing, since it is a new function in this API we can rename it -- you need to adjust the ReplaceWith in the jdk8 extension accordingly though. I suggest we use:
```
toBe(existing)
```

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/pathAssertions.kt
```diff
@@ -0,0 +1,297 @@
...
+ * This assertion is not atomic with respect to concurrent file system operations on the paths the assertions works on.
+ * Its result, in particular its extended explanations, may be wrong if such concurrent file system operations
+ * take place.
+ *
+ * @return An [Expect] for the current subject of the assertion.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ *
+ * @since 0.12.0
+ */
+infix fun <T : Path> Expect<T>.toBe(@Suppress("UNUSED_PARAMETER") readable: readable): Expect<T> =
```
robstoll, line 213: Another word on ReplaceWith in the jdk8 extension of the infix API. You need to use the full qualified name for the pseudo keywords. For instance for this function you have to add:
```
ReplaceWith("ch.tutteli.atrium.api.infix.en_GB.toBe(ch.tutteli.atrium.api.infix.en_GB.readable)")
```
and you need to deprecated the pseudo-keywords as well pointing to the new pseudo-keyword (adjust name in case it changed)

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/pathAssertions.kt
```diff
@@ -0,0 +1,297 @@
...
+ *
+ * This matcher _resolves_ symbolic links. Therefore, if a symbolic link exists at the location the subject points to,
+ * then the search will continue at that location.
+ *
+ * @return An [Expect] for the current subject of the assertion.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ *
+ * @since 0.12.0
+ */
+infix fun <T : Path> Expect<T>.notTo(@Suppress("UNUSED_PARAMETER") exist: exist): Expect<T> =
```
robstoll, line 83: same here rename to 
```
notToBe(existing)
```

assaflei:
Is the build breaking because of the warnings?
```
e: warnings found and -Werror specified
```

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/pathAssertions.kt
```diff
@@ -65,7 +65,7 @@ infix fun <T : Path> Expect<T>.endsNotWith(expected: Path): Expect<T> =
  *
  * @since 0.12.0
  */
-infix fun <T : Path> Expect<T>.to(@Suppress("UNUSED_PARAMETER") exist: exist): Expect<T> =
+infix fun <T : Path> Expect<T>.toBe(@Suppress("UNUSED_PARAMETER") exist: exist): Expect<T> =
```
robstoll, line 68: please also adjust the pseudo-keyword from `exist` to `existing`

apis/infix-en_GB/extensions/jdk8/atrium-api-infix-en_GB-jdk8-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/jdk8/pathAssertions.kt
```diff
@@ -17,6 +17,7 @@ import java.nio.file.Path
  *
  * @since 0.12.0
  */
+@Deprecated("Use the function from the normal jvm module; the jdk8 extension will be removed with 1.0.0", ReplaceWith("ch.tutteli.atrium.api.infix.en_GB.startsWith"))
```
robstoll, line 20: sorry, I did not thought about it before, but we can actually just delete the whole jdk8 extension for the infix API. We would have introduced it with 0.12.0 (next version).

robstoll:
> Is the build breaking because of the warnings?
> 
> ```
> e: warnings found and -Werror specified
> ```

Most likely yes, that's why you need to add suppressions as shown here: https://github.com/robstoll/atrium/pull/497#discussion_r424715474

apis/fluent-en_GB/extensions/jdk8/atrium-api-fluent-en_GB-jdk8-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/jdk8/ChronoLocalDateAssertionsSpec.kt
```diff
@@ -1,70 +1,70 @@
...
-        a1 = a1.isEqual(chronoLocalDate)
-
-        a2 = a2.isBefore(chronoLocalDate)
-        a2 = a2.isBeforeOrEqual(chronoLocalDate)
-        a2 = a2.isAfter(chronoLocalDate)
-        a2 = a2.isAfterOrEqual(chronoLocalDate)
-        a2 = a2.isEqual(chronoLocalDate)
-    }
-}
+package ch.tutteli.atrium.api.fluent.en_GB
```
robstoll, line 1: please revert in all specs and instead add the deprecation
```suggestion
//TODO remove file with 1.0.0
@file:Suppress("DEPRECATION")
package ch.tutteli.atrium.api.fluent.en_GB.jdk8
```

assaflei:
I think I made a small mess here
I ran a rebase from head for 12 commits which were the amount of commits made during this PR and than force pushed it

Now should I execute a rebase master command?

robstoll:
do a squash of your commits and then rebase on master. Have a look on one of the older PRs you have created, we had the same situation there and I wrote down a little step-by-step guide.

assaflei:
Found it
https://github.com/robstoll/atrium/pull/440#issuecomment-619435648

Will try it

assaflei:
Since I made a squash the order of commits is no longer there
I managed to find the sha of first commit but I'm not sure its connected to all the others
Are we sure this method will work now?

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/build.gradle
```diff
@@ -16,6 +16,7 @@ dependencies {
         exclude module: "${rootProject.name}-translations-en_GB-jvm"
     }
     testImplementation prefixedProject('translations-de_CH-jvm')
+    api niok()
```
robstoll, line 19: do we need this?

assaflei: I'll try building without it and see if it works

robstoll:
It looks good to me but have a look yourself:  https://github.com/robstoll/atrium/pull/497/files

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/build.gradle
robstoll: if we do, then please move up to other api dependencies

assaflei:
I dont know how it got to be like that, I usually merge master to my branch to make sure it builds correctly before I push
This time something went lost...

I can create another PR and just add the changes to it
what do you say?

robstoll:
The commits you have made contain only the changes you have made so no need for another PR IMO but you can also go for it if it suits you.
I suggest you have a look at the git-approach explained here:
https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#git

which rebases instead of merging. This makes reviewing for me much easier as it never happens that suddenly changes which are already in master show up. 

Thanks for you last changes, there are two things open:
- https://github.com/robstoll/atrium/pull/497#discussion_r424170678
- please remove git submodules you have added under /act

assaflei:
So instead of merging every time, the practice is to rebase the branch using git rebase upstream/master?

robstoll:
I favour rebase over merge yes.

assaflei:
Nice, I'll try using this approach in the future

assaflei:
So I think this PR is done?

robstoll:
@assaflei yes, merged it :slightly_smiling_face: 
How about https://github.com/robstoll/atrium/issues/495 as next issue? Faster build time for github actions would be nice.

PR #500, opened 2020-05-16
#495 added cache steps to yml

assaflei:

### DIFF
--- a/.github/workflows/java-windows.yml
+++ b/.github/workflows/java-windows.yml
@@ -4,6 +4,8 @@ on: [push, pull_request]
 
 jobs:
     build:
+        env:
+            android-version: 26
         runs-on: windows-latest
         strategy:
             fail-fast: true
@@ -17,12 +19,22 @@ jobs:
                 with:
                     java-version: ${{ matrix.java_version }}
 
+            -   name: Cache android jar
+                id: android-cache
+                uses: actions/cache@v1
+                with:
+                    path: android-jar-cache
+                    key: ${{ runner.os }}-android-version-${{ env.android-version }}
+
             -   name: buildNonDeprecatedJvm
                 run: ./gradlew buildNonDeprecatedJvm
 
-            -   name: download Android SDK's android.jar
+            -   name: download Android SDK's android.jar (or use cached file)
+                if: steps.android-cache.outputs.cache-hit != 'true'
                 run: |
-                    curl -o android.jar -L "https://github.com/Sable/android-platforms/blob/master/android-26/android.jar?raw=true"
+                    md android-jar-cache
+                    cd android-jar-cache
+                    curl -o android.jar -L "https://github.com/Sable/android-platforms/blob/master/android-${{ env.android-version }}/android.jar?raw=true"
                     $file_sha256 = $(certutil -hashfile android.jar sha256)[1] -replace " ",""
                     if ($file_sha256 -Ne "cdc1846376a14b0370cc63454a129606b4a52cc50ada75ef0d4cf956b1ad2daa") {
                         echo "wrong sha256 for android.jar: $file_sha256";
@@ -31,7 +43,7 @@ jobs:
                     }
 
             -   name: check Atrium's -android.jar can be dexed
-                run: ATRIUM_ANDROID_JAR="$PWD/android.jar" ./gradlew checkDexer
+                run: ATRIUM_ANDROID_JAR="$PWD/android-jar-cache/android.jar" ./gradlew checkDexer
                 shell: bash
 
             -   name: Upload windows build code coverage
@@ -43,5 +55,5 @@ jobs:
                     #fail_ci_if_error: true
 
             -   name: composite build atrium-samples
-                run: ./gradlew build
+                run: ./gradlew build --scan
                 working-directory: misc\tools\atrium-samples-test
--- a/.github/workflows/samples-windows.yml
+++ b/.github/workflows/samples-windows.yml
@@ -40,7 +40,7 @@ jobs:
                 working-directory: samples\jvm\junit5
 
             -   name: build MPP sample
-                run: ./gradlew build
+                run: ./gradlew build --debug
                 working-directory: samples\multiplatform
 
             # Kotlin does not yet support to generate jdk 14 byte code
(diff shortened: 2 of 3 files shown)


assaflei:
@robstoll I managed to make the required cache entries on maven, gradle and android jar
It seems like maven cache folder is empty but the rest seem to work

Code is squashed (finally got it to work :))

Update with any remarks

.github/workflows/java-windows.yml
```diff
@@ -17,12 +19,38 @@ jobs:
...
+                    key: ${{ runner.os }}-maven-${{matrix.java_version}}-${{ hashFiles('**/pom.xml') }}
+                    restore-keys: |
+                        ${{ runner.os }}-maven-
+
+            -   name: Cache android jar
+                id: android-cache
+                uses: actions/cache@v1
+                with:
+                    path: android-jar-cache
+                    key: ${{ runner.os }}-android-version-${{matrix.java_version}}-${{ env.android-version }}
```
robstoll, line 43: what is the reason that you want to cache the android.jar per java version? It's always the same and thus could be shared no?

.github/workflows/java-windows.yml
```diff
@@ -17,12 +19,38 @@ jobs:
                 with:
                     java-version: ${{ matrix.java_version }}
 
+            -   name: Cache gradle dependencies
+                uses: actions/cache@v1
+                with:
+                    path: ~/.gradle/caches
+                    key: ${{ runner.os }}-gradle-${{matrix.java_version}}-${{ hashFiles('**/*.gradle*') }}
```
robstoll, line 26: what is the reason that you want to cache the gradle dependencies per java version? It's always the same and thus could be shared no?

.github/workflows/java-windows.yml
```diff
@@ -17,12 +19,38 @@ jobs:
...
+                with:
+                    path: android-jar-cache
+                    key: ${{ runner.os }}-android-version-${{matrix.java_version}}-${{ env.android-version }}
+
             -   name: buildNonDeprecatedJvm
                 run: ./gradlew buildNonDeprecatedJvm
 
-            -   name: download Android SDK's android.jar
+            -   name: download Android SDK's android.jar (or use cached file)
+                if: steps.android-cache.outputs.cache-hit != 'true'
```
robstoll, line 49: it looks like it always downloads and never uses the cache. But maybe due to ${{matrix.java_version}}, see if this already helps

.github/workflows/java-windows.yml
```diff
@@ -45,3 +73,6 @@ jobs:
             -   name: composite build atrium-samples
                 run: ./gradlew build
                 working-directory: misc\tools\atrium-samples-test
+
+            -   name: Stop gradle daemons
+                run: ./gradlew --stop
```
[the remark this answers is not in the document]
assaflei: the caching process was trying to access files in the cache folder and received access denied, so I had to close the daemons to make sure they are not using the files

.github/workflows/java-windows.yml
assaflei: I thought of future support where we might want to differentiate processes according to java versions
It can be removed if not relevant

.github/workflows/java-windows.yml
assaflei: Same as for android, since we differentiate builds per java we might want to differentiate these as well

.github/workflows/java-windows.yml
assaflei: it looks working here: https://github.com/robstoll/atrium/pull/500/checks?check_run_id=688503736

.github/workflows/java-windows.yml
robstoll: I would remove it and add later once required

.github/workflows/java-windows.yml
robstoll: In thos case please remove as well

.github/workflows/java-windows.yml
robstoll: Nice, let's see if it still does in the latest build

robstoll: It still does :+1:

.github/workflows/java-windows.yml
```diff
@@ -17,12 +19,38 @@ jobs:
...
+                    restore-keys: |
+                        ${{ runner.os }}-gradle-
+
+            -   name: Cache maven dependencies
+                uses: actions/cache@v1
+                with:
+                    path: ~/.m2
+                    key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+                    restore-keys: |
+                        ${{ runner.os }}-maven-
```
robstoll, line 36: you can delete this here (only required in samples-workflow.yml)

.github/workflows/java-windows.yml
```diff
@@ -17,12 +19,38 @@ jobs:
                 with:
                     java-version: ${{ matrix.java_version }}
 
+            -   name: Cache gradle dependencies
+                uses: actions/cache@v1
+                with:
+                    path: ~/.gradle/caches
+                    key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
+                    restore-keys: |
+                        ${{ runner.os }}-gradle-
```
robstoll, line 28: please copy to samples-workflow.yml there are also gradle based sample projects

robstoll:
I see you master squashing now :) one thing though, please still use a good commit message `squashed latest commits` isn't really helpful if one goes through the history :wink: Use something like `add caching to github actions`

robstoll:
@assaflei do you already have an idea why the build is failing? Seems strange to me

robstoll:
@assaflei I have pushed your PR to another branch and whatsoever I do, the build always passed but for one build.

But... I also had a look at the build times, for instance:
no-cache: https://github.com/robstoll/atrium/runs/695382522?check_suite_focus=true -- 13m 3s
cache: https://github.com/robstoll/atrium/runs/695958080?check_suite_focus=true -- 12m 47s

Then I have to say, maybe we better close this PR at this point (thanks for your work) as adding caching introduces more complexity to the yml files with very little gain. What do you think?

assaflei:
I think that we can leave the caching of the android jar since it can control execution of a whole step and it works pretty smoothly
Regarding the rest of the steps I too saw that the error is random and I feel like execution times are not really affected
So yes, we can remove the gradle and maven steps

robstoll:
Good, let's do it this way, only cache the android.jar in the end. I wait for your adjustment

robstoll:
@assaflei and in the end it failed because the cache was not per java version :thinking:
I will try and see, if this was the problem then I will take your last commit https://github.com/robstoll/atrium/commit/3db902d8bb379d880af6f3bcaf429285e0952d92 before the maven/gradle removal and create another PR.

robstoll:
@assaflei how about https://github.com/robstoll/atrium/issues/506 as next issue?

PR #521, opened 2020-06-07
configure codecov to not report before 7 builds

robstoll:

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -28,3 +28,9 @@ coverage:
         patch:
             default:
                 target: 85%
+
+odecov:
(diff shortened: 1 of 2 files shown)


.github/.codecov.yml
```diff
@@ -28,3 +28,9 @@ coverage:
         patch:
             default:
                 target: 85%
+
+odecov:
```
jGleitz, line 32: Are you sure that the is no `c` missing here?

robstoll: thanks for spotting this, must be the reason why it did not work. I'll close this one and open a new one to test it.

Issue #329, opened 2020-01-22
rename ...FeatureExpectationSpec to ...ExpectationsSpec and fuse if existing


robstoll:
*Platform* (jvm, js, android): all
*Extension* (none, kotlin 1.3, jdk8): all

## Code related feature
Wait until #328 is carried out.

I think the distinction between ...ExpectationsSpec and ...FeatureExpectationSpec is more confusing than helpful. Therefore rename ...FeatureExpectationSpec to ...ExpectationsSpec where there isn't an existing ...ExpectationsSpec yet or fuse it into an existing ...ExpectationsSpec

Gugatke:
i'll work on it

tarczynskitomek:
Hi! I'll be happy to help if the issue is still available. Just to reiterate what needs to be done:
Find all `.*FeatureAssertionSpec` and rename to `.*AssertionSpec`. If one exists, merge/fuse two specs into one.

robstoll:
Exactly, no need to do everything in one PR. I suggest you rewrite one spec and create a PR

tarczynskitomek:
Sounds like a plan! I'll start with fusing the example `MapFeatureAssertionsSpec`, and we'll see how it goes. 

There're some minor differences I've noticed in the spec config between `MapFeatureAssertionsSpec` and `MapAssertionsSpec`, like different values in nullableMap property, and the latter class using `Map<out String, Int>` while the former an invariant version `Map<String, Int>`, so this is some things to be considered while joining both classes.

robstoll:
@tarczynskitomek unassigning you so that others know they could work on this. Let us know in case you are still working of it and I will re-assign you again.

tarczynskitomek:
Hi @robstoll sorry for the delay, I'll be able to get back to this at the end of the week - plenty of other tasks right now :-(

robstoll:
No worries, I'll let you un-assigned in case someone else want's to contribute in the meantime. I'll re-assign you as soon as you have time again.

shytnik-igor:
I didn't find any files ...FeatureAssertionSpec in the project.
Is this issue still actually/correct?

robstoll:
The files where renamed from ...AssertionSpec to ...ExpectationSpec

robstoll:
@shytnik-igor I have updated the description. Let me know in case you don't intend to work on it, then I am un-assigning you again

shytnik-igor:
I'll work on this

PR #540, opened 2020-07-12
#329 rename ...FeatureAssertionSpec to ...AssertionSpec and fuse if existing 

Gugatke:
#329 implemented


robstoll:
I am closing this PR as you need to revert all changes anyway. Please create a new PR with the new changes
You misunderstood what needs to be done here. You need to rename the suffix `FeatureAssertionSpec` to `AssertionSpec` and fuse the spec in case there exists already a spec with this suffix. For instance:
- MapFeatureAssertionsSpec -> rename to MapAssertionsSpec => as there is already a MapAssertionsSpec, fuse the two

Issue #477, opened 2020-04-26
deprecate defaultTranslationOf(IgnoringCase)


robstoll:
*Platform* (jvm, js, android): all
*Extension* (none, kotlin 1.3, jdk8): none

## Code related feature
We want to remove this functionality in 1.0.0 and thus should deprecate it beforehand

Following the things you need to do:

*lib*
- [ ] add `TODO remove with 1.0.0` to  _containsDefaultTranslationOf and _containsDefaultTranslationOfIgnoringCase
     
*domain*
- [ ] deprecate CharSequenceContainsAssertions.defaultTranslationOf with `@Deprecated("Will be removed with 1.0.0") no need for a replacement
- [ ]  same same but different for CharSequenceContainsAssertions.defaultTranslationOfIgnoringCase 

- [ ] suppress deprecation in CharSequenceContainsAssertionsBuilder

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime, we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

Gugatke:
I'll work on this

robstoll:
@Gugatke let me know in case you need help

robstoll:
@Gugatke un-assinging you so that others can take over

z13z:
I'll work on this

PR #542, opened 2020-07-12
#477 deprecate defaultTranslationOf(IgnoringCase)

z13z:


misc/deprecated/domain/api/atrium-domain-api-common/src/main/kotlin/ch/tutteli/atrium/domain/creating/charsequence/contains/creators/CharSequenceContainsAssertions.kt
```diff
@@ -32,11 +32,25 @@ interface CharSequenceContainsAssertions {
         expected: List<CharSequenceOrNumberOrChar>
     ): AssertionGroup
 
+    //TODO remove with 1.0.0
+    @Deprecated("Use AssertImpl.charSequence.contains.defaultTranslationOf; will be removed with 1.0.0",
+        ReplaceWith(
+            "AssertImpl.charSequence.contains.defaultTranslationOf(checker, expected, *otherExpected)",
+            "ch.tutteli.atrium.creating.AssertImpl"
+        )
+    )
```
robstoll, line 41: You suggest to replace it with the same thing :wink:
I see that the description of the task is no longer up-to-date. We don't plan to provide a replacement, thus change to:
```suggestion
    @Deprecated("Will be removed with 1.0.0")
```
Same for other places

z13z:
Hello
I just rebased my current branch (477) on upstream/master form intellij (rebase current onto selected) and than pushed

robstoll:
I suggest you squash your commits first and then to a rebase. See https://github.com/robstoll/atrium/pull/440#issuecomment-619102893 for a description how I usually do this

PR #544, opened 2020-07-13
removed redundant class full path use

Gugatke:
removed redundant class full path use


misc/verbs-internal/atrium-verbs-internal-js/src/main/kotlin/ch/tutteli/atrium/verbs/internal/registerServices.kt
```diff
@@ -7,5 +7,5 @@ import ch.tutteli.atrium.core.polyfills.registerService
 @Suppress("unused" /* here in order that the code is executed when module is loaded */)
 private val register = run {
 
-    registerService<ch.tutteli.atrium.reporting.ReporterFactory> { ch.tutteli.atrium.verbs.internal.NoAdjustingReporterFactory() }
+    registerService<ch.tutteli.atrium.reporting.ReporterFactory> { NoAdjustingReporterFactory() }
```
robstoll, line 10: thanks for the idea but registerServices.kt is generated

PR #546, opened 2020-07-13
API #477 #comment added required deprecations again

z13z:


misc/deprecated/domain/api/atrium-domain-api-common/src/main/kotlin/ch/tutteli/atrium/domain/creating/charsequence/contains/creators/CharSequenceContainsAssertions.kt
```diff
@@ -32,11 +32,25 @@ interface CharSequenceContainsAssertions {
         expected: List<CharSequenceOrNumberOrChar>
     ): AssertionGroup
 
+    //TODO remove with 1.0.0
+    @Deprecated("Use AssertImpl.charSequence.contains.defaultTranslationOf; will be removed with 1.0.0",
```
robstoll, line 36: @z13z oh well, this looks like an old version and not the latest you had in the other PR. And I already merged :scream: 
Can you please fix it with another PR? thanks

Issue #337, opened 2020-01-31
filter  jasmine out of error-stacks


robstoll:
*Platform* (jvm, js, android):
*Extension* (none, kotlin 1.3, jdk8): none

## Error Reporting related feature
Currently we filter only stackframes of mocha, we could enhance it with stackframes of jasmine
Following the things you need to do:

*lib*
- [ ] core/robstoll-lib/atrium-core-robstoll-lib-js/src/main/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/RemoveRunnerAtriumErrorAdjuster.kt => add jasmine
- [ ] add corresponding tests to ch.tutteli.atrium.core.robstoll.lib.reporting.AdjustStackTest

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime, we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

sabankay1120:
Send me more information and I will work on this

robstoll:
@sabankay1120 what information do you need in addition?

z13z:
I'll work on this

robstoll:
@z13z great that you are that motivated :slightly_smiling_face: Please don't work on more issues in parallel, let's try to get the once you are assigned to into master first.

z13z:
I can't find mocha in README.md (inside notToThrow). Was it deleted?

robstoll:
you find `mocha` around line 496
`Stack frames of Atrium and of test runners (Spek, Kotlintest and JUnit for JVM, mocha for JS) are excluded per default`

z13z:
Errors in AdjustStackTest doesn't contain string "jasmine", even after changing regex in RemoveRunnerAtriumErrorAdjuster.kt to `(\s|\S)*`. how can I produce exception with correct stack trace in given test class to add new test cases

robstoll:
[...]

PR #548, opened 2020-07-15
API #337 #comment added jasmine filter for stackframes

z13z:

### DIFF
--- a/core/robstoll-lib/atrium-core-robstoll-lib-js/src/main/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/RemoveRunnerAtriumErrorAdjuster.kt
+++ b/core/robstoll-lib/atrium-core-robstoll-lib-js/src/main/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/RemoveRunnerAtriumErrorAdjuster.kt
@@ -8,6 +8,6 @@ actual class RemoveRunnerAtriumErrorAdjuster : FilterAtriumErrorAdjuster(), Atri
     }
 
     companion object {
-        val runnerRegex: Regex = Regex("[\\\\|/]mocha[\\\\|/]")
+        val runnerRegex: Regex = Regex("[\\\\|/](mocha|jasmine)[\\\\|/]")
     }
 }
--- a/samples/js/jasmine/src/test/kotlin/org/atriumlib/samples/jasmine/SampleJsTest.kt
+++ b/samples/js/jasmine/src/test/kotlin/org/atriumlib/samples/jasmine/SampleJsTest.kt
@@ -27,6 +27,7 @@ class SampleJsTest {
             throw IllegalArgumentException("oho... hello btw")
         }.toThrow<IllegalArgumentException> {
             messageContains("hello")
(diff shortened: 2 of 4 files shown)


core/robstoll-lib/atrium-core-robstoll-lib-js/src/main/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/RemoveRunnerAtriumErrorAdjuster.kt
```diff
@@ -8,6 +8,6 @@ actual class RemoveRunnerAtriumErrorAdjuster : FilterAtriumErrorAdjuster(), Atri
     }
 
     companion object {
-        val runnerRegex: Regex = Regex("[\\\\|/]mocha[\\\\|/]")
+        val runnerRegex: Regex = Regex("[\\\\|/](mocha|jasmin)[\\\\|/]")
```
robstoll, line 11: shouldn't it be jasmine?

samples/js/jasmine/src/test/kotlin/org/atriumlib/samples/jasmine/SampleJsTest.kt
```diff
@@ -27,6 +27,7 @@ class SampleJsTest {
             throw IllegalArgumentException("oho... hello btw")
         }.toThrow<IllegalArgumentException> {
             messageContains("hello")
+            messageContains("jasmine")
```
robstoll, line 30: we want to check that it does not contain jasmine, your task is to filter it out. use:
```
message {
  contains("hello")
  //TODO your code
}
```

robstoll:
@z13z you still doing something wrong. See, f82ec14 is basically the same commit as you had in your other PR. 
Please squash your changes and force push again.

samples/js/jasmine/src/test/kotlin/org/atriumlib/samples/jasmine/SampleJsTest.kt
z13z: you wan't to change messageContains method right? like this:
`    @Test
    fun expectAnExceptionWithAMessageOccurred() {
        expect {
            throw IllegalArgumentException("oho... hello btw")
        }.toThrow<IllegalArgumentException> {
            message {
                contains("hello")
                contains("jasmine")
            }
        }
    }`

robstoll: No. The test currently fails because you are asserting the wrong thing. You filtered jasmine out of the stacktrace. Why would you expect it to be there?

robstoll: I can give you the answer if you want but I thought you might want to come up with it yourself. see hint above.
instead of 
```
            messageContains("hello")
            messageContains("jasmine")
```
you have to use 
```
message {
  contains("hello")
  //TODO your code
}
```

Issue #481, opened 2020-04-28
accept date and time as string in ISO 8601 format for ChronoLocalDateTime


robstoll:
*Platform* (jvm, js, android): jvm
*Extension* (none, kotlin 1.3, jdk8): jdk8

## Code related feature
```
expect(LocalDateTime.now()).isAfter("2020-04-28T12:00")

//instead of

expect(LocalDateTime.now()).isAfter(LocalDateTime.of(2020,4,28,12,0))
```

Following the things you need to do:

*atrium-logic*:
- [ ] duplicate each method in chronoLocalDateTimeAssertions.kt and:
  - [ ] accept a String instead of a ChronoLocalDateTime and parse it to LocalDateTime and pass it to the existing overload where the following formats shall be supported. This is a slight deviation of ISO 8601, e.g. we do not want to support the alternative notation e.g. 20200401120001 instead of 2020-04-01T12:00:01
        * yyyy-mm-ddThh:mm:ss.sss (up to 9 digits for nanoseconds)
        * yyyy-mm-ddThh:mm:ss
        * yyyy-mm-ddThh:mm
        * yyyy-mm-dd
        => hint: all but the last point is supported by LocalDateTime.parse. 
        Note that #482 deals also with parsing DateTime. You should reuse the logic from there if #482 is already implemented.

*api-fluent*
- [ ] duplicate each method in chronoLocalDateTimeAssertions.kt and:
  - [ ] accept a String instead of a ChronoLocalDateTime - call the corresponding domain function
  - [ ] change since to 0.12.0 (adapt to current [milestone](https://github.com/robstoll/atrium/milestones))
- [ ] add the string version to ChronoLocalDateTimeAssertionsSpec:
  - [ ] add an object StringSpec, create functions in the companion object which corresponds to the signature of the existing functions (i.e. accept a ChronoLocalDateTime) where inside you turn the ChronoLocalDateTime into a string in ISO format and pass it to your new function
 - [ ] add `describe("wrong string format")` after the `include(...)` and add the following test-cases (use a map and forEach to generate the test-cases -- see DelegatingAssertionCheckerSpec for an example):
   - one passes a DateTime in format yyyymmddhhmmssSSS
   - one passes a DateTime in format yyyymmddhhmmss
   - one passes a DateTime in format yyyymmddhhmm
   - one passes a DateTime in format yyyymmdd
   - one passes a DateTime in format yyyy-m-ddThh:mm
[...]

PR #552, opened 2020-07-26
481 accept date and time as string in ISO 8601 format for ChronoLocalDateTime 2

rkiselev:
Hi @robstoll, thanks for your answers.  I think that I have implemented as you said,  but I get an error on build:
ChronoLocalDateTimeAssertionSpec.kt: (18, 9): Type inference failed: inline fun <T, A1> fun1(f: KFunction2<Expect<T>, A1, Expect<T>>): Fun1<T, A1> /* = Pair<String, Expect<T>.(A1) -> Expect<T>> */
cannot be applied to
(KFunction2<Expect<ChronoLocalDateTime<*>>, ChronoLocalDateTime<*>, Expect<ChronoLocalDateTime<*>>>)

and others for each method, could you help me?


robstoll:
@rkiselev can you please base your work on my branch. It already contains all changes you did in the other PR (see instructions in the other PR on how to checkout the branch). This way I see immediately what you have changed since then. thanks.

rkiselev:
@robstoll, I can't change the source branch of PR, but after you have changed PR from master to #481-ChronoLocalDateTime-as-string at file changed tab there are only my last changes from today. Is it enough, or I need to close this PR and create new?

robstoll:
I would expect to see only one commit but it's OK if the changes I see now are the new ones. I'll look at your problem tomorrow

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/chronoLocalDateTimeAssertions.kt
```diff
@@ -74,3 +74,68 @@ fun <T : ChronoLocalDateTime<out ChronoLocalDate>> Expect<T>.isAfterOrEqual(
 fun <T : ChronoLocalDateTime<out ChronoLocalDate>> Expect<T>.isEqual(
     expected: ChronoLocalDateTime<*>
 ): Expect<T> = _logicAppend { isEqual(expected) }
+
+/**
+ * Expects that the subject of the assertion (a [ChronoLocalDateTime])
+ * is before the [expected] [String].
```
robstoll, line 80: Please also add to the other KDoc
```suggestion
 * is before the [expected] [LocalDateTime] given as [String] in (modified) ISO 8601 format.
 * 
 * The string will be converted to a LocalDateTime according to ISO 8601 but with a slight deviation. The alternative notation (e.g. 20200401120001 instead of 2020-04-01T12:00:01) is not supported and we accept a date without time in which case 00:00:00 is used. Following the supported formats from the longest to the shortest:
* * yyyy-mm-ddThh:mm:ss.sss (up to 9 digits for nanoseconds)
* * yyyy-mm-ddThh:mm:ss
* * yyyy-mm-ddThh:mm
* * yyyy-mm-dd
```

rkiselev:
thanks, it works.

rkiselev:
hi, @robstoll, could you please check my changes, I have finished implementation.

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/ChronoLocalDateTimeAssertionSpec.kt
```diff
@@ -1,46 +1,183 @@
...
+        "2020010101" to "wrong string format",
+        "2020-1-01T01:01" to "wrong string format",
+        "2020-01-1T01:01" to "wrong string format",
+        "20-01-01T01:01" to "wrong string format",
+        "2020-01-01T1:01" to "wrong string format",
+        "2020-01-01T01:1" to "wrong string format",
+        "2020-01-01T01:01:1" to "wrong string format",
+        "2020-01-01t01:01:01" to "wrong string format"
+    ).forEach{(value, description) ->
+        run {
```
robstoll, line 33: I would remove this (not necessary)
```suggestion
        
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/ChronoLocalDateTimeAssertionSpec.kt
```diff
@@ -1,46 +1,183 @@
...
+            val now = expect(LocalDateTime.now())
+            describe(description) {
```
robstoll, line 35: Since the description is always the same, I would move it outside of the forEach and turn the mapOf into a listOf. Instead I would use the `value` here.

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/ChronoLocalDateTimeAssertionSpec.kt
```diff
@@ -1,46 +1,183 @@
...
+                it("throws an DateTimeParseException") {
```
robstoll, line 36: Please add the name of the function (also ain the other places)
```suggestion
                it("isBefore -- throws an DateTimeParseException") {
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/ChronoLocalDateTimeAssertionSpec.kt
```diff
@@ -1,46 +1,183 @@
...
+                        now.isBefore(value)
+                    }.toThrow<AtriumError> { messageContains("is before: $value") }
+                }
+                it("doesn't throw any Error") {
+                    expect {
+                        now.isBefore(before.format(ISO_DATE_TIME))
+                    }
+                }
+                it("doesn't throw any Error") {
+                    expect {
```
robstoll, line 86: remove  `expect{...}` as this creates an Expect of a lambda which is never executed. Alternatively you could use `notToThrow` to verify it (instead of toThrow) but we haven't use this so far and I prefer the shorter version

robstoll:
@rkiselev thanks for the update. I am going to adjust a few more things and add the infix API and then merge into master. Thanks for the endurance and your work :slightly_smiling_face: :+1: 
Fancy to tackle the other related issues? I guess they should be easy now.

PR #554, opened 2020-07-29
Add Any.isNoneOf and Any.isNotIn assertions

anesabml:

### DIFF
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
@@ -177,7 +177,7 @@ infix fun <T> Expect<T>.and(assertionCreator: Expect<T>.() -> Unit): Expect<T> =
  *
  * @since 0.13.0
  */
-inline fun <reified T> Expect<T>.isNoneOf(expected: T, vararg otherValues: T): Expect<T> =
+fun <T> Expect<T>.isNoneOf(expected: T, vararg otherValues: T): Expect<T> =
     _logicAppend { isNotIn(expected glue otherValues) }
 
 /**
@@ -194,7 +194,7 @@ inline fun <reified T> Expect<T>.isNoneOf(expected: T, vararg otherValues: T): E
  */
 inline fun <reified T> Expect<T>.isNotIn(expected: IterableLike): Expect<T> {
     val iterable = iterableLikeToIterable<T>(expected)
-    require(iterable.iterator().hasNext()) { "Iterable without elements are not allowed for this function." }
+    require(iterable.iterator().hasNext()) { "IterableLike without elements are not allowed for this function." }
     return _logicAppend { isNotIn(iterable.toList()) }
 }
 
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
@@ -1,6 +1,8 @@
 package ch.tutteli.atrium.api.fluent.en_GB
 
 import ch.tutteli.atrium.creating.Expect
+import ch.tutteli.atrium.domain.builders.utils.iterableLikeToIterable
+import ch.tutteli.atrium.domain.creating.typeutils.IterableLike
 import ch.tutteli.atrium.logic.*
 import ch.tutteli.atrium.reporting.Reporter
 import ch.tutteli.kbox.glue
@@ -173,27 +175,27 @@ infix fun <T> Expect<T>.and(assertionCreator: Expect<T>.() -> Unit): Expect<T> =
  * @return An [Expect] for the current subject of the assertion.
  * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
  *
- * @since 0.9.0
+ * @since 0.13.0
  */
 inline fun <reified T> Expect<T>.isNoneOf(expected: T, vararg otherValues: T): Expect<T> =
-    _logicAppend {
-        isNotIn(expected glue otherValues)
-    }
+    _logicAppend { isNotIn(expected glue otherValues) }
 
 /**
  * Expects that the subject of the assertion is not (equal to) any value of [expected].
+ *
+ * Notice that a runtime check applies which assures that only [Iterable], [Sequence] or one of the [Array] types
+ * are passed. This function expects [IterableLike] (which is a typealias for [Any]) to avoid cluttering the API.
  *
  * @return An [Expect] for the current subject of the assertion.
  * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
  * @throws IllegalArgumentException in case the iterable is empty.
  *
- * @since 0.9.0
+ * @since 0.13.0
  */
-inline fun <reified T> Expect<T>.isNotIn(expected: Iterable<T>): Expect<T> {
-    require(expected.iterator().hasNext()) { "Iterable without elements are not allowed for this function." }
-    return _logicAppend {
-        isNotIn(expected.toList())
-    }
+inline fun <reified T> Expect<T>.isNotIn(expected: IterableLike): Expect<T> {
+    val iterable = iterableLikeToIterable<T>(expected)
+    require(iterable.iterator().hasNext()) { "Iterable without elements are not allowed for this function." }
+    return _logicAppend { isNotIn(iterable.toList()) }
 }
 
 
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
@@ -3,6 +3,7 @@ package ch.tutteli.atrium.api.fluent.en_GB
 import ch.tutteli.atrium.creating.Expect
 import ch.tutteli.atrium.logic.*
 import ch.tutteli.atrium.reporting.Reporter
+import ch.tutteli.kbox.glue
 
 /**
  * Expects that the subject of the assertion is (equal to) [expected].
@@ -165,3 +166,34 @@ inline val <T> Expect<T>.and: Expect<T> get() = this
  */
 infix fun <T> Expect<T>.and(assertionCreator: Expect<T>.() -> Unit): Expect<T> =
     addAssertionsCreatedBy(assertionCreator)
+
+/**
+ * Expects that the subject of the assertion is not (equal to) [expected] and [otherValues].
+ *
+ * @return An [Expect] for the current subject of the assertion.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ *
+ * @since 0.9.0
+ */
+inline fun <reified T> Expect<T>.isNoneOf(expected: T, vararg otherValues: T): Expect<T> =
+    _logicAppend {
+        isNotIn(expected glue otherValues)
+    }
+
+/**
+ * Expects that the subject of the assertion is not (equal to) any value of [expected].
+ *
+ * @return An [Expect] for the current subject of the assertion.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ * @throws IllegalArgumentException in case the iterable is empty.
+ *
+ * @since 0.9.0
+ */
+inline fun <reified T> Expect<T>.isNotIn(expected: Iterable<T>): Expect<T> {
+    require(expected.iterator().hasNext()) { "Iterable without elements are not allowed for this function." }
+    return _logicAppend {
+        isNotIn(expected.toList())
+    }
+}
+
+
--- a/apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/anyAssertions.kt
+++ b/apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/anyAssertions.kt
@@ -1,11 +1,11 @@
 package ch.tutteli.atrium.api.infix.en_GB
 
+import ch.tutteli.atrium.api.infix.en_GB.creating.Values
 import ch.tutteli.atrium.creating.Expect
 import ch.tutteli.atrium.domain.builders.utils.iterableLikeToIterable
 import ch.tutteli.atrium.domain.creating.typeutils.IterableLike
 import ch.tutteli.atrium.logic.*
 import ch.tutteli.atrium.reporting.Reporter
-import ch.tutteli.kbox.glue
 
 /**
  * Expects that the subject of the assertion is (equal to) [expected].
@@ -229,15 +229,18 @@ inline val <T> Expect<T>.it: Expect<T> get() : Expect<T> = this
 inline val <T> Expect<T>.its: Expect<T> get() : Expect<T> = this
 
 /**
- * Expects that the subject of the assertion is not (equal to) [expected] and [otherValues].
+ * Expects that the subject of the assertion is not (equal to) in [values].
+ *
+ * @param values The values which are not expected to be contained within the subject of the assertion
+ *   -- use the function `values(t, ...)` to create a [Values].
  *
  * @return An [Expect] for the current subject of the assertion.
  * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
  *
  * @since 0.13.0
  */
-inline fun <reified T> Expect<T>.isNoneOf(expected: T, vararg otherValues: T): Expect<T> =
-    _logicAppend { isNotIn(expected glue otherValues) }
+infix fun <T> Expect<T>.isNoneOf(values: Values<T>): Expect<T> =
+    _logicAppend { isNotIn(values.toList()) }
 
 /**
  * Expects that the subject of the assertion is not (equal to) any value of [expected].
@@ -251,7 +254,7 @@ inline fun <reified T> Expect<T>.isNoneOf(expected: T, vararg otherValues: T): E
  *
  * @since 0.13.0
  */
-inline fun <reified T> Expect<T>.isNotIn(expected: IterableLike): Expect<T> {
+inline infix fun <reified T> Expect<T>.isNotIn(expected: IterableLike): Expect<T> {
     val iterable = iterableLikeToIterable<T>(expected)
     require(iterable.iterator().hasNext()) { "Iterable without elements are not allowed for this function." }
     return _logicAppend { isNotIn(iterable.toList()) }
--- a/apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/anyAssertions.kt
+++ b/apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/anyAssertions.kt
@@ -256,6 +256,6 @@ infix fun <T> Expect<T>.isNoneOf(values: Values<T>): Expect<T> =
  */
 inline infix fun <reified T> Expect<T>.isNotIn(expected: IterableLike): Expect<T> {
     val iterable = iterableLikeToIterable<T>(expected)
-    require(iterable.iterator().hasNext()) { "Iterable without elements are not allowed for this function." }
+    require(iterable.iterator().hasNext()) { "IterableLike without elements are not allowed for this function." }
     return _logicAppend { isNotIn(iterable.toList()) }
 }
--- a/apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/anyAssertions.kt
+++ b/apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/anyAssertions.kt
@@ -1,8 +1,11 @@
 package ch.tutteli.atrium.api.infix.en_GB
 
 import ch.tutteli.atrium.creating.Expect
+import ch.tutteli.atrium.domain.builders.utils.iterableLikeToIterable
+import ch.tutteli.atrium.domain.creating.typeutils.IterableLike
 import ch.tutteli.atrium.logic.*
 import ch.tutteli.atrium.reporting.Reporter
+import ch.tutteli.kbox.glue
 
 /**
  * Expects that the subject of the assertion is (equal to) [expected].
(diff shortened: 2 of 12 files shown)


anesabml:
@robstoll could you please check the `isNotIn` in `DefaultAnyAssertions` implementation I run some tests but they fail, I am not sure how I should implement it.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
```diff
@@ -165,3 +166,34 @@ inline val <T> Expect<T>.and: Expect<T> get() = this
...
+ *
+ * @return An [Expect] for the current subject of the assertion.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ *
+ * @since 0.9.0
+ */
+inline fun <reified T> Expect<T>.isNoneOf(expected: T, vararg otherValues: T): Expect<T> =
+    _logicAppend {
+        isNotIn(expected glue otherValues)
+    }
```
robstoll, line 181: since it is not much code I would put it on the same line
```suggestion
    _logicAppend { isNotIn(expected glue otherValues) }
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
```diff
@@ -165,3 +166,34 @@ inline val <T> Expect<T>.and: Expect<T> get() = this
...
 infix fun <T> Expect<T>.and(assertionCreator: Expect<T>.() -> Unit): Expect<T> =
     addAssertionsCreatedBy(assertionCreator)
+
+/**
+ * Expects that the subject of the assertion is not (equal to) [expected] and [otherValues].
+ *
+ * @return An [Expect] for the current subject of the assertion.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ *
+ * @since 0.9.0
```
robstoll, line 176:
```suggestion
 * @since 0.13.0
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
```diff
@@ -165,3 +166,34 @@ inline val <T> Expect<T>.and: Expect<T> get() = this
...
+    }
+
+/**
+ * Expects that the subject of the assertion is not (equal to) any value of [expected].
+ *
+ * @return An [Expect] for the current subject of the assertion.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ * @throws IllegalArgumentException in case the iterable is empty.
+ *
+ * @since 0.9.0
```
robstoll, line 190:
```suggestion
 * @since 0.13.0
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
```diff
@@ -165,3 +166,34 @@ inline val <T> Expect<T>.and: Expect<T> get() = this
...
+ */
+inline fun <reified T> Expect<T>.isNotIn(expected: Iterable<T>): Expect<T> {
```
robstoll, line 192: let's use  `IterableLike` :slightly_smiling_face: 
Please also add the note to the KDOC we have added to the other methods about IterableLike (please make sure you are on the latest master, I have improved it a bit).

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
```diff
@@ -165,3 +166,34 @@ inline val <T> Expect<T>.and: Expect<T> get() = this
...
+    require(expected.iterator().hasNext()) { "Iterable without elements are not allowed for this function." }
```
robstoll, line 193: Once you use `IterableLike` then re-use the logic from `toVarArg` to convert it to a Iterable.
Modify toVarArg, split it up into transforming to Iterable and calling iterableToPair (i.e. move iterableToPair out of the when and move the `when` into an own method `iterableLikeToIterable`)

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultAnyAssertions.kt
```diff
@@ -52,4 +53,13 @@ class DefaultAnyAssertions : AnyAssertions {
         container.changeSubject.reportBuilder()
             .downCastTo(subType)
             .build()
+
+    override fun <T> isNotIn(container: AssertionContainer<T>, expected: List<T>): Assertion {
```
robstoll, line 57: My fault, `List` is actually not required, let's rewrite to `Iterable`
```suggestion
    override fun <T> isNotIn(container: AssertionContainer<T>, expected: Iterable<T>): Assertion {
```

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultAnyAssertions.kt
```diff
@@ -52,4 +53,13 @@ class DefaultAnyAssertions : AnyAssertions {
...
+        val assertions = expected.map { assertionBuilder.representationOnly.failing.withRepresentation(it).build() }
```
robstoll, line 58:
```suggestion
        val assertions = expected.map { 
          assertionBuilder.representationOnly
            .withTest { /* TODO your assertion logic hear instead of hard coded */ false }
            .withRepresentation(it)
            .build() 
        }
```

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
```diff
@@ -234,13 +296,18 @@ abstract class AnyAssertionsSpec(
         notToBe: Fun1<T?, T?>,
         isSame: Fun1<T?, T?>,
         isNotSame: Fun1<T?, T?>,
-        value: T
+        isNoneOf: Fun2<T?, T?, Array<T?>>,
+        isNotIn: Fun1<T?, Iterable<T?>>,
+        value: T,
+        emptyArray: Array<T?>
```
anesabml, line 302: I added an empty Array parameter because I can't instantiate an array inside this function, it needs to be an inline function, if you have any suggestion on how to avoid this please let me know.

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/anyAssertions.kt
```diff
@@ -224,3 +227,32 @@ inline val <T> Expect<T>.it: Expect<T> get() : Expect<T> = this
...
+
+/**
+ * Expects that the subject of the assertion is not (equal to) [expected] and [otherValues].
+ *
+ * @return An [Expect] for the current subject of the assertion.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ *
+ * @since 0.13.0
+ */
+inline fun <reified T> Expect<T>.isNoneOf(expected: T, vararg otherValues: T): Expect<T> =
```
robstoll, line 239: This one we need to re-work a bit as it will not work out for the infix API.
```suggestion
infix fun <T> Expect<T>.isNoneOf(values: Values<T>): Expect<T> =
```
See other places which use`Values` and add a note how Values can be constructed to KDoc
Note that you need to add `infix` as well in order that it can be used as infix function. I guess that `inline` and `reified` are not needed here but I might be wrong.

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/anyAssertions.kt
```diff
@@ -224,3 +227,32 @@ inline val <T> Expect<T>.it: Expect<T> get() : Expect<T> = this
...
+ * Notice that a runtime check applies which assures that only [Iterable], [Sequence] or one of the [Array] types
+ * are passed. This function expects [IterableLike] (which is a typealias for [Any]) to avoid cluttering the API.
+ *
+ * @return An [Expect] for the current subject of the assertion.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ * @throws IllegalArgumentException in case the iterable is empty.
+ *
+ * @since 0.13.0
+ */
+inline fun <reified T> Expect<T>.isNotIn(expected: IterableLike): Expect<T> {
```
robstoll, line 254:
```suggestion
infix fun <T> Expect<T>.isNotIn(expected: IterableLike): Expect<T> {
```
most important is that you add `infix` but I guess we don't need inline nor reified here.

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/AnyAssertionsSpec.kt
```diff
@@ -25,6 +26,14 @@ class AnyAssertionsSpec : ch.tutteli.atrium.specs.integration.AnyAssertionsSpec(
...
     fun1(Expect<Int?>::isNotSameAs).withNullableSuffix(),
     fun1(Expect<DataClass?>::isNotSameAs).withNullableSuffix(),
+    fun2(Expect<Int>::isNoneOf),
+    fun2(Expect<DataClass>::isNoneOf),
+    fun2(Expect<Int?>::isNoneOf).withNullableSuffix(),
+    fun2(Expect<DataClass?>::isNoneOf).withNullableSuffix(),
+    fun1(Expect<Int>::isNotIn),
+    fun1(Expect<DataClass>::isNotIn),
+    fun1(Expect<Int?>::isNotIn).withNullableSuffix(),
+    fun1(Expect<DataClass?>::isNotIn).withNullableSuffix(),
```
robstoll, line 36: please add isNoneOf and isNotIn to ambiguityTest below (this way you will see that your current code does not work out for the infix API)

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultAnyAssertions.kt
```diff
@@ -54,8 +55,13 @@ class DefaultAnyAssertions : AnyAssertions {
             .downCastTo(subType)
             .build()
 
-    override fun <T> isNotIn(container: AssertionContainer<T>, expected: List<T>): Assertion {
-        val assertions = expected.map { assertionBuilder.representationOnly.failing.withRepresentation(it).build() }
+    override fun <T> isNotIn(container: AssertionContainer<T>, expected: Iterable<T>): Assertion {
+        val assertions = expected.map { value ->
+            assertionBuilder.representationOnly
+                .withTest { container.maybeSubject.fold(trueProvider) { it != value } }
```
robstoll, line 61: my bad, there is a better overload than I suggested in the last comment:
```suggestion
                .withTest(container) { it != value }
```

misc/deprecated/domain/builders/atrium-domain-builders-common/src/main/kotlin/ch/tutteli/atrium/domain/builders/utils/VarArgHelper.kt
```diff
@@ -37,18 +37,21 @@ interface VarArgHelper<out T> {
...
+        is Array<*> -> iterableLike.map { it as T }
+        is CharArray -> iterableLike.map { it as T }
+        is ByteArray -> iterableLike.map { it as T }
+        is ShortArray -> iterableLike.map { it as T }
+        is IntArray -> iterableLike.map { it as T }
+        is LongArray -> iterableLike.map { it as T }
+        is FloatArray -> iterableLike.map { it as T }
+        is DoubleArray -> iterableLike.map { it as T }
+        is BooleanArray -> iterableLike.map { it as T }
         else -> throw IllegalArgumentException("toVarArg accepts arguments of types Iterable, Sequence, Array")
```
robstoll, line 55:
```suggestion
        else -> throw IllegalArgumentException("iterableLikeToIterable accepts arguments of types Iterable, Sequence, Array")
```

misc/deprecated/domain/builders/atrium-domain-builders-common/src/main/kotlin/ch/tutteli/atrium/domain/builders/utils/VarArgHelper.kt
```diff
@@ -37,18 +37,21 @@ interface VarArgHelper<out T> {
  * @throws IllegalArgumentException in case the iterable is empty.
  */
 inline fun <reified T> toVarArg(iterableLike: IterableLike): Pair<T, Array<out T>> =
+    iterableToPair(iterableLikeToIterable(iterableLike))
+
+inline fun <reified T> iterableLikeToIterable(iterableLike: IterableLike): Iterable<T> =
```
robstoll, line 42: please check if `inline` and `reified` are really required.

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
```diff
@@ -132,6 +148,16 @@ abstract class AnyAssertionsSpec(
                         expectSubject.isNotSameFun(1)
                     }.toThrow<AssertionError> { messageContains(IS_NOT_SAME.getDefault()) }
                 }
+                it("${isNoneOf.name} throws AssertionError") {
+                    expect {
+                        expectSubject.isNoneOfFun(1, emptyArray())
```
robstoll, line 153: Please use two numbers here (I know it does not 100% fit the context but I think that's fine -- or create another context if you want)

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
```diff
@@ -132,6 +148,16 @@ abstract class AnyAssertionsSpec(
...
+                    }.toThrow<AssertionError> { messageContains(IS_NONE_OF.getDefault()) }
```
robstoll, line 154:
```suggestion
                    }.toThrow<AssertionError> { 
                      message {
                        contains(IS_NONE_OF.getDefault()) 
                        //TODO check that only failing assertions are reported
                     }
                  }
```

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
```diff
@@ -132,6 +148,16 @@ abstract class AnyAssertionsSpec(
...
+                }
+                it("${isNotIn.name} throws AssertionError") {
+                    expect {
+                        expectSubject.isNotInFun(listOf(1))
```
robstoll, line 158: same here

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
robstoll: Fine with me. I guess you could change it to `inline fun <reified T : Any>` and create an empty array inside

robstoll:
@anesabml the build fails because the new functions are missing in the scala API of Atrium. Totally fine if I provide this after your PR. Unless you know a bit scala and would like to provide it as well of course.

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
anesabml: Unfortunately, Kotlin doesn't support nested inline functions is not supported yet so I can't create an array inside the function.

robstoll: true, local inline functions are not supported

misc/deprecated/domain/builders/atrium-domain-builders-common/src/main/kotlin/ch/tutteli/atrium/domain/builders/utils/VarArgHelper.kt
anesabml: if we remove `inline` and `reified` from `iterableLikeToIterable` we will get a warning in the mapping `Unchecked cast: Any? to T`, Should I remove `inline` and `reified` and suppress the warning?

robstoll: ah right forgot about that. no keep it in this case (and in the other places I said maybe we can remove it)

anesabml:
@robstoll since `isNoneOf` parameters are different in api-fluent and api-infix I can't pass the `isNoneOf` parameters to `integration.AnyAssertionsSpec`, I have a solution in mind and it is to create another non-infix `isNoneOf` function similar to the one in api-fluent and use it, please if you have other suggestions I would appreciate it. Thank you

robstoll:
@anesabml I assume you are now using `Values` for the infix api. Your approach is the same we already took. See for instance apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/CharSequenceContainsAtLeastAssertionsSpec.kt => containsAtLeast

Which means, create also a corresponding non-infix function in the companion object of AnyAssertionsSpec in the infix API

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
```diff
@@ -151,6 +167,36 @@ abstract class AnyAssertionsSpec(
...
                 }
             }
+            context("one equals only one of the others") {
+                it("${isNoneOf.name} throws AssertionError") {
+                    expect {
+                        expectSubject.isNoneOfFun(1, arrayOf(2))
+                    }.toThrow<AssertionError> {
+                        message {
+//                            contains(IS_NONE_OF.getDefault())
+                            contains("1")
```
robstoll, line 177:
```suggestion
                            contains(IS_NONE_OF.getDefault())
                            contains("1")
                            containsNot("2")
```
there is a high chance that this will not pass as `1` and `2` is not enough context (might be in the string nonetheless, e.g. as part of an identity hash).
I suggest you add `listBulletPoint: String,` as argument to the spec (see for instance IterableNoneAssertionsSpec) and then check:
```suggestion
                            contains(IS_NONE_OF.getDefault(), "$listBulletPoint 1)
                            containsNot("$listBulletPoint 2")
```

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
```diff
@@ -151,6 +167,36 @@ abstract class AnyAssertionsSpec(
...
+                        }
+                    }
+                }
+                it("${isNotIn.name} throws AssertionError") {
+                    expect {
+                        expectSubject.isNotInFun(listOf(1, 2))
+                    }.toThrow<AssertionError> {
+                        message {
+                            contains(IS_NONE_OF.getDefault())
+                            contains("1")
```
robstoll, line 187: same as above

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
anesabml: the message doesn't contain `listBulletPoint`, here's the message `message: expected that subject: 1`

robstoll: If that is really all we got as message, then something is wrong. But I am pretty sure there is more, `is none of` should at least be there

anesabml: my bad, the issue is fixed by changing `"${listBulletPoint} 1"` to `"${listBulletPoint}1"`.

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/anyAssertions.kt
```diff
@@ -224,3 +227,35 @@ inline val <T> Expect<T>.it: Expect<T> get() : Expect<T> = this
...
+ *
+ * @return An [Expect] for the current subject of the assertion.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ * @throws IllegalArgumentException in case the iterable is empty.
+ *
+ * @since 0.13.0
+ */
+inline infix fun <reified T> Expect<T>.isNotIn(expected: IterableLike): Expect<T> {
+    val iterable = iterableLikeToIterable<T>(expected)
+    require(iterable.iterator().hasNext()) { "Iterable without elements are not allowed for this function." }
```
robstoll, line 259:
```suggestion
    require(iterable.iterator().hasNext()) { "IterableLike without elements are not allowed for this function." }
```

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/AnyAssertionsSpec.kt
```diff
@@ -81,6 +91,18 @@ class AnyAssertionsSpec : ch.tutteli.atrium.specs.integration.AnyAssertionsSpec(
 
         private fun notToBeNull(expect: Expect<Int?>, assertionCreator: Expect<Int>.() -> Unit) =
             expect notToBeNull assertionCreator
+
+        private fun getIsNoneOfInt(expect: Expect<Int>, expected: Int, otherValues: Array<out Int>): Expect<Int> =
```
robstoll, line 95: Please adjust also the other functions. Have you seen this somewhere or why the `get` prefix? (might be it is not needed at this other place either).
```suggestion
        private fun isNoneOfInt(expect: Expect<Int>, expected: Int, otherValues: Array<out Int>): Expect<Int> =
```

robstoll:
Looks good to me, going to merge later.
Already thank you for your contribution 🙂👍

anesabml:
@robstoll I am really enjoying working on this project I've learned a lot of new things, thank you for giving me this opportunity.

robstoll:
Cool win win in this case 🙂
You are more than welcome to tackle more issues if you like.

PR #566, opened 2020-09-01
Limit the text to represent calling DetailedObjectFormatter#format

Valefant:
The specific implementation of the jvm and js version seem only to be formatting classes, so I left them out.
I can change it though, just lest me know.

### DIFF
--- a/core/robstoll-lib/atrium-core-robstoll-lib-common/src/main/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/DetailedObjectFormatter.kt
+++ b/core/robstoll-lib/atrium-core-robstoll-lib-common/src/main/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/DetailedObjectFormatter.kt
@@ -41,24 +41,24 @@ abstract class DetailedObjectFormatterCommon(
      * @return The formatted [value].
      */
     @Suppress( /* TODO remove with 1.0.0 */ "DEPRECATION")
-    override fun format(value: Any?): String = limitRepresentation(when (value) {
+    override fun format(value: Any?): String = when (value) {
         null -> Text.NULL.string
         is LazyRepresentation -> format(safeEval(value))
         is Char -> "'$value'"
         is Boolean -> value.toString()
         is String -> format(value)
         is CharSequence -> format(value)
-        is Text -> value.string
-        is Translatable -> translator.translate(value)
+        is Text -> limitRepresentation(value.string)
+        is Translatable -> limitRepresentation(translator.translate(value))
         is KClass<*> -> format(value)
         is Enum<*> -> format(value)
         is Throwable -> format(value)
         //TODO remove with 1.0.0
-        is StringBasedRawString -> value.string
-        is ch.tutteli.atrium.reporting.translating.TranslatableBasedRawString -> translator.translate(value.translatable)
+        is StringBasedRawString -> limitRepresentation(value.string)
+        is ch.tutteli.atrium.reporting.translating.TranslatableBasedRawString -> limitRepresentation(translator.translate(value.translatable))
 
-        else -> value.toString() + classNameAndIdentity(value)
-    })
+        else -> limitRepresentation(value.toString()) + classNameAndIdentity(value)
+    }
 
     private fun safeEval(lazyRepresentation: LazyRepresentation) =
         //TODO remove try-catch with 1.0.0 should no longer be necessary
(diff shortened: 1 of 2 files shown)


core/robstoll-lib/atrium-core-robstoll-lib-common/src/main/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/DetailedObjectFormatter.kt
```diff
@@ -41,7 +41,7 @@ abstract class DetailedObjectFormatterCommon(
      * @return The formatted [value].
      */
     @Suppress( /* TODO remove with 1.0.0 */ "DEPRECATION")
-    override fun format(value: Any?): String = when (value) {
+    override fun format(value: Any?): String = limitRepresentation(when (value) {
```
robstoll, line 44: I would like to see a different solution where one still sees what type the subject was and what identity hash the instance had. I think it would also make sense if we use an ellipsis. Say you have a very long list then the final result should be something like:
```
expected that subject: [1, 2, 3, 4, ...       (java.util.Arrays.ArrayList <1234789>)
```
and not
```
expected that subject: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 1
```
where the `...` in the first example shows that the output would have been longer.

core/robstoll-lib/atrium-core-robstoll-lib-common/src/main/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/DetailedObjectFormatter.kt
```diff
@@ -76,6 +76,7 @@ abstract class DetailedObjectFormatterCommon(
     private fun format(throwable: Throwable) = throwable::class.fullName
 
     private fun classNameAndIdentity(any: Any): String = INDENT + "(${any::class.fullName}${identityHash(" ", any)})"
+    private fun limitRepresentation(value: String, limit: Int = 10000) = if (value.length > limit) value.substring(0, limit) else value
```
robstoll, line 79: I would not provide `limit` as parameter as we don't use it (can still be introduce if necessary). Besides, default parameters create an extra class on the JVM byte-level which is hard to deal with regarding backward compatibility (because Kotlin does not give mechanisms to chose names for instance).

robstoll (changes requested):

core/robstoll-lib/atrium-core-robstoll-lib-common/src/main/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/DetailedObjectFormatter.kt
Valefant: Thanks for the explanation. This is good to know, i'll take it out

core/robstoll-lib/atrium-core-robstoll-lib-common/src/main/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/DetailedObjectFormatter.kt
Valefant: Oh sure. This certainly sounds more useful.
I get the example, but what do you mean exactly with ellipsis?

robstoll: suffix the representation with ` ...` in case it is longer than 10'000  (I am not a native speaker but I think `...` is called ellipsis in English :laughing:)

Valefant: Me neither but it seems plausible 😂 
But yea, I can do that

Valefant: I pushed my latest progress. If I still haven't got it right, I can change it again

robstoll (approved):
@Valefant LGTM, many thanks, hat trick in only 2 days :tada:

PR #571, opened 2020-09-06
Accept a string for ChronoLocalDate

Valefant:
This Pull Request is a work in progress.

@robstoll Some tests are failing and I don't know why. 
I tried to examine it but I still have not any idea.

In `ChronoLocalDateAsStringAssertionSpec` I left the
```kotlin
include(object : ChronoLocalDateAssertionSpec(
    fun1(::isBefore),
    fun1(::isBeforeOrEqual),
    fun1(::isAfter),
    fun1(::isAfterOrEqual),
    fun1(::isEqual),
    describePrefix
) {})
```
in there. I derived it from `ChronoLocalDateTimeAsStringAssertionSpec`.
Now there are 32 failing test cases.
image

The infix notation is missing.
I will implement it shortly after the tests are working.


Valefant:
Oh I think I found it. It seems to be an quotation error within the report.
`messageContains` looks for a specific string, which does not match, because the date is surrounded by quotes.

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/chronoLocalDateAssertions.kt
```diff
@@ -68,3 +68,68 @@ fun <T : ChronoLocalDate> Expect<T>.isAfterOrEqual(expected: ChronoLocalDate): E
  */
 fun <T : ChronoLocalDate> Expect<T>.isEqual(expected: ChronoLocalDate): Expect<T> =
     _logicAppend { isEqual(expected) }
+
+/**
+ * Expects that the subject of the assertion (a [ChronoLocalDate])
+ * is before the [expected] [String].
```
robstoll, line 74:
```suggestion
 * is before the [expected] [LocalDate] given as [String].
```

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultChronoLocalDateAssertions.kt
```diff
@@ -47,4 +48,38 @@ class DefaultChronoLocalDateAssertions : ChronoLocalDateAssertions {
         expected: ChronoLocalDate
     ): Assertion = container.createDescriptiveAssertion(SAME_DAY, expected) { it.isEqual(expected) }
 
+    override fun <T : ChronoLocalDate> isBefore(
+        container: AssertionContainer<T>,
+        expected: String
+    ): Assertion = container.createDescriptiveAssertion(IS_BEFORE, expected) {
+        it.isBefore(LocalDate.parse(expected))
+    }
```
robstoll, line 56: please delegate to the method which expects a ChronoLocalDate instead of duplicating the logic
```suggestion
    ): Assertion = container.isBefore(LocalDate.parse(expected))
```

robstoll (changes requested):
@Valefant I assume you don't have the problem with the spec any more, correct?  Otherwise let me know and I take a look into it

Valefant:
> @Valefant I assume you don't have the problem with the spec any more, correct? Otherwise let me know and I take a look into it

For whatever reason, the delegation of the method fixed the quotation error. I am not sure why though 🤔

robstoll:
@valefant Note that backward-compatibility tests are failing. You can run them locally with `gr bcTests`

Valefant:
> @Valefant Note that backward-compatibility tests are failing. You can run them locally with `gr bcTests`

Indeed. The changes I made are not reflected in the bc code.
How does the bc process actually work? I don't know how to make my written code reflect the changes onto it.

robstoll:
The bc-test is taking the test of an older version of Atrium, compiles against your branch and runs them. The problem is Kotlin, it is no longer able to infer the types correctly. You have fixed this problem in your branch (see line 16 in ChronoLocalDateAssertionsSpec.kt : https://github.com/robstoll/atrium/pull/571/files#diff-1ffb3058c8fdb20ad5b0cd2a5e0b0c9cR16). This binary backward compatibility issue is OK and thus we need to help the bc-test and fix the problem in the old source. 
Have a look at /misc/tools/atrium-bc-test/build.gradle copy line 417 until 421 and adapt for ChronoLocalDateAssertionsSpec

Valefant:
> The bc-test is taking the tests of an older version of Atrium, compiles them against your branch and runs them. The problem here is actually Kotlin, the compiler is no longer able to infer the types correctly. You have fixed this problem in your branch (see line 16 in ChronoLocalDateAssertionsSpec.kt : https://github.com/robstoll/atrium/pull/571/files#diff-1ffb3058c8fdb20ad5b0cd2a5e0b0c9cR16) by providing the types. This backward compatibility issue is OK (we accept it) and thus we need to help the bc-test and fix the problem in the old source to make the test pass.
> Have a look at /misc/tools/atrium-bc-test/build.gradle copy line 417 until 421 and adapt for ChronoLocalDateAssertionsSpec

Ah I understand now. Thanks for the explanation. I adjusted it

robstoll (approved):

Issue #590, opened 2020-09-29
add Path.contains as shortcut for Path.isDirectory + Path.resolve + Path.exists


robstoll:
*Platform* (all, jvm, js, android): all
*Extension* (none, kotlin 1.3): none

## Code related feature
```
expect(path).contains("test.txt", "dummy.log")

//instead of

expect(path).isDirectory() and {
  resolve("test.txt") { exists() }
  resolve("dummy.log") { exists() }
}
```

Following the things you need to do:

*api-fluent*
- [ ] provide a fun `contains` which expects `path: String, vararg otherPaths: String` and returns `Expect<T>` (see resolve with expect an assertionCreator in pathAssertions.kt as a guideline)
- [ ] add `@since 0.14.0` (adapt to current [milestone](https://github.com/robstoll/atrium/milestones)) to KDOC
- [ ] add `contains` to PathAssertionsSpec in specs-common and extend it in atrium-api-fluent-en_GB-common/src/test

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime, we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

SylvainGirod:
Hello, can I work on this ?

robstoll:
Sure

ezesalas7:
Hello, I'll work on this...

SylvainGirod:
> Hello, I'll work on this...

I think you can if @robstoll is ok. I couldn't manage to take the time to work on this.

robstoll:
@ezesalas7 go ahead and let me know if you need help.
@SylvainGirod come back once you have time again, your help is always appreciated.

ezesalas7:
Thanks @robstoll, I am setting up the project now. First time contributing to open source, pretty exited! Hope I can get It done :)

jgrgt:
I'll work on this

robstoll:
@jgrgt there is already someone working on this. Please take another issue unless @ezesalas7 does not have time

[...]

PR #600, opened 2020-10-07
Make generateLogic support varargs + avoid index-based regex

jgrgt:
I started looking at https://github.com/robstoll/atrium/issues/590 and noticed one of the problems with fixing that is the use of vararg: generateLogic failed to work because it could not match that 'vararg'.

So I fixed that. Because I could not figure out the index-based regex, I moved everything to named groups first and then added the vararg. I don't like the way I had to abuse string interpolation to specify the named group referenced, but that's the best I could come up with.

Note that the vararg does not get 'expanded', so the underlying code needs to accept an `Array` instead of the vararg.

### DIFF
--- a/logic/generateLogic.gradle
+++ b/logic/generateLogic.gradle
@@ -123,16 +123,19 @@ def createGenerateLogicTaskForPackage(
             def identifier = /[a-zA-Z0-9]+/
             def newLine = /(?:\r\n?|\n)/
             def newLineAndIndent = /$newLine\s*/
-            def parameter = /,(?: |$newLineAndIndent)($identifier): ([^:]+?)/
-            def returnType = /(?:$newLineAndIndent)?\)(:.+)/
+            def parameter = { paramNumber -> /,(?: |$newLineAndIndent)(?<vararg${paramNumber}>vararg |)(?<paramName$paramNumber>$identifier): (?<typeName$paramNumber>[^:]+?)/}
+            def returnType = /(?:$newLineAndIndent)?\)(?<returnType>:.+)/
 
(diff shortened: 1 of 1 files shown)


robstoll (changes requested):
@jgrgt thanks for your work but I am afraid that I don't have use for the vararg part. We never use vararg on the logic level, that's also the reason why we don't support it. Have a look at #590 the task list is only for api-fluent nothing to do in the logic level.
Anyhow.. I still think that the named groups are an improvement. Thus, can you please revert the vararg part and only keep the named groups.

jgrgt:
@robstoll I removed the vararg parts but left the named groups.

logic/generateLogic.gradle
```diff
@@ -123,16 +123,19 @@ def createGenerateLogicTaskForPackage(
...
-            def patterns = (6..0).collect {
-                def steps = (0..<it * 2).findAll { it % 2 == 0 }
+            def parameter = { paramNumber -> /,(?: |$newLineAndIndent)(?<paramName$paramNumber>$identifier): (?<typeName$paramNumber>[^:]+?)/}
+            def returnType = /(?:$newLineAndIndent)?\)(?<returnType>:.+)/
+
+            def typeIdentifier = / *fun (?<generics><.+?> )?(?<funcName>$identifier)\((?:$newLineAndIndent)?$identifier: (?<type>$identifier(?:\.$identifier)*<.+>)/
+            def patterns = (6..0).collect { int amountOfParameters ->
+                def params = []
+                if (amountOfParameters > 0) {
+                    params = (1..amountOfParameters).toList()
```
robstoll, line 133: tiny detail, toList is not required
```suggestion
                    params = (1..amountOfParameters)
```

robstoll (approved):

robstoll:
@jgrgt thanks for your first contribution to Atrium :slightly_smiling_face: :+1:

PR #602, opened 2020-10-08
Fuse collection and iterable specs

tarczynskitomek:
Fuse test *FeatureAssertionSpecs for Iterable and Collection interfaces.

@robstoll Is the change size fine for you? I don't want to make gigantic commits, as they are a pain to review, neither spam you with too many PRs.


misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/CollectionAssertionsSpec.kt
```diff
@@ -27,10 +36,13 @@ abstract class CollectionAssertionsSpec(
     val isDescr = DescriptionBasic.IS.getDefault()
     val isNotDescr = DescriptionBasic.IS_NOT.getDefault()
     val empty = DescriptionCollectionAssertion.EMPTY.getDefault()
+    val fluent = expect(listOf(1, 2) as Collection<Int>)
+    val sizeDescr = DescriptionCollectionAssertion.SIZE.getDefault()
 
-    describeFun(isEmpty, isNotEmpty) {
+    describeFun(isEmpty, isNotEmpty, sizeFeature, size) {
```
robstoll, line 42: please move to own describeFun. We usually group functions which take part in all tests this is not the case here, we don't make an assertion about size in context collection is empty and vice versa for isEmpty/isNonEmpty in context list with two entries.

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/CollectionAssertionsSpec.kt
```diff
@@ -12,13 +12,22 @@ import org.spekframework.spek2.style.specification.Suite
...
         describePrefix,
         isEmpty.forSubjectLess(),
-        isNotEmpty.forSubjectLess()
+        isNotEmpty.forSubjectLess(),
+        sizeFeature.forSubjectLess().adjustName { "$it feature" },
+        size.forSubjectLess { isGreaterThan(2) }
+    ) {})
+
+    include(object : AssertionCreatorSpec<Collection<Int>>(
+        describePrefix, listOf(1),
```
robstoll, line 29: I don't know why you have changed this from String to Int but listOf(1) and toBeDescr: 1 makes it look like there is a relation. You can either keep Int and change listOf to something like listOf(999) or go back to String

robstoll (changes requested):

robstoll:
@tarczynskitomek PR size is good like this. I also prefer smaller PRs. They usually get merged quicker and thus the authors have less chances of running into a merge conflict

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/CollectionAssertionsSpec.kt
tarczynskitomek: Yes, you're right, this is indeed confusing, I'll revert it back to String.

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/CollectionAssertionsSpec.kt
tarczynskitomek: :+1: Will do - didn't think about that when fusing :(

robstoll: No problems

robstoll (approved):

PR #622, opened 2020-10-10
Fuse MapEntryAssertionSpecs

tarczynskitomek:


tarczynskitomek:
For some reason this one is more problematic :upside_down_face: A question regarding: `IterableLIkeElementComparableSpecs.scala`, as it depends on `ch.tutteli.atrium.specs.integration.IterableFeatureAssertionsSpec` (which I've only discovered the hard way), what's you suggestion @robstoll in this situation? From Kotlin code perspective the IterableFeatureAssertionsSpec class is not used, but removing it breaks the dependency in Scala.

robstoll:
@tarczynskitomek you can comment out the spec in scala. I am going to fix it later on. Feel free to do the adjustment yourself

tarczynskitomek:
> @tarczynskitomek you can comment out the spec in scala. I am going to fix it later on. Feel free to do the adjustment yourself

I've spend some time trying to figure out a way to call `hasNext` and `hasNotNext` functions defined in `iterableAssertions.kt` from Scala code, but unfortunatelly failed to get this working - a very limitted knowledge of Scala doesn't help :cry: I've ended up commenting-out the Scala spec code, as you've suggested.

Apparently, Kotlin -> Scala integration does not seems so smooth like Kotlin -> Java.

robstoll:
No worries, there are two problems with Kotlin/scala integration. Scala, like Kotlin, has own Function/Unit types. The bigger one is that Java supports only invariant type parameters and therefore Kotlin has to use wildcards all over the resulting bytecode. And only Kotlin knows that if you have twice the same `T : CharSequecence` resulting twice in `? extends CharSequence` in Java that those two wildcards are the same. Scala doesn't and so you need to use unsafe casts to work around this problem. I hope Java will support co-/contra-variant type parameters one day. Then it will be smoother.

tarczynskitomek:
> No worries, there are two problems with Kotlin/scala integration. Scala, like Kotlin, has own Function/Unit types. The bigger one is that Java supports only invariant type parameters and therefore Kotlin has to use wildcards all over the resulting bytecode. And only Kotlin knows that if you have twice the same `T : CharSequecence` resulting twice in `? extends CharSequence` in Java that those two wildcards are the same. Scala doesn't and so you need to use unsafe casts to work around this problem. I hope Java will support co-/contra-variant type parameters one day. Then it will be smoother.
> 
> I'll review later

Thanks for the explanations, :heart: My experience is mostly with combining Java - Groovy primarily when using Spock testing framework. After getting used to Groovy's quirks it's quite smooth. 

Hopefully interoperability between languages will get better with time.

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/MapEntryAssertionsSpec.kt
```diff
@@ -92,4 +103,138 @@ abstract class MapEntryAssertionsSpec(
             }
         }
     }
+
+    val keyName = "key"
```
robstoll, line 107: rename MapEntryFeatureAssertionsSpec to MapEntryAssertionsSpec and re-use here via
```
include(object: MapEntryAssertionsSpec(...){})
```
instead of duplicating

robstoll (changes requested):

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/MapEntryAssertionsSpec.kt
tarczynskitomek: I don't like the duplication too, but note that `MapEntryAssertionsSpec` already exists, so simply renaming and including this way is not possible, or most probably there's something that I don't quite get in your comment :(

robstoll: Yes, I guess a misunderstanding and somehow I copied the wrong class names :laughing: sorry about that. I meant rename KeyValueLikeFeatureAssertionSpec to KeyValueLikeAssertionSpec and include it here. The include should happen exactly here. All other changes in this PR are good.
See above in this class, we also have:
```
include(object : SubjectLessSpec<Map.Entry<String, Int>>(
        describePrefix,
        isKeyValue.forSubjectLess("key", 1)
    ) {})
```
the same can be used here. See the original MapEntryFeatureAssertionSpec it extended KeyValueLikeFeatureAssertions. You can use the same parameters, just use object:... instead of extending.
I hope my explanation is not too confusing

tarczynskitomek: Got it, no worries :-)

tarczynskitomek: One more thing - `KeyValueLikeFeatureAssertionsSpec` is deprecated, and `MapEntryFeatureAssertionSpec` extended `KeyValueLikeAssertionsSpec` already, so there's no need to rename it - otherwise everything is clear :-) Pushed new changes :-)

robstoll (approved):

Issue #482, opened 2020-04-28
accept date and time as string in ISO 8601 format for ChronoZonedDateTime


robstoll:
*Platform* (jvm, js, android): jvm
*Extension* (none, kotlin 1.3, jdk8): jdk8

## Code related feature
```
expect(ZonedDateTime.now()).isAfter("2020-04-28T12:00Z")

//instead of

expect(ZonedDateTime.now()).isAfter(ZonedDateTime.of(2020,4,28,12,0,0,0, ZoneId.of("Z")))
```

Following the things you need to do:

*atrium-logic*:
- [ ] duplicate each method in ChronoZonedDateTimeAssertions and:
  - [ ] accept a String instead of a ChronoZonedDateTime and parse it to ZonedDateTime and pass it to the existing overload where the following formats shall be supported. This is a slight deviation of ISO 8601, e.g. we do not want to support the alternative notation e.g. 20200401120001Z instead of 2020-04-01T12:00:01Z
    The format is composed of {DateTime}{ZoneDesignator}, first the spec for DateTime:
     * yyyy-mm-ddThh:mm:ss.SSS (up to 9 digits for nanoseconds)
     * yyyy-mm-ddThh:mm:ss
     * yyyy-mm-ddThh:mm
     * yyyy-mm-dd

      And for ZoneDesignator:
      * Z
      * Z+hh
      * Z+hh:mm
      * Z-hh
      * Z-hh:mm
        
      Following a few examples how it looks combined:
      - yyyy-mm-ddThh:mm:ssZ
      - yyyy-mm-ddThh:mmZ+01
      - yyyy-mm-ddZ-04:30
        
      Right now we don't intend to support things like Europe/Zurich as ZoneDesignators (would be supported by jdk). So do a check upfront if in correct format and throw a DateTimeParseException if wrong and only then use ZonedDateTime.parse
        Note that #481 deals also with parsing DateTime. You should reuse the logic from there.        
       
*api-fluent*
- [ ] duplicate each method in chronoZonedDateTimeAssertions.kt and:
  - [ ] accept a String instead of a ChronoZonedDateTime - call the corresponding logic function
  - [ ] change since to 0.14.0 (adopt to current [milestone](https://github.com/robstoll/atrium/milestones))
- [ ] add the string version to ChronoZonedDateTimeAssertionsSpec:
[...]

PR #625, opened 2020-10-11
#482 accept date as string for chronozoneddatetime

Valefant:
Let's go


Valefant:
@robstoll I have one issues that I can't explain myself.

In `DefaultChronoZonedDateTimeAssertions.kt` lines 95-97 are showing as errors.
I tested the `parseZonedDateTime` in separate scratch File and it works without problems.

image

robstoll:
Hard to guess without checking out your code. Will do that later on. Note that Intellij has a few bugs when it comes to JVM only. Does gradle output the same error? Are the functions you use all available in jdk8?

Valefant:
> Hard to guess without checking out your code. Will do that later on. Note that Intellij has a few bugs when it comes to JVM only. Does gradle output the same error? Are the functions you use all available in jdk8?

Gradle does output the same error. I checked the functions, they are all available in jdk8.
I am using the jdk8 locally too. The code is running when I run it in a scratch file 🤔 

Maybe I am missing something

robstoll:
Maybe the problem is that you used Kotlin 1.4 in your worksheet whereas we still use Kotlin 1.2 in Atrium. Kotlin 1.4 improved the support for Java's SAM-interface interoperability. You can solve it as follows:
```
        val parsed = formatter.parseBest(
            data,
            TemporalQuery { temporal -> ZonedDateTime.from(temporal) },
            TemporalQuery { temporal -> LocalDateTime.from(temporal) },
            TemporalQuery { temporal -> LocalDate.from(temporal) }
        )
```

Valefant:
@robstoll That was it, thank you 😄 

Now some tests are failing because the `DateTimeFormatter` does not recognize a date like 
`2019-12-24T10:15:30.00000021+01:00[Europe/Zurich]`.  It is not containg the `Z` and additionally the text in the brackets `[Europe/Zurich]` is not parsed. 

How should I handle this case?

Valefant:
> @robstoll That was it, thank you 😄
> 
> Now some tests are failing because the `DateTimeFormatter` does not recognize a date like
> `2019-12-24T10:15:30.00000021+01:00[Europe/Zurich]`. It is not containg the `Z` and additionally the text in the brackets `[Europe/Zurich]` is not parsed.
> 
> How should I handle this case?

Ok I tried to implement a solution. 
I am curious if it is any good

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultChronoZonedDateTimeAssertions.kt
```diff
@@ -44,4 +47,86 @@ class DefaultChronoZonedDateTimeAssertions : ChronoZonedDateTimeAssertions {
     ): Assertion = container.createDescriptiveAssertion(IS_EQUAL_TO, expected) {
         it.isEqual(expected)
     }
+
+    override fun <T : ChronoZonedDateTime<out ChronoLocalDate>> isBefore(
+        container: AssertionContainer<T>,
+        expected: String
+    ): Assertion {
+        return container.isBefore(parseZonedDateTime(expected))
+    }
```
robstoll, line 56: detail but would be nice if you could change (and the following) to a one-liner
```suggestion
    ): Assertion = container.isBefore(parseZonedDateTime(expected))
```

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultChronoZonedDateTimeAssertions.kt
```diff
@@ -44,4 +47,86 @@ class DefaultChronoZonedDateTimeAssertions : ChronoZonedDateTimeAssertions {
...
+    ): Assertion {
+        return container.isEqual(parseZonedDateTime(expected))
+    }
+
+    private fun parseZonedDateTime(data: String): ZonedDateTime {
+        /**
+         * The specification of the zone can look like this [Europe/Paris]
+         * and needs to be handled by the native jdk implementation.
+         */
+        if (data.indexOf('[') > -1 && (data.indexOf('[') < data.indexOf(']'))) {
```
robstoll, line 91: personally I would assign `data.indexOf('[') `  to a variable instead of calling indexOf twice but it's also OK this way

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultChronoZonedDateTimeAssertions.kt
```diff
@@ -44,4 +47,86 @@ class DefaultChronoZonedDateTimeAssertions : ChronoZonedDateTimeAssertions {
...
+            return ZonedDateTime.parse(data)
+        }
+
+        val formatter = DateTimeFormatterBuilder()
+            .parseCaseSensitive()
+            .append(DateTimeFormatter.ISO_LOCAL_DATE)
+            .optionalStart().appendLiteral('T').append(DateTimeFormatter.ISO_LOCAL_TIME).optionalEnd()
+            .optionalStart().appendLiteral('Z').optionalStart().appendOffsetId().optionalEnd().optionalEnd()
```
robstoll, line 99: Sorry, totally my fault, I made a mistake in the description. We want to accept either
` Z`  as time zone or `+hh:mm`  or `-hh:mm` and not `Z+hh:mm` or `Z-hh:mm`

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultChronoZonedDateTimeAssertions.kt
robstoll: I am not sure if we really should support [Europe/Paris] and the like. In the issue description I had:
``` 
Right now we don't intend to support things like Europe/Zurich as ZoneDesignators (would be supported by jdk). So do a check upfront if in correct format and throw a DateTimeParseException if wrong and only then use ZonedDateTime.parse
```
Now that you have implemented it, I would say keep it. But state a note in KDoc that its supported only experimentally and could be removed at any time.

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/chronoZonedDateTimeAssertions.kt
```diff
@@ -74,3 +74,73 @@ infix fun <T : ChronoZonedDateTime<out ChronoLocalDate>> Expect<T>.isAfterOrEqua
 infix fun <T : ChronoZonedDateTime<out ChronoLocalDate>> Expect<T>.isEqual(
     expected: ChronoZonedDateTime<*>
 ): Expect<T> = _logicAppend { isEqual(expected) }
+
+/**
+ * Expects that the subject of the assertion (a [ChronoZonedDateTime])
+ * is before or equals the [expected] [ChronoZonedDateTime] given as [String].
+ * Check [ch.tutteli.atrium.logic.impl.DefaultChronoZonedDateTimeAssertions.parseZonedDateTime] to see which values are allowed for [expected].
```
robstoll, line 81: I think we should write the supported formats here. I doubt someone wants to go and have a look at an internal method.
Also, we can then write that ZoneDesignator like [Europe/Zurich] is only supported experimentally.

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/ChronoZonedDateTimeAsStringAssertionSpec.kt
```diff
@@ -0,0 +1,212 @@
...
+                }
+            }
+        }
+    }
+
+    describe("allowed shortcuts") {
+        mapOf(
+            "2020-01-02T03:04:05Z" to ZonedDateTime.of(2020, 1, 2, 3, 4, 5, 0, ZoneOffset.UTC) as ChronoZonedDateTime<*>,
+            "2020-01-02T03:04Z" to ZonedDateTime.of(2020, 1, 2, 3, 4, 0, 0, ZoneOffset.UTC),
+            "2020-01-02Z" to ZonedDateTime.of(2020, 1, 2, 0, 0, 0, 0, ZoneOffset.UTC)
```
robstoll, line 130: I have not thought about this shortcut but I actually like it :+1: 
Can you also add "2020-01-02+01:30" which should result in `2020-01-02T00:00:00+01:30`

robstoll (changes requested):

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultChronoZonedDateTimeAssertions.kt
Valefant: Ah ok. Yea sure, I'll adjust it

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultChronoZonedDateTimeAssertions.kt
Valefant: Alright, I am dropping it for now.

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/ChronoZonedDateTimeAsStringAssertionSpec.kt
Valefant: Ok, I dropped the named zones. 
I think I implemented all your suggestions. 
Tell me what you think of it now 😄

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/chronoZonedDateTimeAssertions.kt
robstoll: Nice work in api-fluent, can you please copy it to api-infix as well

robstoll (changes requested):
One last detail and we are good to merge

Issue #473, opened 2020-04-25
update js samples to use the new org.jetbrains.kotlin.js gradle plugin


robstoll:
*Platform* (jvm, js, android): js
*Extension* (none, kotlin 1.3, jdk8): none

## Non-Code related feature
The current samples for mocha and jasmine had been based on kotlin-examples which used the old kotlin plugin. By now, there is a new kotlin plugin (kotlin-examples is not updated though) but we could try to use the new Plugin instead. 

The new plugin claims to have a few advantages
> The plugin also provides useful testing features, for example:
> -    Source maps generation
> -    Test reports generation
> -    Test run results in the console

So that sounds good. I guess it should be easy for mocha as the default runner they use is mocha but it might actually be hard for jasmine. Therefore I suggest to start with mocha only (first PR) and try afterwards if there is a way to reconfigure the plugin to use jasmine instead (second PR).

Following the description by JetBrains:
https://kotlinlang.org/docs/reference/js-project-setup.html

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime, we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

developerfred:
I'll work on this

robstoll:
@developerfred nice :slightly_smiling_face: 
I just pushed some cleanups https://github.com/robstoll/atrium/pull/474, please fork again or rebase on master once it is merged

developerfred:
thanks @robstoll.

robstoll:
@developerfred un-assigning so that pthers can take over. Let me know in case you still work on it.

pablohildo:
I'll work on this.

robstoll:
@pablohildo unassigning you so that others could take over

chadmarchand:
I'll work on this

Issue #631, opened 2020-10-14
Atrium stops working for JS with Kotlin 1.4


robstoll:
*Affected Version*: almost all, as soon as Kotlin 1.4 is used by the consuming project
*API* (fluent-en_GB, infix-en_GB): both
*Platform* (jvm, js): js
*Extension* (none, kotlin 1.3): none

## How to reproduce the problem  
See https://github.com/robstoll/atrium/pull/620

And https://youtrack.jetbrains.com/issue/KT-42610
## Describe the bug
CoroutineBasedLocale... is using coroutines from the experimental package. This one was dropped with Kotlin 1.4

## Expected behaviour
Atrium should also work with Kotlin 1.4

Rewrite CoroutineBased... Simplest is to pre-compute everything by putting the Locale into a List and turn the list into a Sequence

chadmarchand:
I'll work on this

chadmarchand:
@robstoll Is there a trick to running the spec tests? I'm trying to run LocaleOrderDeciderSpec, have tried a couple different methods to no avail (gradle tasks and the Spek plugin for IntelliJ). The Spek plugin gives me this error when I try to run with Run -> Speks in package ch.tutteli... (Or Run -> tests in package ch.tuttelli...)

`/media/hdd/projects/atrium/core/api/atrium-core-api-jvm/src/module/module-info.java
java: modules are not supported in -source 1.8
  (use -source 9 or higher to enable modules)`

If I understand correctly the java versions should come from the gradle file so I shouldn't have to change any IntelliJ project settings, but I've tried anyways to change to Java 9 and still have the same issue.

robstoll:
I am using 
IntelliJ IDEA 2020.2.3 (Community Edition)
Build #IC-202.7660.26, built on October 6, 2020
Runtime version: 11.0.8+10-b944.34 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.

With the Spek Framework plugin 2.0.13-IJ2020.2

And I can just click on the play gutter in core/robstoll-lib/atrium-core-robstoll-lib-common/src/test/kotlin/ch/tutteli/atrium/core/robstoll/lib/reporting/translating/LocaleOrderDeciderSpec.kt and the test is executed as it should (and is green)

However, what I cannot do is clicking on a package and run all tests -> it seems like in this case spec is not used (most likely kotlin.test kicks in)

[...]

PR #634, opened 2020-10-17
#473 use new Kotlin JS plugin in Mocha sample

chadmarchand:

### DIFF
--- a/samples/js/mocha/build.gradle
+++ b/samples/js/mocha/build.gradle
@@ -61,8 +61,8 @@ task establishDependencyToAtrium(dependsOn: [compileTestKotlinJs]) {
         srcFile.write(newContent, 'UTF-8')
     }
 }
-task runMocha(type: NpxTask, group: 'Verification', dependsOn: [compileTestKotlinJs, establishDependencyToAtrium]) {
-    command = "mocha"
+task runMocha(type: NodeTask, group: 'Verification', dependsOn: [compileTestKotlinJs, establishDependencyToAtrium]) {
+    script = file("$buildDir/js/node_modules/mocha/bin/mocha")
     args = [compileTestKotlinJs.outputFile, '--full-trace']
 }
 check.dependsOn runMocha
--- a/samples/js/mocha/build.gradle
+++ b/samples/js/mocha/build.gradle
@@ -36,14 +36,18 @@ dependencies {
     testImplementation "org.jetbrains.kotlin:kotlin-test-js"
 }
 
+kotlin {
+    target {
+        nodejs()
+    }
+}
+
 // we only configure compileTestKotlinJs and not compileKotlinJs as well
 // so that you can also use UMD as moduleKind for compileKotlinJs
 compileTestKotlinJs.configure {
     // mocha requires commonjs
     kotlinOptions.moduleKind = "commonjs"
 }
-
-kotlin.target.nodejs { }
 
 //either you use this task or create a testSetup.kt in your src/test with the following content
 //
--- a/samples/js/mocha/build.gradle
+++ b/samples/js/mocha/build.gradle
@@ -43,32 +43,7 @@ compileTestKotlinJs.configure {
     kotlinOptions.moduleKind = "commonjs"
 }
 
-//gather all required dependencies and put them into the node_modules folder of this (sub)-project
-task populateNodeModules(type: Copy, dependsOn: compileKotlinJs) {
-    from compileKotlinJs.destinationDir
-
-    afterEvaluate {
-        // in contrast to kotlin examples we use testRuntimeClasspath this way we also pick up runtimeOnly dependencies
-        configurations.testRuntimeClasspath.each {
-            if (it.name.endsWith("jar")) {
-                from zipTree(it.absolutePath).matching { include '*.js', '*.js.map' }
-            }
-        }
-    }
-
-    into "$buildDir/node_modules"
-}
-
-//node, npm and dependencies of mocha is installed in the same folder for all sub-projects
-def nodeModulesParentDir = file("$rootProject.projectDir/gradle/")
-node {
-    //download this way we don't have to download each time
-    download = true
-    npmVersion = '6.4.1'
-    workDir = file("$rootProject.projectDir/.gradle/nodejs")
-    npmWorkDir = file("$rootProject.projectDir/.gradle/npm")
-    nodeModulesDir = nodeModulesParentDir
-}
+kotlin.target.nodejs { }
 
 //either you use this task or create a testSetup.kt in your src/test with the following content
 //
@@ -79,14 +54,14 @@ node {
 // private val establishDependencyToAtrium = dependOnAtrium()
 task establishDependencyToAtrium(dependsOn: [compileTestKotlinJs]) {
     doLast {
-        def srcFile = file("$buildDir/classes/kotlin/test/${project.name}_test.js")
+        def srcFile = file("$buildDir/js/packages/${project.name}-test/kotlin/${project.name}-test.js")
         String newContent = srcFile.getText('UTF-8')
             .replace("(function (_, Kotlin,", "(function (_, Kotlin, \$module\$${atrium_api.replace('-', '_')},")
             .replace("(module.exports, require('kotlin'),", "(module.exports, require('kotlin'), require('$atrium_api'),")
         srcFile.write(newContent, 'UTF-8')
     }
 }
-task runMocha(type: NpxTask, group: 'Verification', dependsOn: [compileTestKotlinJs, populateNodeModules, establishDependencyToAtrium]) {
+task runMocha(type: NpxTask, group: 'Verification', dependsOn: [compileTestKotlinJs, establishDependencyToAtrium]) {
     command = "mocha"
     args = [compileTestKotlinJs.outputFile, '--full-trace']
 }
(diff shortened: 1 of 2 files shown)


samples/js/mocha/build.gradle
```diff
@@ -36,9 +36,9 @@ dependencies {
...
 }
 
-// we only configure compileTestKotlin2Js and not compileKotlin2Js as well
-// so that you can also use UMD as moduleKind for compileKotlin2Js
-compileTestKotlin2Js.configure {
+// we only configure compileTestKotlinJs and not compileKotlinJs as well
+// so that you can also use UMD as moduleKind for compileKotlinJs
+compileTestKotlinJs.configure {
     // mocha requires commonjs
     kotlinOptions.moduleKind = "commonjs"
```
robstoll, line 43: Do you know if we still need line 49 and 50 with the new plugin or is this now on per default?

samples/js/mocha/build.gradle
```diff
@@ -51,13 +51,15 @@ compileTestKotlin2Js.configure {
...
-task populateNodeModules(type: Copy, dependsOn: compileKotlin2Js) {
-    from compileKotlin2Js.destinationDir
+task populateNodeModules(type: Copy, dependsOn: compileKotlinJs) {
+    from compileKotlinJs.destinationDir
 
     afterEvaluate {
         // in contrast to kotlin examples we use testRuntimeClasspath this way we also pick up runtimeOnly dependencies
         configurations.testRuntimeClasspath.each {
-            from zipTree(it.absolutePath).matching { include '*.js', '*.js.map' }
+            if (it.name.endsWith("jar")) {
```
robstoll, line 60: I really hoped that this hack was not required with the new plugin any more. Are you sure this is still required? Would you mind to try out if it works also without populateNodeModules?

samples/js/mocha/build.gradle
chadmarchand: You are correct that the source maps are generated by default now, I will remove these lines.

samples/js/mocha/build.gradle
chadmarchand: I tried without populateNodeModules and it did not work. No node_modules appear in the build directory, and it fails with the error "Cannot find module 'kotlin'" during the runMocha task. I messed around with other options but wasn't able to get it working. It may be possible once kotlin-gradle-plugin is updated from 1.3.71 as it seems like there are lots of changes there.

robstoll: that's a pity. Maybe the problem stems from the fact that we configure node, might be the new plugin (v. 1.3.71) does this on its own?

chadmarchand: I was able to get it working by removing this node configuration and making some other small adjustments, however there is one complication. The new plugin also generates a yarn.lock file, and the fact that the project is named "mocha" seems to interfere with this. Unless the project name is changed in settings.gradle to something else, it seems like there is a conflict with the actual mocha library or something that causes the yarn.lock to not be created properly, and the following error: "
Execution failed for task ':kotlinNpmInstall'.
> Cannot find mocha@6.2.2 in yarn.lock
"
I tried changing the project name to "mymocha" just to see and there is no such issue, all tests run successfully.

robstoll: Nice 🎉
then lets rename the project in settings to atrium-with-mocha

robstoll:
@chadmarchand hard to review as you seem to have cherry-picked changes from master. Can you please rebase on master instead and force push. I should only see your changes under files changed (https://github.com/robstoll/atrium/pull/634/files)

robstoll:
@chadmarchand sorry, somehow I did not get a notification that you rebased. It looks good now, the only problem is that the build fails:
https://travis-ci.com/github/robstoll/atrium/jobs/402559249#L573

I'll also take a look if I can figure out something. Apparently it works on Windows but fails on linux :thinking:

chadmarchand:
> @chadmarchand sorry, somehow I did not get a notification that you rebased. It looks good now, the only problem is that the build fails:
> https://travis-ci.com/github/robstoll/atrium/jobs/402559249#L573
> 
> I'll also take a look if I can figure out something. Apparently it works on Windows but fails on linux 
> 
> * works on ubuntu, must be specific to travis-ci's trusty image

I can't seem to reproduce the failure (also on ubuntu), but the log shows something about an unexpected reserved keyword. I temporarily added a stacktrace to try to see what is happening in the next build log as I'm not sure what would cause that error to occur.

chadmarchand:
I was unable to learn much because npx strips the error/stack information and just logs the error message ("Unexpected reserved word"). I will try more ideas later.

chadmarchand:
The latest commit seems to have fixed the Travis build (I think before, the npx command may have been using the wrong bin folder). It seems like the java-windows github action is still failing though, seemingly due to a weird yarn error:


```
2020-10-24T16:22:27.3184833Z warning workspace-aggregator-d140276a-961a-4703-a522-38a2972a65db > multiplatform-test > mocha > mkdirp@0.5.3: Legacy versions of mkdirp are no longer supported. Please update to mkdirp 1.x. (Note that the API surface has changed to use Promises in 1.x.)
2020-10-24T16:22:30.6993490Z error An unexpected error occurred: "ENOENT: no such file or directory, open 'C:\\Users\\runneradmin\\AppData\\Local\\Yarn\\Cache\\v6\\npm-es-abstract-1.18.0-next.1-6e3a0a4bda717e5023ab3b8e90bec36108d22c68-integrity\\node_modules\\es-abstract\\.yarn-metadata.json'".
```

And in the previous run, a similar but different error

```
2020-10-23T00:56:00.2529045Z error An unexpected error occurred: "EPERM: operation not permitted, unlink 'C:\\Users\\runneradmin\\AppData\\Local\\Yarn\\Cache\\v6\\npm-lodash-4.17.20-b44a9b6297bcb698f1c51a3545a2b3b368d59c52-integrity\\node_modules\\lodash\\.yarn-tarball.tgz'".
2020-10-23T00:56:05.5851117Z error https://registry.yarnpkg.com/es-abstract/-/es-abstract-1.17.7.tgz: Extracting tar content of undefined failed, the file appears to be corrupt: "ENOENT: no such file or directory, stat 'C:\\Users\\runneradmin\\AppData\\Local\\Yarn\\Cache\\v6\\npm-es-abstract-1.17.7-a4de61b2f66989fc7421676c1cb9787573ace54c-integrity\\node_modules\\es-abstract\\helpers\\getProto.js'"
```

There seems to be something flaky occuring

robstoll:
The strangest thing is, your changes should not affect the windows build. Only the samples build should be affected and this one is green. Seems like a github action bug to me that the different builds are not isolated enough. You could try it out in a different branch where you don't change the mocha sample but just add an additional project based on the new kotlin plugin, include it into github actions and travis and see the result (I have the gut feeling that the problem in travis is related to Atrium).

robstoll:
I forgot that I have setup a composite build - atrium-samples-test which is run in the windows build.

chadmarchand:
I created a new branch with just the default new kotlinJS project for the org.jetbrains.kotlin.js plugin, and included it in the github action and the travis build, and it still failed: https://github.com/robstoll/atrium/runs/1303691204?check_suite_focus=true

I think it may be related to this issue in the plugin: https://youtrack.jetbrains.com/issue/KT-37258 Though we don't run in continuous mode, the symptom looks similar. According to the changelog it was fixed in 1.4. Looking at the source code, it makes sense because they add the --network-concurrency option to the yarn command used in kotlinNpmInstall, which I think should increase the reliability.

robstoll:
Brilliant investigation. In this case we need to update the sample to kotlin 1.4 first. However, this fails due to #631 
Do you fancy to provide the fix?

robstoll (approved):

robstoll:
Approved this PR so that it counts towards your hacktoberfest PRs (in case you take part). Obviously I am not going to merge it until we resolve the problem. Many thanks already for your work 👍

chadmarchand:
> Approved this PR so that it counts towards your hacktoberfest PRs (in case you take part). Obviously I am not going to merge it until we resolve the problem. Many thanks already for your work 👍

No problem, thanks for sharing this interesting project! I will take a look at #631.

robstoll:
@chadmarchand both sample projects, mocha and jasmine use now Kotlin 1.4.10 as well. Which means you should no longer be blocked here.

chadmarchand:
Thanks, I should have some time to update this PR later this week.

chadmarchand:
It seems to be failing due to the "Cannot read property 'intrinsics' of undefined" error - strange, because you can also see this issue in e.g. the build to bump kotlin_version up in the mocha sample project, but there it does not cause the build to fail: https://github.com/robstoll/atrium/runs/1331976718

robstoll:
a few questions and idea:
- was it also like this the last time that the jasmine sample failed and not the mocha sample (strange the jasmine is even carried out earlier than the mocha)?
- can you try it out with the npx instead of node, maybe using both mix somehow things up?

chadmarchand:
It was the jasmine sample that failed as well previously. I think you may be correct with the npx vs. npm command issue. It is still failing currently because I think I need to play with the options a bit, but I think the issue seems to be that the jasmine tests run using a different version of node than before.

Previously, the error appeared in the log as a "UnhandledPromiseRejectionWarning" along with the warning: "DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code." Which seems to be exactly what is happening now. I think the new plugin introduces a different node version and it then gets used by the other samples. I will try to think of some additional changes to account for this.

robstoll:
So maybe it would be easiest to change the jasmine sample also in this PR? Then we don't have the situation that different kotlin gradle plugins are used in the same github action-build

robstoll:
I found the problem, going to fix it. Thus merging this into another branch than master.

PR #638, opened 2020-10-18
Refactor iteration

mohammaddvi:

### DIFF
--- a/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
+++ b/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
@@ -167,7 +167,9 @@ class DefaultIterableLikeAssertions : IterableLikeAssertions {
         val list = transformToList(container, converter)
 
         val duplicates = createIndexAssertions(list) { (_, element) ->
-            list.count { e -> e == element } > 1
+            val first = list.indexOf(element)
+            val second = list.lastIndexOf(element)
+            first != second
         }
         createHasElementPlusFixedClaimGroup(
             list,


logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
```diff
@@ -167,7 +167,9 @@ class DefaultIterableLikeAssertions : IterableLikeAssertions {
         val list = transformToList(container, converter)
 
         val duplicates = createIndexAssertions(list) { (_, element) ->
-            list.count { e -> e == element } > 1
+            val first = list.indexOf(element)
```
robstoll, line 170: I see your intend, good idea. You could skip the list.indexOf(element) because we already know the index of the element (rename `_`  to `first`). However, in the worst case (there is no duplicate) we still go through the list once for every element.

To get to a very good result, you need to inline transformToList and createIndexAssertions here. IMO there should be a solution where you go through the list exactly once (with the help of another data structure which keeps track of what you have already seen for instance -- maybe there are other solutions).

In order that we can reduce it

mohammaddvi: Maybe we need a data structure that its order of search be O(1) like hashMap.because the order of searching in the list is O(n). I`ll work more on it...

robstoll: Sounds like a plan, just make sure you don't lose the duplicates.

mohammaddvi: I want to share some knowledge with you to find out an appropriate solution to our problem. I think we can`t find a way to find duplication of a specific element in O(1) with a list, we also can`t using HashSet because as you told before we don`t want to lose the duplication. Using hashMap needs key and values of all data and create this hashmap needs to iterate once so no good point in here unless we refactor where we are put data in the list. we can just find the element in O(1) by using HashMap that means we need another lookup to find there is duplication for this element or not.

mohammaddvi: on the other hand, if we want to use HashMap we need index as I saw in transformToList our data is generic E.right?

robstoll: I think we should be able to achieve O(n) for this problem. For this you need to inline transformToList and createIndexAssertions. What I mean by that is that you copy the content of those methods into the containsNoDuplicate methods. This way you have full control of how you track the duplicates. Personally, I would iterate the iterable and create two data structures, one for the duplicates which we are going to present in the end and one which keeps track of an already seen element (including index). Let me know if this hint already suffices to proceed, otherwise I can give further hints.

robstoll: @mohammaddvi let me know in case you need more help

robstoll:
@mohammaddvi are you still working on this one?

robstoll:
closing, I doubt you are still working on it, otherwise let me know it

PR #639, opened 2020-10-18
Update Gradle wrapper to 6.7

fejd:
Update the Gradle wrapper used by the root and sample
projects from 6.3 to 6.7.

### DIFF
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/pathAssertions.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/pathAssertions.kt
@@ -287,6 +287,18 @@ fun <T : Path> Expect<T>.isDirectory(): Expect<T> =
     _logicAppend { isDirectory() }
 
 /**
+ * Expects that the subject of the assertion (a [Path]) is an absolute path;
+ * meaning that the [Path] specified in this instance starts at the file system root.
+ *
+ * @return An [Expect] for the current subject of the assertion.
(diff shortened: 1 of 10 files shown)


robstoll:
@fejd I guess you have updated the number manually. Use the utility from gradle itself to update the wrapper. At least the jar should change as well.

fejd:
Ah sorry, yes that's how I've done in the past. I'll use `./gradlew wrapper --gradle-version 6.7` instead, and map the distribution type to what's currently in the individual properties files. Not part of this commit but should all projects use the same distribution type, either -bin or -all? I also noticed that the update utility prefers Windows style carriage returns (`^M`) in the .bat files, but the existing files use Linux/Unix line endings. I'll fix that with `dos2unix` to make the change easier to review, but I could make a subsequent commit to fix that if you want. It'd make future upgrades a bit smoother I think.

robstoll:
> but the existing files use Linux/Unix line endings.

maybe worth adding a .gitattributes file declaring clrf for bat files?

Fine for me if you do everything in one commit but two commits are also good. What you think is better :)

atrium-scala/gradlew.bat
```diff
@@ -0,0 +1,89 @@
+@rem
```
robstoll, line 1: I'm sorry, this just changed in master (atrium-scala was renamed to atrium-scala2).
Please rebase your changes on master and re-do the update in atrium-scala2

btw. nice that you have spotted this, forgot to include it in the description :+1:

robstoll (changes requested):

samples/maven/.gitattributes
```diff
@@ -1,6 +0,0 @@
-#
```
robstoll, line 1: Can you please keep these files in the samples and copy it to the root.
The samples are independent projects where the idea is that you as a user can copy this as a starting point of a new project.

Issue #660, opened 2020-10-20
add Expect.because to document the reason for an assertion


robstoll:
*Platform* (all, jvm, js, android): all
*Extension* (none, kotlin 1.3): none

## Code related feature
See robstoll/atrium-roadmap#15 for the full background. 

We want to be able to state the reason for an assertion (could also be additional context to explain the assertion etc.).

```
expect(proposeAFileName())
  .because("? is not allowed in file names on Windows") {
    containsNot("?")
  }
```

And it should look like the following in reporting:
```
expect: "file?Name"     (String <11>)
◆ contains not: "?"    (String <12>) 
ℹ because: ? is not allowed in file names on Windows
```

Following the things you need to do:

*core*
- [x] introduce a new ExplanatoryAssertionGroupType named InformationAssertionGroupType
- [x] add withInformationType to ExplanatoryGroup.GroupTypeOption

*core-robstoll-lib*
- [x] add `informationBulletPoint`  to TextExplanatoryAssertionGroupFormatter, use "ℹ ", and adjust formatGroupHeaderAndGetChildParameterObject accordingly

*specs*
- [x] adjust AsciiBulletPointReporterFactory, also add InformationAssertionGroupType and use `(i)` as ascii bullet point 
- [x] add InformationAssertionGroupType to AssertionFormatterControllerSpec (see WarningAssertionGroupType)
- [x] copy TextWarningAssertionGroupFormatterSpec and replace Warning with Information
  - [x] extend it in core-robstoll-lib

*logic*
- [x] extend AnyAssertions with a function `because` which takes a string and an assertionCreator-lambda (see other assertion functions in anyAssertions which take an assertionCreator-lambda)
- [x] implement `because` in DefaultAnyAssertions.kt by using the assertionBuilder and create an invisibleAssertionGroup where it consists of two assertions, first the collected assertions and second an ExplanatoryAssertionGroup with your new type containing only an ExplanatoryAssertion with the corresponding text, prefixed with `"because "` where because should be translated (use "weil " for german) -> i.e. u need to use a TranslatableWithArgs (search for existing code)
- [x] run `./gradlew generateLogic` this will re-generate any.kt under src/generated

*api-fluent*
- [x] provide fun `because` in anyAssertions.kt and delegate to logic (see other functions in anyAssertion)
- [x] add `@since 0.14.0` (adapt to current [milestone](https://github.com/robstoll/atrium/milestones)) to KDOC
[...]

PR #666, opened 2020-10-24
#660 add except because

Valefant:
Let's go 

### DIFF
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
@@ -42,6 +42,19 @@ fun <T> Expect<T>.isSameAs(expected: T): Expect<T> = _logicAppend { isSameAs(exp
 fun <T> Expect<T>.isNotSameAs(expected: T): Expect<T> = _logicAppend { isNotSameAs(expected) }
 
 /**
+ * Allows to state a reason for one or multiple assertions for a given subject.
+ * @param reason The explanation for the assertion/s.
+ * @param assertionCreator The group of assertions to make.
+ *
+ * @return An [Expect] for the current subject of the assertion.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ *
+ * @since 0.14.0
+ */
+fun <T> Expect<T>.because(reason: String, assertionCreator: Expect<T>.() -> Unit): Expect<T> =
+    _logicAppend { because(reason, assertionCreator) }
+
+/**
  * Expects that the subject of the assertion is either `null` in case [assertionCreatorOrNull]
  * is `null` or is not `null` and holds all assertions [assertionCreatorOrNull] creates.
  *
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
@@ -43,7 +43,7 @@ fun <T> Expect<T>.isNotSameAs(expected: T): Expect<T> = _logicAppend { isNotSame
 
 /**
  * Allows to state a reason for one or multiple assertions for the current subject.
- * @param reason The explanation for the assertion/s.
+ * @param reason The explanation for the assertion(s) created by [assertionCreator].
  * @param assertionCreator The group of assertions to make.
  *
  * @return An [Expect] for the current subject of the assertion.
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
@@ -42,7 +42,7 @@ fun <T> Expect<T>.isSameAs(expected: T): Expect<T> = _logicAppend { isSameAs(exp
 fun <T> Expect<T>.isNotSameAs(expected: T): Expect<T> = _logicAppend { isNotSameAs(expected) }
 
 /**
- * Allows to state a reason for one or multiple assertions for a given subject.
+ * Allows to state a reason for one or multiple assertions for the current subject.
  * @param reason The explanation for the assertion/s.
  * @param assertionCreator The group of assertions to make.
  *
@@ -50,6 +50,8 @@ fun <T> Expect<T>.isNotSameAs(expected: T): Expect<T> = _logicAppend { isNotSame
  * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
  *
  * @since 0.14.0
+ *
+ * @sample ch.tutteli.atrium.api.fluent.en_GB.samples.AnyAssertionSamples.because
  */
 fun <T> Expect<T>.because(reason: String, assertionCreator: Expect<T>.() -> Unit): Expect<T> =
     _logicAppend { because(reason, assertionCreator) }
--- a/apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/anyAssertions.kt
+++ b/apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/anyAssertions.kt
@@ -1,5 +1,6 @@
 package ch.tutteli.atrium.api.infix.en_GB
 
+import ch.tutteli.atrium.api.infix.en_GB.creating.KeyWithCreator
 import ch.tutteli.atrium.api.infix.en_GB.creating.Values
 import ch.tutteli.atrium.creating.Expect
 import ch.tutteli.atrium.logic.*
@@ -40,6 +41,24 @@ infix fun <T> Expect<T>.isSameAs(expected: T): Expect<T> = _logicAppend { isSame
  * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
  */
 infix fun <T> Expect<T>.isNotSameAs(expected: T): Expect<T> = _logicAppend { isNotSameAs(expected) }
+
+/**
+ * Allows to state a reason for one or multiple assertions for a given subject.
+ * @param keyWithCreator Combines the two parameters which are needed for [ch.tutteli.atrium.logic.because]
+ *
+ * @return An [Expect] for the current subject of the assertion.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ *
+ * @since 0.14.0
+ */
+infix fun <T> Expect<T>.because(keyWithCreator: KeyWithCreator<String, T>): Expect<T> =
+    _logicAppend { because(keyWithCreator.key, keyWithCreator.assertionCreator) }
+
+/**
+ * Helper function to create an [KeyWithCreator] based on the given [reason] and [assertionCreator].
+ */
+fun <T> of(reason: String, assertionCreator: Expect<T>.() -> Unit): KeyWithCreator<String, T> =
+    KeyWithCreator(reason, assertionCreator)
 
 /**
  * Expects that the subject of the assertion is either `null` in case [assertionCreatorOrNull]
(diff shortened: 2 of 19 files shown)


Valefant:
@robstoll Instead of creating a new `OfWithCreator` i used the existing `KeyWithCreator` for the infix api. I change change it, if you want to :-)

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
```diff
@@ -40,6 +40,7 @@ abstract class AnyAssertionsSpec(
     isNotInDataClass: Fun1<DataClass, Iterable<DataClass>>,
     isNotInNullableInt: Fun1<Int?, Iterable<Int?>>,
     isNotInNullableDataClass: Fun1<DataClass?, Iterable<DataClass?>>,
+    because: Fun2<String, String, Expect<String>.() -> Unit>,
```
Valefant, line 43: This is unused momentarily, therefore one build failed.
I am not sure why we need it and how to use it exactly in the spec context.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
```diff
@@ -41,6 +41,19 @@ fun <T> Expect<T>.isSameAs(expected: T): Expect<T> = _logicAppend { isSameAs(exp
  */
 fun <T> Expect<T>.isNotSameAs(expected: T): Expect<T> = _logicAppend { isNotSameAs(expected) }
 
+/**
+ * Allows to state a reason for one or multiple assertions for a given subject.
```
jGleitz, line 45: The reason will be for the *current* subject, not for some provided one

```suggestion
 * Allows to state a reason for one or multiple assertions for the current subject.
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
```diff
@@ -41,6 +41,19 @@ fun <T> Expect<T>.isSameAs(expected: T): Expect<T> = _logicAppend { isSameAs(exp
...
+ * @param reason The explanation for the assertion/s.
```
jGleitz, line 46: I’d add that we are talking about the assertions created by `assertionCreator`:

```suggestion
 * @param reason The explanation for the assertion(s) created by [assertionCreator].
```

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/anyAssertions.kt
```diff
@@ -41,6 +42,24 @@ infix fun <T> Expect<T>.isSameAs(expected: T): Expect<T> = _logicAppend { isSame
  */
 infix fun <T> Expect<T>.isNotSameAs(expected: T): Expect<T> = _logicAppend { isNotSameAs(expected) }
 
+/**
+ * Allows to state a reason for one or multiple assertions for a given subject.
```
jGleitz, line 46:
```suggestion
 * Allows to state a reason for one or multiple assertions for the current subject.
```

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
```diff
@@ -746,6 +745,36 @@ abstract class AnyAssertionsSpec(
...
+        }
+
+        it("provoke the failing of one assertion") {
+            expect {
+                expect("filename?")
+                    .because("? is not allowed in file names on Windows") {
+                        containsNot("?")
+                        startsWith("f")
+                    }
+            }.toThrow<AssertionError>()
```
jGleitz, line 763: This test case is currently useless: It only tests that `containsNot` is working correctly. To make it meaningful, we should test that the thrown `AssertionError` contains the explanation, formatted exactly like we want it to be formatted.

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
```diff
@@ -746,6 +745,36 @@ abstract class AnyAssertionsSpec(
...
+        }
+
+        it("provoke the failing of two assertions") {
+            expect {
+                expect(21)
+                    .because("we use the definition that teens are between 12 and 18 years old") {
+                        isGreaterThanOrEqual(12)
+                        isLessThan(18)
+                        isNoneOf(21)
+                    }
```
jGleitz, line 773: this test case currently only tests that `isLessThan` works correctly (see above).

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
jGleitz: This parameter is the function that should be tested. Your test currently tests the `because` implementation from the `fluent` API. However, this test class is meant to be generic, meaning that it can test both the `infix` and the `fluent` API. 

The test cases you write should never call `because` from the `fluent` API. Instead, they should use this parameter. Atrium has some helpers to achieve this. If you look at the other test cases, you can find out how to use them.

If you need more help, do not hesitate to ask!

jGleitz:
I see a design issue about group type naming that we should discuss (cc @robstoll):

This PR introduces `InformationAssertionGroupType : ExplanatoryAssertionGroupType`. As I have stated in the past, I currently find it _very_ difficult to understand the different group types atrium has on the `logic` level.

Can we choose better names? For example, I think “explanatory” is a more specific term than “information”, however, the inheritance relations points in the exact opposite way.

 1. To my mind, `ExplanatoryAssertionGroupType` is the supertype for everything that provides further information about the assertions. So maybe `DetailAssertionGroupType` would be an apt name?

 2. The new assertion group type we are adding is about explaining why the assertions were created. Since `ExplanatoryAssertionGroupType` has already been used, I’d like to try and find a new, more descriptive name for it. How about `ReasonAssertionGroupType`?

3. I think we need an additional type for debug information (something like “DebugInformationAssertionGroupType” or “HelpAssertionGroupType”. But that is a separate discussion ([see atrium-roadmap#9](https://github.com/robstoll/atrium-roadmap/issues/9)).

robstoll:
@jGleitz thanks for your feedback. Could you please create an issue in atrium-roadmap for your points and discuss it there? I think a PR is the wrong place to take this further

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/anyAssertions.kt
```diff
@@ -41,6 +42,24 @@ infix fun <T> Expect<T>.isSameAs(expected: T): Expect<T> = _logicAppend { isSame
...
+ * @return An [Expect] for the current subject of the assertion.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ *
+ * @since 0.14.0
+ */
+infix fun <T> Expect<T>.because(keyWithCreator: KeyWithCreator<String, T>): Expect<T> =
+    _logicAppend { because(keyWithCreator.key, keyWithCreator.assertionCreator) }
+
+/**
+ * Helper function to create an [KeyWithCreator] based on the given [reason] and [assertionCreator].
```
robstoll, line 58:
```suggestion
 * Helper function to create a [KeyWithCreator] based on the given [reason] and [assertionCreator].
```

core/api/atrium-core-api-common/src/main/kotlin/ch/tutteli/atrium/assertions/builders/explanatoryGroup.kt
```diff
@@ -27,6 +27,11 @@ interface ExplanatoryGroup {
          */
         override val withWarningType: AssertionsOption<WarningAssertionGroupType, FinalStep>
 
+        /**
+         * Builder to create an [AssertionGroup] with a [InformationAssertionGroupType].
```
robstoll, line 31:
```suggestion
         * Builder to create an [AssertionGroup] with an [InformationAssertionGroupType].
```

core/api/atrium-core-api-common/src/main/kotlin/ch/tutteli/atrium/assertions/builders/explanatoryGroup.kt
```diff
@@ -159,6 +164,11 @@ interface ExplanatoryAssertionGroupTypeOption {
      */
     val withWarningType: AssertionsOption<WarningAssertionGroupType, ExplanatoryAssertionGroupFinalStep>
 
+    /**
+     * Builder to create an [AssertionGroup] with a [InformationAssertionGroupType].
```
robstoll, line 168:
```suggestion
     * Builder to create an [AssertionGroup] with an [InformationAssertionGroupType].
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
```diff
@@ -41,6 +41,19 @@ fun <T> Expect<T>.isSameAs(expected: T): Expect<T> = _logicAppend { isSameAs(exp
...
+ * @param assertionCreator The group of assertions to make.
+ *
+ * @return An [Expect] for the current subject of the assertion.
+ * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ *
+ * @since 0.14.0
```
robstoll, line 52: Could you please provide a sample in addition in AnyAssertionSamples?  thanks
```suggestion
 * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
 *
 * @sample ch.tutteli.atrium.api.fluent.en_GB.samples.AnyAssertionSamples.because
 * @since 0.14.0
```

robstoll (changes requested):

jGleitz:
@robstoll I am happy to create an issue in atrium-roadmap, however, shouldn’t we at least discuss the name for the assertion group that is being added in _this PR_? I think that `InformationAssertionGroupType` adds to the confusion on the logic level.

robstoll:
I would keep InformationAssertionGroupType for now as it is consistent with `WarningAssertionGroupType`. Changing WarningAssertionGroupType would be a binary backward compatibility break. We can change things with 1.0.0

Valefant:
@robstoll `ch/tutteli/atrium/api/infix/en_GB/AnyAssertionsSpec.kt` is giving me a hard time, because it uses the AsciiReporter which tests for `(i)` and not for ℹ for the infix test. I can't remove it though, because some of the tests are depending on it.

I tried to specify another Reporter for my test cases but that didn't work either. One option could be that I would outsource the test for `because` into another spec, but that doesn't seem right. What should I do now? 🤔

robstoll:
@Valefant pass the bullet point for InformationAssertionGroupType to the spec in atrium-specs. For api-fluent you pass ":information_source: " and  for infix you pass "(i) ". I don't have the source code handy at hand, search for rootBulletPoint, I think you should find examples this way.

robstoll:
@jGleitz we can actually also break things earlier as it is not in the API but on the logic level :wink: Nevertheless, I prefer to merge this PR and break both types at the same time than to have to different naming. I look forward to your proposal in atrium-roadmap

Valefant:
@robstoll Hmmm all those scala errors are weird 🤔

robstoll:
I see, I should maybe provide the scala implementation in advance so that the only thing to do for Kotlin devs is to activate it in the spec. Can you please comment out the whole AnyAssertionSpec in atrium-scala2?
@jGleitz could you please checkout the branch and check if it behaves as you expect

jGleitz:
I see two issues:

 1. The output looks like this:
```
expected that subject: "filename?"        <1733163717>
◆ does not contain: 
  ⚬ value: "?"        <1051594688>
    ⚬ ▶ number of matches: 1
        ◾ is: 0        (kotlin.Int <503551375>)
  ℹ because ? is not allowed in file names on Windows
```

We decided `not` to indent the `i` (https://github.com/robstoll/atrium-roadmap/issues/15#issuecomment-711416949), so the output should look like this: 

```
expected that subject: "filename?"        <1733163717>
◆ does not contain: 
  ⚬ value: "?"        <1051594688>
    ⚬ ▶ number of matches: 1
        ◾ is: 0        (kotlin.Int <503551375>)
ℹ because ? is not allowed in file names on Windows
```

  2. The sample use rather silly examples. From my point of view, you should _not_ use `because` like this, because you are stating the obvious. I’d prefer if we’d use some of the examples we discussed in (atrium-roadmap#15)[https://github.com/robstoll/atrium-roadmap/issues/15], because those convey the real-world use case for `because`.

robstoll:
@Valefant do you need help in addressing the review comments of jGleitz?

PR #683, opened 2020-10-27
Draft: Move from tutteli-gradle-dokka to plain dokka 1.4

### LINKED ISSUE #641: update to new dokka plugin
## Non-Code related feature
Atrium is currently using dokka 0.9.8 via the tutteli-dokka plugin. 
We would like to update to dokka 1.4.10

Following the things we need to do (see https://github.com/robstoll/atrium/pull/683 most likely you can build on this):
- [ ] remove `classpath "ch.tutteli:tutteli-gradle-dokka:$tutteli_plugins_version"`  from build.gradle
- [ ] configure the new dokka plugin instead. You don't have to bother what was configured before, you can start from scratch but we have the following requirements:
  - [ ] we need only the html report (disable the rest)
  - [ ] make sure it is not included in the normal build (./gradlew build should not trigger it) - might be this is already the case. Use ` ./gradlew atrium-core-api-jvm:build --dry-run` and check if dokka is included. If it is, then find ways to deactivate it. We want to be able to run dokka separately
  - [ ] check if the samples for toBe are correctly rendered
- [ ] we still want an overall report when running ghPages. You probably need to adapt gradle/scripts/gh-pages.gradle  => see https://github.com/Kotlin/dokka/blob/master/runners/gradle-plugin/MIGRATION.md and https://github.com/Kotlin/dokka#multi-module-projects
  => make sure we still suppress the packages defined at the end of the gh-pages.gradle script.

*Hint*, the following might help to setup the plugin, note thought that Atrium is using groovy and not kotlin for gradle
https://github.com/Kotlin/dokka#using-dokka

## Your first contribution?

jgrgt:
Attempt to fix  #641. Kind of strugling with the submodules. But this produces some code documentation already. Though it does not look very nice...

Some things are commented out for now, they seem to make dokka go into an infinite loop... So there's still work to be done.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,7 @@ buildscript {
         niok_version = '1.3.4'
         niok = { "ch.tutteli.niok:niok:$niok_version" }
         kotlin_version = '1.3.72'
+        dokka_version = "1.4.10.2"
 
         // test
         jacoco_tool_version = '0.8.6'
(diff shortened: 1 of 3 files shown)


robstoll:
@jgrgt I suggest you write in the dokka channel on kotlinlang.slack.com in case you encounter a strange behaviour with dokka. I am sure they can help better than I can.

robstoll:
@jgrgt any progress here?

jgrgt:
@robstoll Sorry, no. Did not get a response on the dokka Slack. And I will not have the time to dig into this further in the short term. Feel free to close this and unassign me from the issue...

robstoll:
ok, closing it, maybe I can still use part of it. We will see, currently I don't have time neither. thanks already for this :+1:

PR #685, opened 2020-10-28
Path.hasDirectoryEntry

jGleitz:
* rename `Expect<Path>.contains` to `Expect<Path>.hasDirectoryEntry`
 * expose it in the infix API
 * fix its symlink behaviour

Exposes `LinkOptions` for `exists` and `existsNot` on the Domain, but not on the API level.
We can pick it up from there should we ever want to expose it.

### DIFF
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/pathAssertions.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/pathAssertions.kt
@@ -196,7 +196,7 @@ fun <T : Path> Expect<T>.resolve(other: String, assertionCreator: Expect<Path>.(
  * Therefore, if a symbolic link exists at the location the subject points to,
  * search will continue at the location the link points at.
  *
- * This assertion is not atomic with respect to concurrent file system operations on the paths the assertions works on.
+ * This assertion is not atomic with respect to concurrent file system operations on the paths the assertion works on.
  * Its result, in particular its extended explanations, may be wrong if such concurrent file system operations
  * take place.
  *
@@ -255,7 +255,7 @@ fun <T : Path> Expect<T>.isExecutable(): Expect<T> =
  * Therefore, if a symbolic link exists at the location the subject points to, search will continue
  * at the location the link points at.
  *
- * This assertion is not atomic with respect to concurrent file system operations on the paths the assertions works on.
+ * This assertion is not atomic with respect to concurrent file system operations on the paths the assertion works on.
  * Its result, in particular its extended explanations, may be wrong if such concurrent file system operations
  * take place.
  *
@@ -275,7 +275,7 @@ fun <T : Path> Expect<T>.isRegularFile(): Expect<T> =
  * Therefore, if a symbolic link exists at the location the subject points to, search will continue
  * at the location the link points at.
  *
- * This assertion is not atomic with respect to concurrent file system operations on the paths the assertions works on.
+ * This assertion is not atomic with respect to concurrent file system operations on the paths the assertion works on.
  * Its result, in particular its extended explanations, may be wrong if such concurrent file system operations
  * take place.
  *
@@ -321,7 +321,7 @@ fun <T : Path> Expect<T>.isRelative(): Expect<T> =
  * the link points at. If a symbolic link exists at one of the entries, this will fulfill the respective assertion and
  * the entry’s symbolic link will not be followed.
  *
- * This assertion is not atomic with respect to concurrent file system operations on the paths the assertions work on.
+ * This assertion is not atomic with respect to concurrent file system operations on the paths the ssertions work on.
  * The result, in particular its extended explanations, may be wrong if such concurrent file system operations
  * take place.
  *
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/pathAssertions.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/pathAssertions.kt
@@ -321,7 +321,7 @@ fun <T : Path> Expect<T>.isRelative(): Expect<T> =
  * the link points at. If a symbolic link exists at one of the entries, this will fulfill the respective assertion and
  * the entry’s symbolic link will not be followed.
  *
- * This assertion is not atomic with respect to concurrent file system operations on the paths the ssertions work on.
+ * This assertion is not atomic with respect to concurrent file system operations on the paths the assertion works on.
  * The result, in particular its extended explanations, may be wrong if such concurrent file system operations
  * take place.
  *
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/pathAssertions.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/pathAssertions.kt
@@ -7,7 +7,7 @@ package ch.tutteli.atrium.api.fluent.en_GB
 
 import ch.tutteli.atrium.creating.Expect
 import ch.tutteli.atrium.logic.*
-import ch.tutteli.kbox.forElementAndForEachIn
+import ch.tutteli.kbox.glue
 import java.nio.charset.Charset
 import java.nio.file.Path
 
@@ -312,17 +312,17 @@ fun <T : Path> Expect<T>.isRelative(): Expect<T> =
     _logicAppend { isRelative() }
 
 /**
- * Expects that the subject of the assertion (a [Path]) is a directory;
- * meaning that there is a file system entry at the location the [Path] points to and that is a directory.
+ * Expects that the subject of the assertion (a [Path]) is a directory having the provided entries.
+ * That means that there is a file system entry at the location the [Path] points to and that it is a directory.
+ * Furthermore, every argument string resolved against the subject yields an existing file system entry.
  *
- * Every argument string is expected to exist as a child file or child directory for the subject of the assertion.
+ * This assertion _resolves_ symbolic links for the subject, but not for the entries.
+ * Therefore, if a symbolic link exists at the location the subject points to, the search will continue at the location
+ * the link points at. If a symbolic link exists at one of the entries, this will fulfill the respective assertion and
+ * the entry’s symbolic link will not be followed.
  *
- * This assertion _resolves_ symbolic links.
- * Therefore, if a symbolic link exists at the location the subject points to, search will continue
- * at the location the link points at.
- *
- * This assertion is not atomic with respect to concurrent file system operations on the paths the assertions works on.
- * Its result, in particular its extended explanations, may be wrong if such concurrent file system operations
+ * This assertion is not atomic with respect to concurrent file system operations on the paths the assertions work on.
+ * The result, in particular its extended explanations, may be wrong if such concurrent file system operations
  * take place.
  *
  * @return An [Expect] for the current subject of the assertion.
@@ -330,12 +330,8 @@ fun <T : Path> Expect<T>.isRelative(): Expect<T> =
  *
  * @since 0.14.0
  */
-fun <T : Path> Expect<T>.contains(path: String, vararg otherPaths: String): Expect<T> =
-    isDirectory() and {
-        forElementAndForEachIn(path, otherPaths) { p ->
-            resolve(p) { exists() }
-        }
-    }
+fun <T : Path> Expect<T>.hasDirectoryEntry(entry: String, vararg otherEntries: String): Expect<T> =
+    _logicAppend { hasDirectoryEntry(entry glue otherEntries) }
 
 /**
  * Creates an [Expect] for the property [Path.extension][ch.tutteli.niok.extension]
--- a/apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/creating/path/DirectoryEntries.kt
+++ b/apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/creating/path/DirectoryEntries.kt
@@ -1,7 +1,7 @@
 package ch.tutteli.atrium.api.infix.en_GB.creating.path
 
 /**
- *  Parameter object which collects directory entires (as [String]s).
+ *  Parameter object which collects directory entries (as [String]s).
  *  Use the function `directoryEntry(String, vararg String)` to create this representation.
  *
  *  @since 0.14.0
--- a/apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/creating/path/DirectoryEntries.kt
+++ b/apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/creating/path/DirectoryEntries.kt
@@ -1,0 +1,11 @@
+package ch.tutteli.atrium.api.infix.en_GB.creating.path
+
+/**
+ *  Parameter object which collects directory entires (as [String]s).
+ *  Use the function `directoryEntry(String, vararg String)` to create this representation.
+ *
+ *  @since 0.14.0
+ */
+// TODO make inline once we drop support for Kotlin 1.2
+class DirectoryEntries(val entries: List<String>)
+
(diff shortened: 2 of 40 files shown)


jGleitz:
I also added `@formatter:off` to the generated logic files because my IDE kept breaking them on commit

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/pathAssertions.kt
```diff
@@ -312,30 +312,26 @@ fun <T : Path> Expect<T>.isRelative(): Expect<T> =
...
- * Therefore, if a symbolic link exists at the location the subject points to, search will continue
- * at the location the link points at.
+ * This assertion _resolves_ symbolic links for the subject, but not for the entries.
+ * Therefore, if a symbolic link exists at the location the subject points to, the search will continue at the location
+ * the link points at. If a symbolic link exists at one of the entries, this will fulfill the respective assertion and
+ * the entry’s symbolic link will not be followed.
  *
- * This assertion is not atomic with respect to concurrent file system operations on the paths the assertions works on.
- * Its result, in particular its extended explanations, may be wrong if such concurrent file system operations
+ * This assertion is not atomic with respect to concurrent file system operations on the paths the assertions work on.
```
robstoll, line 324: Could you please fix this typo also in other files?

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/creating/path/DirectoryEntries.kt
```diff
@@ -0,0 +1,11 @@
+package ch.tutteli.atrium.api.infix.en_GB.creating.path
+
+/**
+ *  Parameter object which collects directory entires (as [String]s).
```
robstoll, line 4:
```suggestion
 *  Parameter object which collects directory entries (as [String]s).
```

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/creating/path/DirectoryEntries.kt
```diff
@@ -0,0 +1,11 @@
...
+ *  Use the function `directoryEntry(String, vararg String)` to create this representation.
+ *
+ *  @since 0.14.0
+ */
+// TODO make inline once we drop support for Kotlin 1.2
```
robstoll, line 9:
```suggestion
// TODO make inline once we drop support for Kotlin 1.2 (most likely with 1.0.0)
```
For me because I plan to go through TODO*1.0.0 at some point

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/creating/path/DirectoryEntries.kt
```diff
@@ -0,0 +1,11 @@
...
+class DirectoryEntries(val entries: List<String>)
```
robstoll, line 10: In all other helper objects we use `T, Arra<<out T>` and extend from VarArgHelper. I would do the same here

logic/atrium-logic-jvm/src/generated/kotlin/ch/tutteli/atrium/logic/path.kt
```diff
@@ -24,8 +26,8 @@ fun <T : Path> AssertionContainer<T>.startsNotWith(expected: Path): Assertion =
 fun <T : Path> AssertionContainer<T>.endsWith(expected: Path): Assertion = impl.endsWith(this, expected)
 fun <T : Path> AssertionContainer<T>.endsNotWith(expected: Path): Assertion = impl.endsNotWith(this, expected)
 
-fun <T : Path> AssertionContainer<T>.exists(): Assertion = impl.exists(this)
-fun <T : Path> AssertionContainer<T>.existsNot(): Assertion = impl.existsNot(this)
+fun <T : Path> AssertionContainer<T>.exists(vararg linkOptions: LinkOption): Assertion = impl.exists(this, *linkOptions)
```
robstoll, line 29: we never used varag on the logic level. Unless required to call the jvm API I would use List instead.

logic/generateLogic.gradle
```diff
@@ -84,16 +85,16 @@ def createGenerateLogicTaskForPackage(
             def identifier = /[a-zA-Z0-9]+/
             def newLine = /(?:\r\n?|\n)/
             def newLineAndIndent = /$newLine\s*/
-            def parameter = { paramNumber -> /,(?: |$newLineAndIndent)(?<paramName$paramNumber>$identifier): (?<typeName$paramNumber>[^:]+?)/}
+            def parameter = { paramNumber -> /,(?: |$newLineAndIndent)(?<vararg$paramNumber>vararg )?(?<paramName$paramNumber>$identifier): (?<typeName$paramNumber>[^:]+?)/}
```
robstoll, line 88: revert, not required if we don't use varag in logic

logic/generateLogic.gradle
```diff
@@ -84,16 +85,16 @@ def createGenerateLogicTaskForPackage(
...
                 def params = numOfParams > 0 ? (1..numOfParams) : []
                 new Tuple3<Pattern, String, String>(
                     Pattern.compile(typeIdentifier + params.collect { paramNumber -> parameter(paramNumber)}.join("") + returnType),
-                    /fun ${ '${generics}${type}.${funcName}' }\(/ + params.collect { paramNumber -> /${'${paramName' + paramNumber + '}'}: ${ '${typeName' + paramNumber + '}'}/ }.join(", ") + /\)${ '${returnType}' } =/ + (numOfParams > 1 ? "$ln    " : " "),
-                    /.${ '${funcName}' }\(this/ + (numOfParams > 0 ? ", " : "") + params.collect { paramNumber -> /${ '${paramName'+ paramNumber + '}'}/ }.join(", ") + /\)/
+                    /fun ${ '${generics}${type}.${funcName}' }\(/ + params.collect { paramNumber -> /${'${vararg' + paramNumber + '}'}${'${paramName' + paramNumber + '}'}: ${ '${typeName' + paramNumber + '}'}/ }.join(", ") + /\)${ '${returnType}' } =/ + (numOfParams > 1 ? "$ln    " : " "),
+                    /.${ '${funcName}' }\(this/ + (numOfParams > 0 ? ", " : "") + params.collect { paramNumber -> /${'${vararg' + paramNumber + '}'}${ '${paramName'+ paramNumber + '}'}/ }.join(", ") + /\)/
```
robstoll, line 97: revert

logic/generateLogic.gradle
```diff
@@ -117,6 +118,9 @@ def createGenerateLogicTaskForPackage(
                 patterns.forEach { triple ->
                     tmp = tmp.replaceAll(triple.first, triple.second + implValName + triple.third)
                 }
+                // when calling a varargs method, the call needs to use the spread operator (“*”), but the regex
+                // can only generate the string “vararg ”
+                tmp = tmp.replaceAll(/(=.*)vararg /, /$1\*/)
```
robstoll, line 123: revert

robstoll (changes requested):

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/pathAssertions.kt
jGleitz: Done.

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/creating/path/DirectoryEntries.kt
jGleitz: Done.

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/creating/path/DirectoryEntries.kt
jGleitz: Obsolete because of using `VarArgHelper` -> cannot be inlined

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/creating/path/DirectoryEntries.kt
jGleitz: Done.

jGleitz:
About using `vararg` on the logic level: Is there any specific reason not to it? Using a `List` feels awkward to me. We could make it a `LinkOption?`, because there is only one `LinkOption` anyway, but it somehow feels wrong. The Java standard library has established the pattern of `vararg` enum options and I think it is a good pattern.

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/pathAssertions.kt
```diff
@@ -321,7 +321,7 @@ fun <T : Path> Expect<T>.isRelative(): Expect<T> =
  * the link points at. If a symbolic link exists at one of the entries, this will fulfill the respective assertion and
  * the entry’s symbolic link will not be followed.
  *
- * This assertion is not atomic with respect to concurrent file system operations on the paths the assertions work on.
+ * This assertion is not atomic with respect to concurrent file system operations on the paths the ssertions work on.
```
robstoll, line 324:
```suggestion
 * This assertion is not atomic with respect to concurrent file system operations on the paths the assertion works on.
```

robstoll (changes requested):

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/pathAssertions.kt
jGleitz: wtf

robstoll:
I am not using it due to the copy semantic of Array. This is the reason many libraries often provide multiple overloads like the following:
```
test (t: T)
test (t1: T, t2: T)
test (t1: T, t2: T, t3: T)
test (t1: T, t2: T, t3: T, varag T)
```

as far as I know, `vararg` are only used on the API level also in the JVM mainly due to copy semantic
Another thing is heap pollution (https://en.wikipedia.org/wiki/Heap_pollution) even though this can be handled is not a problem if used with care IMO.

jGleitz:
Well, heap pollution is not a problem when not using generics, and the array copy issue seems a lot like premature optimization to me. But alright, I’ll go for `LinkOption?`.

logic/generateLogic.gradle
jGleitz: Done

logic/generateLogic.gradle
jGleitz: Done

logic/generateLogic.gradle
jGleitz: Done

logic/atrium-logic-jvm/src/generated/kotlin/ch/tutteli/atrium/logic/path.kt
jGleitz: Done

robstoll:
The JIT these days are much better than when I learned to use arrays in Java sparsely. Yet, I am not sure if they are so good that they can really optimize the copy semantic away. In your example it really doesn't matter as you only pass one argument (even if this should change it is only a few). But... vararg also introduce a bit more complexity and if we introduce it on the logic level then it should most likely used everywhere. 
You are not the first one trying to introduce vararg on the logic level (you are actually the 2nd person and both in this month, strange coincidence :laughing: ) in case there should be more people wanting to use vararg on the logic level, then we can re-discuss this of course.

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultPathAssertions.kt
```diff
@@ -180,4 +183,12 @@ class DefaultPathAssertions : PathAssertions {
         other: String
     ): FeatureExtractorBuilder.ExecutionStep<T, Path> = container.f1<T, String, Path>(Path::resolve, other)
 
+    override fun <T : Path> hasDirectoryEntry(container: AssertionContainer<T>, entries: List<String>): Assertion =
+        assertionBuilder.invisibleGroup.withAssertions(
+            listOf(isDirectory(container)) +
+                entries.map { entry ->
+                    resolve(container, entry)
+                        .collect { _logic.addAssertion(exists(_logic, NOFOLLOW_LINKS)) }
```
robstoll, line 191: don't use the methods of this class directly as one might use a different implementation. Always take the indirection of _logic
```suggestion
                    container.resolve(entry).collect { _logic.exists(NOFOLLOW_LINKS) }
```

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultPathAssertions.kt
```diff
@@ -180,4 +183,12 @@ class DefaultPathAssertions : PathAssertions {
         other: String
     ): FeatureExtractorBuilder.ExecutionStep<T, Path> = container.f1<T, String, Path>(Path::resolve, other)
 
+    override fun <T : Path> hasDirectoryEntry(container: AssertionContainer<T>, entries: List<String>): Assertion =
+        assertionBuilder.invisibleGroup.withAssertions(
+            listOf(isDirectory(container)) +
```
robstoll, line 188: if the subject is not a directory, then all other assertions will fail as well. Maybe fail fast is better in this case? On the other hand, we get more context in the error report which might be nice as well.
Keep it this way

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultPathAssertions.kt
robstoll: same applies to isDirectory, also use `container.isDirectory()`

robstoll:
@jGleitz how about we expose the LinkOption as vararg in the API? If we do, then it would make sense to use List on the logic level. We could of course also expose it as LinkOption?

jGleitz:
I don't want to expose `LinkOption` in this PR. Taking your lessons to heart, I think we should not expose it on the API level before anyone needs it. However, having it already on the logic level makes it trivial to expose it. And I agree that I makes sense to think about whether the solution we build today makes sense in the future.

If you care about copying an array with only one element, I figured you'd also care about creating an empty array for an empty list every time the method is called without arguments. That is why I went with `LinkOption?`.

Personally, I think that thinking about low-level operations at this level of detail has no benefit for performance at all¹. That is why I would have gone for `vararg` all the way through. But I am happy to implement whatever approach you think is the best compromise for clean code, API consistency, and “performance”.

¹at that level, any compiler optimization can invalidate our assumptions. If array copying was a relevant performance issue, I'd hope that JetBrains would optimize it away.

robstoll:
I am with you that we don't need it in this PR but I could create an issue. Since you had the need yourself on the logic level I figured it makes sense to expose it on the API level. I create a good first issue but not implement it myself.

And let's keep LinkOption? for now, Java uses a vararg to be future proof regarding backward compatibility, we can break things faster than java and currently there is just one LinkOption so IMO `LinkOption?` is the better fit

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultPathAssertions.kt
```diff
@@ -180,4 +183,12 @@ class DefaultPathAssertions : PathAssertions {
...
+                entries.map { entry ->
+                    resolve(container, entry)
+                        .collect { _logic.addAssertion(exists(_logic, NOFOLLOW_LINKS)) }
+                }
```
robstoll, line 192: don't use the methods of this class directly as one might use a different implementation. Always take the indirection of _logic
```suggestion
            listOf(container.isDirectory()) +
                entries.map { entry ->
                    container.resolve(entry).collect { _logic.exists(NOFOLLOW_LINKS) }
                }
```

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultPathAssertions.kt
```diff
@@ -180,4 +183,11 @@ class DefaultPathAssertions : PathAssertions {
         other: String
     ): FeatureExtractorBuilder.ExecutionStep<T, Path> = container.f1<T, String, Path>(Path::resolve, other)
 
+    override fun <T : Path> hasDirectoryEntry(container: AssertionContainer<T>, entries: List<String>): Assertion =
+        assertionBuilder.invisibleGroup.withAssertions(
+            listOf(container.isDirectory()) +
+                entries.map { entry ->
+                    container.resolve(entry).collect { _logic.exists(NOFOLLOW_LINKS) }
```
robstoll, line 190:
```suggestion
                    container.resolve(entry).collect { _logicAppend { exists(NOFOLLOW_LINKS) } }
```

jGleitz:
Thanks for taking care of the missing pieces of this PR, @robstoll

robstoll:
thanks for the implementation :+1:

PR #699, opened 2020-11-16
rename logic.iterable.contains to logic.iterablelike.contains

robstoll:


robstoll:
closing this, as it is a bc break, going to do it with 1.0.0

PR #714, opened 2020-12-13
🚧 Adopt the ‘to+infinitive’ Convention

jGleitz:
Let’s get the ball rolling!

This PR will introduce the [‘to+infinitive’ convention](https://github.com/robstoll/atrium-roadmap/issues/93) for atrium.
It will not be merged before the convention has been applied to all of the fluent API. 
It exists to gather feedback and find merge conflicts during the transition.

### DIFF
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
@@ -1,8 +1,8 @@
 package ch.tutteli.atrium.api.fluent.en_GB
 
 import ch.tutteli.atrium.creating.Expect
-import ch.tutteli.atrium.logic.creating.typeutils.IterableLike
 import ch.tutteli.atrium.logic.*
+import ch.tutteli.atrium.logic.creating.typeutils.IterableLike
 import ch.tutteli.atrium.logic.utils.iterableLikeToIterable
 import ch.tutteli.atrium.reporting.Reporter
 import ch.tutteli.kbox.glue
@@ -30,22 +30,56 @@ fun <T> Expect<T>.notToBe(expected: T): Expect<T> = _logicAppend { notToBe(expec
 /**
  * Expects that the subject of the assertion is the same instance as [expected].
  *
+ * Deprecated as atrium moves to a consistent ‘to + infinitive’ naming convention. Use [toBeTheSameAs] instead.
+ * This function will be removed in version 1.0.0. See
+ * [atrium-roadmap#93](https://github.com/robstoll/atrium-roadmap/issues/93) for details and to give feedback.
+ *
  * @return An [Expect] for the current subject of the assertion.
  * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
- *
- * @sample ch.tutteli.atrium.api.fluent.en_GB.samples.AnyAssertionSamples.isSameAs
  */
-fun <T> Expect<T>.isSameAs(expected: T): Expect<T> = _logicAppend { isSameAs(expected) }
+@Deprecated(
+    "Replaced by toBeTheSameAs. Will be removed with version 1.0.0",
+    ReplaceWith("this.toBeTheSameAs<T>(expected)")
+) // TODO remove with 1.0.0
+fun <T> Expect<T>.isSameAs(expected: T): Expect<T> = toBeTheSameAs(expected)
+
+/**
+ * Expects that the subject of `this` assertion is the same instance as [expected].
+ *
+ * @return an [Expect] for the current subject of `this` assertion.
+ * @throws AssertionError if the subject of `this` assertion is not the same instance as [expected].
+ *
+ * @since 0.16.0
+ * @sample ch.tutteli.atrium.api.fluent.en_GB.samples.AnyAssertionSamples.toBeTheSameAs
+ */
+fun <T> Expect<T>.toBeTheSameAs(expected: T): Expect<T> = _logicAppend { toBeTheSameAs(expected) }
 
 /**
  * Expects that the subject of the assertion is not the same instance as [expected].
  *
+ * Deprecated as atrium moves to a consistent ‘to + infinitive’ naming convention. Use [notToBeTheSameAs] instead.
+ * This function will be removed in version 1.0.0. See
+ * [atrium-roadmap#93](https://github.com/robstoll/atrium-roadmap/issues/93) for details and to give feedback.
+ *
  * @return An [Expect] for the current subject of the assertion.
  * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
- *
- * @sample ch.tutteli.atrium.api.fluent.en_GB.samples.AnyAssertionSamples.isNotSameAs
  */
-fun <T> Expect<T>.isNotSameAs(expected: T): Expect<T> = _logicAppend { isNotSameAs(expected) }
+@Deprecated(
+    "Replaced by notToBeTheSameAs. Will be removed with version 1.0.0",
+    ReplaceWith("this.notToBeTheSameAs<T>(expected)")
+) // TODO remove with 1.0.0
+fun <T> Expect<T>.isNotSameAs(expected: T): Expect<T> = notToBeTheSameAs(expected)
+
+/**
+ * Expects that the subject of `this` assertion is not the same instance as [expected].
+ *
+ * @return an [Expect] for the current subject of `this` assertion.
+ * @throws AssertionError if the subject of `this` assertion is the same instance as [expected].
+ *
+ * @since 0.16.0
+ * @sample ch.tutteli.atrium.api.fluent.en_GB.samples.AnyAssertionSamples.notToBeTheSameAs
+ */
+fun <T> Expect<T>.notToBeTheSameAs(expected: T): Expect<T> = _logicAppend { notToBeTheSameAs(expected) }
 
 /**
  * Expects that the subject of the assertion is either `null` in case [assertionCreatorOrNull]
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/AnyAssertionsSpec.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/AnyAssertionsSpec.kt
@@ -1,11 +1,7 @@
 package ch.tutteli.atrium.api.fluent.en_GB
 
 import ch.tutteli.atrium.creating.Expect
-import ch.tutteli.atrium.specs.feature0
-import ch.tutteli.atrium.specs.fun1
-import ch.tutteli.atrium.specs.fun2
-import ch.tutteli.atrium.specs.withFeatureSuffix
-import ch.tutteli.atrium.specs.withNullableSuffix
+import ch.tutteli.atrium.specs.*
 import kotlin.reflect.KFunction2
 import kotlin.reflect.KProperty1
 
@@ -18,14 +14,14 @@ class AnyAssertionsSpec : ch.tutteli.atrium.specs.integration.AnyAssertionsSpec(
     fun1(Expect<DataClass>::notToBe),
     fun1(Expect<Int?>::notToBe).withNullableSuffix(),
     fun1(Expect<DataClass?>::notToBe).withNullableSuffix(),
-    fun1(Expect<Int>::isSameAs),
-    fun1(Expect<DataClass>::isSameAs),
-    fun1(Expect<Int?>::isSameAs).withNullableSuffix(),
-    fun1(Expect<DataClass?>::isSameAs).withNullableSuffix(),
-    fun1(Expect<Int>::isNotSameAs),
-    fun1(Expect<DataClass>::isNotSameAs),
-    fun1(Expect<Int?>::isNotSameAs).withNullableSuffix(),
-    fun1(Expect<DataClass?>::isNotSameAs).withNullableSuffix(),
+    fun1(Expect<Int>::toBeTheSameAs),
+    fun1(Expect<DataClass>::toBeTheSameAs),
+    fun1(Expect<Int?>::toBeTheSameAs).withNullableSuffix(),
+    fun1(Expect<DataClass?>::toBeTheSameAs).withNullableSuffix(),
+    fun1(Expect<Int>::notToBeTheSameAs),
+    fun1(Expect<DataClass>::notToBeTheSameAs),
+    fun1(Expect<Int?>::notToBeTheSameAs).withNullableSuffix(),
+    fun1(Expect<DataClass?>::notToBeTheSameAs).withNullableSuffix(),
     fun2(Expect<Int>::isNoneOf),
     fun2(Expect<DataClass>::isNoneOf),
     fun2(Expect<Int?>::isNoneOf).withNullableSuffix(),
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/AnyAssertionSamples.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/AnyAssertionSamples.kt
@@ -38,24 +38,24 @@ class AnyAssertionSamples {
     }
 
     @Test
-    fun isSameAs() {
+    fun toBeTheSameAs() {
         val list = listOf(3)
-        expect(list).isSameAs(list)
+        expect(list).toBeTheSameAs(list)
 
         fails {
-            // fails because isSameAs is based on identity, use toBe for equality
-            expect(listOf(3)).isSameAs(listOf(3))
+            // fails because toBeTheSameAs is based on identity, use toBe for equality
+            expect(listOf(3)).toBeTheSameAs(listOf(3))
         }
     }
 
     @Test
-    fun isNotSameAs() {
-       // holds because isSameAs is based on identity, use notToBe for equality
-        expect(listOf(2)).isNotSameAs(listOf(2))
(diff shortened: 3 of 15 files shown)


jGleitz:
You’ll notice that I made cosmetical adaptions to the docs of the new functions. I think they are improvements and am happy to discuss them!

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
```diff
@@ -30,22 +30,56 @@ fun <T> Expect<T>.notToBe(expected: T): Expect<T> = _logicAppend { notToBe(expec
...
+    ReplaceWith("this.toBeTheSameAs<T>(expected)")
+) // TODO remove with 1.0.0
+fun <T> Expect<T>.isSameAs(expected: T): Expect<T> = toBeTheSameAs(expected)
+
+/**
+ * Expects that the subject of `this` assertion is the same instance as [expected].
  *
- * @sample ch.tutteli.atrium.api.fluent.en_GB.samples.AnyAssertionSamples.isSameAs
+ * @return an [Expect] for the current subject of `this` assertion.
+ * @throws AssertionError if the subject of `this` assertion is not the same instance as [expected].
```
robstoll, line 50: Might needs to be in the text as it depends on the context whether this function throws or not. I am also happy with removing the `@throws AssertionError` from all functions since this adds little to the KDoc

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
```diff
@@ -30,22 +30,56 @@ fun <T> Expect<T>.notToBe(expected: T): Expect<T> = _logicAppend { notToBe(expec
...
+    "Replaced by notToBeTheSameAs. Will be removed with version 1.0.0",
+    ReplaceWith("this.notToBeTheSameAs<T>(expected)")
+) // TODO remove with 1.0.0
+fun <T> Expect<T>.isNotSameAs(expected: T): Expect<T> = notToBeTheSameAs(expected)
+
+/**
+ * Expects that the subject of `this` assertion is not the same instance as [expected].
+ *
+ * @return an [Expect] for the current subject of `this` assertion.
+ * @throws AssertionError if the subject of `this` assertion is the same instance as [expected].
```
robstoll, line 77: `is not the same`

I think we should either remove the `@throws` or keep the generic message. To cumbersome and error prone to maintain/keep in sync with the description IMO and there is no gain in repeating it, or is there?

robstoll (changes requested):

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
```diff
@@ -30,22 +30,56 @@ fun <T> Expect<T>.notToBe(expected: T): Expect<T> = _logicAppend { notToBe(expec
...
  * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ */
+@Deprecated(
+    "Replaced by toBeTheSameAs. Will be removed with version 1.0.0",
+    ReplaceWith("this.toBeTheSameAs<T>(expected)")
+) // TODO remove with 1.0.0
+fun <T> Expect<T>.isSameAs(expected: T): Expect<T> = toBeTheSameAs(expected)
+
+/**
+ * Expects that the subject of `this` assertion is the same instance as [expected].
```
robstoll, line 47: I am not entirely sure if `this` helps or confuses because by only reading this sentence, I would guess this function is an extension of `Assertion` due to the way `this` is wrapped with back-ticks. 
However, I think it would be an improvement if it is not wrapped with back-ticks and could be changed everywhere.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
```diff
@@ -30,22 +30,56 @@ fun <T> Expect<T>.notToBe(expected: T): Expect<T> = _logicAppend { notToBe(expec
...
  *
- * @sample ch.tutteli.atrium.api.fluent.en_GB.samples.AnyAssertionSamples.isSameAs
+ * @return an [Expect] for the current subject of `this` assertion.
```
robstoll, line 49: Thanks for all the little improvements :+1: 
Same problematic as above with `this` and maybe it would also make sense to remove `current`? So my suggestion including your improvement:
`@return An [Expect] for the subject of this assertion.`

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
```diff
@@ -30,22 +30,56 @@ fun <T> Expect<T>.notToBe(expected: T): Expect<T> = _logicAppend { notToBe(expec
 /**
  * Expects that the subject of the assertion is the same instance as [expected].
  *
+ * Deprecated as atrium moves to a consistent ‘to + infinitive’ naming convention. Use [toBeTheSameAs] instead.
```
robstoll, line 33: Detail but more aligned with the readme:
```suggestion
 * Deprecated as Atrium moves to a consistent ‘to + infinitive’ naming convention. Use [toBeTheSameAs] instead.
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
jGleitz: I think the doc either needs to state clearly when the exception is thrown, or not at all. ‘Might’ makes me mad because I’d like to know _when_.

Proposal:

> if the subject of `this` assertion is not the same instance as [expected] and this function is not called from inside an assertion creator lambda. If the call happens from inside an assertion creator lambda, the exception will be thrown by the function receiving the lambda.

But I am also completely fine with just removing all the `@throws`. I just think that if they are there, they need to be clear.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
jGleitz: I usually like writing ‘this’ as `this`, to point out that we are talking about the programming meaning of ‘this’. However, I agree that it is a little irritating because the receiver type is `Expect`.

Maybe this works:

 > Expects that the subject of `this` expectation is the same instance as [expected].

But I am also fine with just writing ‘this’, without the backticks. You’ll be the judge.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
jGleitz: I am very happy to leave out ‘current’.

I would like to keep the lowercase ‘an’ since what follows is a phrase started by the word ‘return’. To capitalize, it should be a proper sentence (see also the [Javadoc convention for @param](https://www.oracle.com/de/technical-resources/articles/java/javadoc-tool.html#@param)

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
jGleitz: by the way, ‘is the same’ is correct here: The assertion throws if the instance are actually the same.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
robstoll: So I suggest to remove them, I guess most people will expect that it throws an AssertionError and if in doubt, it is easy to find out if required. Less cluttering the code with something which has only little gain is a code thing IMO.

Btw. it is hard to come up with something which is actually true in all cases. 

> if the subject of this assertion is not the same instance as [expected] and this function is not called from inside an assertion creator lambda. If the call happens from inside an assertion creator lambda, the exception will be thrown by the function receiving the lambda. But only if it is not contributed to an ExplanatoryAssertionGroup or similar because such AssertionGroups usually do not cause an Exception... unless someone has explicitly marked them to fail. Yet, it still depends on the Expect if it is going to throw or not, in case of the default implementation, it is actually only the RootExpect which will throw... but you never know :laughing: 

And that's really not helpful either for a user -- I guess it is kind of helpful for a core contributor, to better understand when an exception is thrown but should probably be documented elsewhere.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
robstoll: > Expects that the subject of `this` expectation is the same instance as [expected].

Sound good, we should replace `subject of the assertion` in all places with ``subject of  `this` expectation``

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
robstoll: true :slightly_smiling_face:

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
robstoll: So:
```
@return an [Expect] for the subject of `this` expectation.
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
jGleitz: :+1:. I think we should try to adopt the lingo ‘expectation’ instead of ‘assertion’ everywhere.

robstoll: Agreed. Let's wait with such changes until we merge into master to avoid conflicts. I create a project, where we can track such tasks

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
robstoll: I suggest we remove them also after the merge, created a task in the project:
https://github.com/robstoll/atrium/projects/1#column-12080551

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
jGleitz: I’d suggest the opposite: Do these simple improvements now, because they are the smaller, easier changes, and start the renaming afterwards. What do you think?

robstoll: Good as well if you create a PR targeting master, then we can do it before merging the renames into master. But we should wait until I merge the MapLike.contains as we need to change those files as well.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
```diff
@@ -30,22 +30,56 @@ fun <T> Expect<T>.notToBe(expected: T): Expect<T> = _logicAppend { notToBe(expec
...
+ * [atrium-roadmap#93](https://github.com/robstoll/atrium-roadmap/issues/93) for details and to give feedback.
+ *
  * @return An [Expect] for the current subject of the assertion.
  * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ */
+@Deprecated(
+    "Replaced by toBeTheSameAs. Will be removed with version 1.0.0",
+    ReplaceWith("this.toBeTheSameAs<T>(expected)")
+) // TODO remove with 1.0.0
+fun <T> Expect<T>.isSameAs(expected: T): Expect<T> = toBeTheSameAs(expected)
```
robstoll, line 44: @jGleitz IMO we should provide once again a script in order that users have a smooth migration experience. Do you see the draft release here? https://github.com/robstoll/atrium/releases/tag/untagged-c4eefbbdd8d22b72bc20

jGleitz: > Do you see the draft release

No, I get a 404 for that link.

Writing a script feels like duplicating what the `@ReplaceWith` annotations do anyway. Isn’t there any IntelliJ functionality to say ‘perform all migrations’?

robstoll: Not that I know of, you were able to choose `Replace for all` for a particular ReplaceWith but last time I wanted to use it I couldn't find the option. Moreover, from my experience, the applications of the ReplaceWith by Intellij are many times buggy and don't work as they should. I was always happy that I had the scripts. I'll put a draft of the release note into the wiki so that you could provide the replacements as well
Up to you if you want to help out or not. It's also fine if I collect/write down the replacements at the end of 0.16.0

jGleitz: I see. I’ll prioritize work on my open pull requests, but will help on the script if I find the time.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
```diff
@@ -30,22 +30,56 @@ fun <T> Expect<T>.notToBe(expected: T): Expect<T> = _logicAppend { notToBe(expec
...
  *
+ * Deprecated as atrium moves to a consistent ‘to + infinitive’ naming convention. Use [toBeTheSameAs] instead.
+ * This function will be removed in version 1.0.0. See
+ * [atrium-roadmap#93](https://github.com/robstoll/atrium-roadmap/issues/93) for details and to give feedback.
+ *
  * @return An [Expect] for the current subject of the assertion.
  * @throws AssertionError Might throw an [AssertionError] if the assertion made is not correct.
+ */
+@Deprecated(
+    "Replaced by toBeTheSameAs. Will be removed with version 1.0.0",
```
robstoll, line 41: I plan to remove things maybe earlier than 1.0.0. They will certainly stay for a while but maybe we already remove it in 0.18.0 or 0.19.0
```suggestion
    "Replaced by toBeTheSameAs. Will be removed latest with version 1.0.0",
```

robstoll: Done for existing `@Deprecated` annotations

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
robstoll: going to address this now.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyAssertions.kt
robstoll: going to address this now.

robstoll:
closing this in favour of #866 (other parts have been already merged into master via other PRs)

PR #729, opened 2020-12-21
#660 because

robstoll:

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,8 @@ For instance, the [README of v0.14.0](https://github.com/robstoll/atrium/tree/v0
     - [Shortcut Functions](#shortcut-functions-1)
     - [Sophisticated Assertion Builders](#sophisticated-assertion-builders-1)
     - [Others](#others)
-  - [Path Assertions](#path-assertions)    
+  - [Path Assertions](#path-assertions)
+  - [Attach a reason](#attach-a-reason)
   - [Data Driven Testing](#data-driven-testing)
   - [Further Examples](#further-examples)  
   - [Sample Projects](#sample-projects)
@@ -389,7 +390,7 @@ expected that subject: () -> kotlin.Nothing        (readme.examples.ReadmeSpec$1
         » stacktrace: 
           ⚬ readme.examples.ReadmeSpec$1$4$1.invoke(ReadmeSpec.kt:76)
           ⚬ readme.examples.ReadmeSpec$1$4$1.invoke(ReadmeSpec.kt:51)
-          ⚬ readme.examples.ReadmeSpec$1$4.invoke(ReadmeSpec.kt:654)
+          ⚬ readme.examples.ReadmeSpec$1$4.invoke(ReadmeSpec.kt:691)
           ⚬ readme.examples.ReadmeSpec$1$4.invoke(ReadmeSpec.kt:51)
 ```
 </ex-toThrow1>
@@ -1486,19 +1487,19 @@ There are more assertion functions, a full list can be found in
 Atrium’s assertions for paths give detailed failure hints explaining what happened on the file system.
 For example, `exists` will explain which entry was the first one missing:
 
-<ex-path-exsits>
+<ex-path-exists>
 
 ```kotlin
 expect(Paths.get("/usr/bin/noprogram")).exists()
 ```
-↑ <sub>[Example](https://github.com/robstoll/atrium/tree/master/misc/tools/readme-examples/src/main/kotlin/readme/examples/ReadmeSpec.kt#L352)</sub> ↓ <sub>[Output](#ex-path-exsits)</sub>
-<a name="ex-path-exsits"></a>
+↑ <sub>[Example](https://github.com/robstoll/atrium/tree/master/misc/tools/readme-examples/src/main/kotlin/readme/examples/ReadmeSpec.kt#L352)</sub> ↓ <sub>[Output](#ex-path-exists)</sub>
+<a name="ex-path-exists"></a>
 ```text
 expected that subject: /usr/bin/noprogram        (sun.nio.fs.UnixPath <1234789>)
 ◆ to: exist
     » the closest existing parent directory is /usr/bin
 ```
-</ex-path-exsits>
+</ex-path-exists>
 
 Atrium will give details about why something cannot be accessed, for example when checking whether a file is writable:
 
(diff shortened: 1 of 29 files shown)


misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
```diff
@@ -174,7 +183,10 @@ abstract class AnyAssertionsSpec(
                         expectSubject.isNoneOfFun(1, arrayOf(2))
                     }.toThrow<AssertionError> {
                         message {
-                            contains(IS_NONE_OF.getDefault(), "${listBulletPoint}1")
+                            containsRegex(
+                                "\\Q$rootBulletPoint${IS_NONE_OF.getDefault()}\\E:.*$separator" +
+                                    "$indentBulletPoint${listBulletPoint}1"
+                            )
```
jGleitz, line 189: This is super hard to parse for a human.  Some time ago, I wouldn’t even have known what `\\Q` and `\\E` do.

Can we split this up into multiple, simpler `contains` statement? That would be easier to read both in code and in reporting.

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
```diff
@@ -746,6 +756,50 @@ abstract class AnyAssertionsSpec(
         }
     }
 
+    prefixedDescribe("because") {
+        val becauseFun = because.lambda
+        val becauseFunForInt = becauseInt.lambda
+
+        fun Expect<String>.containsBecause(reason: String) =
+            contains.exactly(1).regex("$separator\\Q${informationBulletPoint}${BECAUSE.getDefault().format(reason)}\\E")
```
jGleitz, line 764: Why is this a regex in the first place? is `$separator` a regex?

translations/de_CH/atrium-translations-de_CH-common/src/main/kotlin/ch/tutteli/atrium/translations/DescriptionAnyAssertion.kt
```diff
@@ -12,5 +12,6 @@ enum class DescriptionAnyAssertion(override val value: String) : StringBasedTran
     IS_A("ist eine Instanz vom Typ"),
     IS_SAME("ist dieselbe Instanz wie"),
     IS_NOT_SAME("ist nicht dieselbe Instanz wie"),
-    IS_NONE_OF("ist nicht")
+    IS_NONE_OF("ist nicht"),
+    BECAUSE("weil")
```
jGleitz, line 16: I think this should be ‘denn’, because ‘denn’ can be continued with the same grammatical structure as ‘because’. I fear with ‘weil’ we’ll get a lot of „weil: ? sind nicht in Dateinamen erlaubt“, which hurts!

jGleitz (changes requested):
I see some minor issues.

The changes we discussed in https://github.com/robstoll/atrium-roadmap/issues/91 are not in here on purpose, because we want to do them en bloque, am I right?

translations/de_CH/atrium-translations-de_CH-common/src/main/kotlin/ch/tutteli/atrium/translations/DescriptionAnyAssertion.kt
robstoll: good with me, going to change it

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
robstoll: Actually, does not need to be one in this case as we don't deal with multi line with stuff inbetween as in other places. Already though about simplifying it myself but then did not bother. Going to improve it :+1:

robstoll:
an to answer your question:
> The changes we discussed in robstoll/atrium-roadmap#91 are not in here on purpose, because we want to do them en bloque, am I right?

Yes, I prefer to refactor AssertionGroups later on

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/AnyAssertionsSpec.kt
robstoll: I really want to make sure the formatting is correct, hence we need to stick to this.

jGleitz (changes requested):
I’d like to revisit the README

jGleitz:
@robstoll I wrote a loooong comment about the README in my second review, but the comment somehow vanished, probably because of your force push. Do you have any idea whether we can recover it?

jGleitz:
Hu, your force push was two ours ago. But where’s my comment?

jGleitz (commented):
test

README.md
```diff
@@ -1541,6 +1542,148 @@ expected that subject: /tmp/atrium-path/directory/subfolder/file        (sun.nio
...
+
+Following a checklist as a rule of thumb:
+1. check if there is not already a predefined assertion function
+2. use a [feature assertion](#feature-assertions) instead.
+3. write an own assertion function (e.g. a [boolean based Assertions](#boolean-based-assertions)).
+4. [let us know](https://github.com/robstoll/atrium/issues/new?template=feature_request.md&title=using because) why you still want to use `because` - maybe Atrium lacks another feature
+5. use `because`
+
+</details>
+
```
jGleitz, line 1686: Okay, so I’ll rewrite my comment about this block. You’ll have to imagine that the initial one was a lot more eloquent :laughing: 

I don’t like this block in its current form. I think I know where you are coming from: You don’t want users to degenerate Atrium into a boolean matching function. I get the thought. Somebody coming from the omnipresent JUnit who is used to writing 

```Java
assertFalse(list.isEmpty(), "found no matching elements!")
```

might indeed get the idea to write 

```
expect(list.isEmpty()).because("found no matching elements") { toBe(false) }
```

_However_, I think there is a strong reason for every user not to write something like that: _The rest of the README_. I can’t imagine somebody learning about `because` before having seen how to use Atrium correctly before. I also don’t think that anybody would voluntarily write something as ugly as that, once they have learned how to do it correctly.

With that in mind, I see the following disadvantages with the README block in its current form:

 1. One should _never_ give negative examples. Never. First, study after study has shown that negative examples can often lead to the opposite effect of what was intended. Second, people scan through READMEs and copy the first block of code they see, without reading the surrounding text.
 2. The block is quite long for something that could be summarized as ‘please use Atrium according to the very README you are currently reading’
 3. I think the block makes it sound like `because` is a last resort. But it isn’t. `because` is for documenting non-obvious requirements. This use case clearly distinguishable from building a missing expectation function.
 4. I have a feeling that it is not easy to understand what the block is trying to tell us if one is not reading carfully.


I’d like us to try to reduce this block to a handful of sentences. Something among the lines of:

> 💬  Use `because` only to give reasons for non-obvious expectations.

> `because` can be a useful tool for explaining why there is a certain expectation. Sometimes, it is not directly obvious why we should expect something. In such cases, using `because` can make your code and your error messages easier to understand for other developers (including yorself in three months).

> Having said that, you should _not_ use `because` if you are missing a specific predefined expectation function. You can use [feature extractors](#feature-assertions), write [your own expectation function](#boolean-based-assertions) or [propose an addition to Atrium](https://github.com/robstoll/atrium/issues/new?template=feature_request.md&title=Missing%20Expectation%20Function) in such cases. 

> Just like code comments, `because` can be helpful, but should not be overused.



I have the strong feeling that my wording was a lot better in lost comment … :woman_shrugging:

robstoll:
> Hu, your force push was two hours ago. So that probably wasn’t the reason. But where’s my comment?

no idea, my first thought was, that you made it on the commit instead on the PR (so it would be here: https://github.com/robstoll/atrium/commit/9fa38ac23b67daec4408241d89750abd26e188bd) but I don't see it. Anyway, you wrote it again...

README.md
robstoll: @jGleitz note that this information is inside a details block which one needs to expand first (see https://github.com/robstoll/atrium/tree/%23660-because#attach-a-reason). So most users will not read it anyway... especially not the ones skimming through the README to copy & paste code. But I am fine with changing this to your text. I figured it is more convincing if you see the results but yeah... in the end I don't really care if people I am not working with are using it differently than I would. Hence we can also delete the whole block. What do you think?

jGleitz: I'm aware that it is in a details block, however, this address only my first point, if at all. So I think changing it will be an improvement 👍

robstoll: OK, if I get you right, then you vote for keeping the details block but improve it. right?

robstoll: @jGleitz In both cases (keep and improve or delete), I let you do the adjustments and in the same go, you can improve the samples in api-fluent and api-infix :wink:

jGleitz: >  OK, if I get you right, then you vote for keeping the details block but improve it. right?

Exactly.

> I let you do the adjustments

will do!

robstoll:
@jGleitz I am going to merge this PR as I want to use InformationType in other places. I created an issue instead which I assigned to you.

PR #730, opened 2020-12-22
build everything on github actions

robstoll:
Since travis starts to be unreliable with its credit system.

### DIFF
--- a/.github/workflows/bbc-ubuntu.yml
+++ b/.github/workflows/bbc-ubuntu.yml
@@ -1,0 +1,28 @@
+name: BBC Ubuntu
+
+on: [push, pull_request]
+
+jobs:
+    build:
+        env:
+            JAVA_VERSION: '11'
+        runs-on: ubuntu-latest
+
+        steps:
+            -   uses: actions/checkout@v2
+            -   name: "Set up JDK ${{ matrix.java_version }}"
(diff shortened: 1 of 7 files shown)


.github/workflows/bbc-ubuntu.yml
```diff
@@ -0,0 +1,28 @@
+name: BBC Ubuntu
```
jGleitz, line 1:
```suggestion
name: Backwards-Backwards Compatibility Ubuntu
```

When I started developing on Atrium, it took me a long time to figure out what a British TV network has to do with my Pull Request :wink: . I think giving this a proper name helps contributors.

.github/workflows/bc-ubuntu.yml
```diff
@@ -0,0 +1,28 @@
+name: BC Ubuntu
```
jGleitz, line 1:
```suggestion
name: Backwards Compatibility Ubuntu
```

.github/workflows/bbc-ubuntu.yml
robstoll: Fair enough :)

robstoll: btw. it means binary backward compatiblity (fixed it directly in master)

PR #743, opened 2020-12-23
Build Improvements

jGleitz:
I have a notebook running Ubuntu on a 3 year old Intel notebook processor with 2 physical / 4 virtual cores.
This PR improves the build on my machine. I tested with OpenJDK 11 with J9.

 * mockk hung the file system tests on my machine. I don’t know why. I removed mockk, which fixed the issue.
 * I enabled Gradle’s parallel build feature, which gives me build times of ~5m10s instead of ~6m20s, so a reduction of 1m10s. On machines with more cores, the the speedup should be even more pronounced.
 * I fixed the configuration of the Node.JS dependency copy task. The missing dependencies only caused a problem in a parallel build.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -444,6 +444,7 @@ def createJsTestTask(String... subprojectNames) {
             from compileKotlin2Js.destinationDir
 
             prefixedProject('verbs-internal-js').afterEvaluate {
+                // TODO 0.16.0 or 0.17.0, check if still required with the new Kotlin MPP plugin
                 configurations.testRuntimeClasspath.allDependencies.withType(ProjectDependency).each {
                     dependsOn(it.dependencyProject.assemble)
                 }
--- a/build.gradle
+++ b/build.gradle
@@ -444,6 +444,10 @@ def createJsTestTask(String... subprojectNames) {
             from compileKotlin2Js.destinationDir
 
             prefixedProject('verbs-internal-js').afterEvaluate {
+                configurations.testRuntimeClasspath.allDependencies.withType(ProjectDependency).each {
+                    dependsOn(it.dependencyProject.assemble)
+                }
+
                 configurations.testRuntimeClasspath.each {
                     from zipTree(it.absolutePath).matching { include '*.js', '*.js.map' }
                 }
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -444,6 +444,10 @@ def createJsTestTask(String... subprojectNames) {
             from compileKotlin2Js.destinationDir
 
             prefixedProject('verbs-internal-js').afterEvaluate {
+                configurations.testRuntimeClasspath.allDependencies.withType(ProjectDependency).each {
```
robstoll, line 447: Can you please add a TODO for me, I guess this should no longer be necessary once we switched to the new Kotlin MPP plugin
```suggestion
                // TODO 0.16.0 or 0.17.0, check if still required with the new Kotlin MPP plugin
                configurations.testRuntimeClasspath.allDependencies.withType(ProjectDependency).each {
```

logic/atrium-logic-jvm/src/test/kotlin/ch/tutteli/atrium/logic/impl/creating/filesystem/hints/SymbolicLinkResolvingSpec.kt
```diff
@@ -1,3 +1,6 @@
+//TODO remove file with 1.0.0
```
robstoll, line 1: Why do you suggest to remove the file with 1.0.0?

gradle.properties
```diff
@@ -1 +1,2 @@
 kotlin.code.style=official
+org.gradle.parallel=true
```
robstoll, line 2: I guess this causes the build to fail because multiple projects share the same .nodejs folder and now with parallel they get into conflicts. You can change this in build.gradle => search for workDir, make it per project
Might be you have to do the same for the npmWorkDir

logic/atrium-logic-jvm/src/test/kotlin/ch/tutteli/atrium/logic/impl/creating/filesystem/hints/SymbolicLinkResolvingSpec.kt
jGleitz: that’s a copy & paste error.

robstoll:
closing this for now as part of it was already merged and I guess it will take some time until you get the other part

PR #745, opened 2020-12-23
Remove Mockk

jGleitz:
This is #743 but without the parallel build. Debugging the build logic to figure out what is wrong will take longer, so I think it’s worth getting these changes in first.

### DIFF
--- a/logic/atrium-logic-jvm/src/test/kotlin/ch/tutteli/atrium/logic/impl/creating/filesystem/hints/SymbolicLinkResolvingSpec.kt
+++ b/logic/atrium-logic-jvm/src/test/kotlin/ch/tutteli/atrium/logic/impl/creating/filesystem/hints/SymbolicLinkResolvingSpec.kt
@@ -18,10 +18,7 @@ import ch.tutteli.niok.createDirectory
 import ch.tutteli.niok.createFile
 import ch.tutteli.niok.createSymbolicLink
 import ch.tutteli.spek.extensions.memoizedTempFolder
-import io.mockk.confirmVerified
-import io.mockk.every
-import io.mockk.mockk
-import io.mockk.verify
+import com.nhaarman.mockitokotlin2.*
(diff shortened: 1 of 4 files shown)


logic/atrium-logic-jvm/src/test/kotlin/ch/tutteli/atrium/logic/impl/creating/filesystem/hints/SymbolicLinkResolvingSpec.kt
```diff
@@ -1,3 +1,6 @@
+//TODO remove file with 1.0.0
+@file:Suppress("DEPRECATION")
```
robstoll, line 2: remove

robstoll (changes requested):
Just remove the copy& paste error and it looks good

logic/atrium-logic-jvm/src/test/kotlin/ch/tutteli/atrium/logic/impl/creating/filesystem/hints/SymbolicLinkResolvingSpec.kt
jGleitz: :facepalm:

jGleitz: Done.

robstoll:
@jGleitz do you want to remove mockk in other places as well?

robstoll (approved):

jGleitz:
>  do you want to remove mockk in other places as well?

I tried. But in one instance, we use `mockk` in a commons project. I saw no easy solution for that, so I left it. Do you have a suggestion?

robstoll:
Unfortunately not no, probably the reason why we stuck with it even though my last knowledge is that it is not further developed

PR #749, opened 2020-12-23
Check in CI that the README was Re-Generated

jGleitz:


robstoll (approved):

robstoll:
Actually, we cannot/should not add this check and I remember why I have not establish it in the first place. Contributors using a different OS than ubuntu would always get this error.

robstoll (changes requested):
Therefore I suggest we either close this PR or we tweak the report of Path assertions that it outputs always the same paths which is quite a hack though

jGleitz:
Ah, I remember. 

How about this: the CI always updates the `README` on `master` after pushes to master?

robstoll:
Why not, but I prefer if a PR is created if needed instead of directly pushing to master

robstoll:
Implemented the creation of a PR in case of a diff in readme (see https://github.com/robstoll/atrium/pull/757). therefore closing this

PR #758, opened 2020-12-27
677 add samples iterator assertions api infix

lilimapradhan9:


robstoll (approved):
@lilimapradhan9 nice, thanks for adding the additional comments, going to re-use it for the fluent-api as well

Issue #724, opened 2020-12-20
show only failing sub assertions in a summary assertion group


robstoll:
*Platform* (all, jvm, js): all
*Extension* (none, kotlin 1.3): none

## Code related feature
Consider the following reporting
```
expected that subject: {a=null, null=1, b=2}        (java.util.LinkedHashMap <1031246792>)
◆ contains only, in order: 
  ✔ ▶ element 0: a=null        (java.util.LinkedHashMap.Entry <1834736756>)
      ◾ ▶ key: "a"        <528658809>
          ◾ equals: "a"        <528658809>
      ◾ ▶ value: null
          ◾ equals: null
  ✘ ▶ element 1: null=1        (java.util.LinkedHashMap.Entry <601057428>)
      ◾ ▶ key: null
          ◾ equals: null
      ◾ ▶ value: 1        (kotlin.Int <1614645392>)
          ◾ is less than: 1        (kotlin.Int <1614645392>)
```
Atrium could be more helpful in spotting the error for `element: 1` by only showing the failing assertions:
```
  ✘ ▶ element 1: null=1        (java.util.LinkedHashMap.Entry <601057428>)
      ◾ ▶ value: 1        (kotlin.Int <1614645392>)
          ◾ is less than: 1        (kotlin.Int <1614645392>)
```
Also, I think it would be enough if the successful assertions only show that an element was successful but not the assertions behind it. So after all improvements it would look like the following:
```
expected that subject: {a=null, null=1, b=2}        (java.util.LinkedHashMap <1031246792>)
◆ contains only, in order: 
  ✔ ▶ element 0: a=null        (java.util.LinkedHashMap.Entry <1834736756>)
  ✘ ▶ element 1: null=1        (java.util.LinkedHashMap.Entry <601057428>)
      ◾ ▶ value: 1        (kotlin.Int <1614645392>)
          ◾ is less than: 1        (kotlin.Int <1614645392>)
```

In the spirit of #292, it would make sense that one can configure the assertion function to show all assertions or not within the summary assertion group. But per default, sub assertions within an summary assertion group should only be shown if failing.

jakubriegel:
Hi! I'll do it

robstoll:
@jakubriegel thanks but this isn't an easy one and I will tackle it myself in the context of another change. Please focus on another issue of you want to support Atrium

PR #768, opened 2021-01-06
Remove size check feature assertion from iterable contains

ivanmiklec:
I managed to remove size check feature from contains.in.order.only.grouped.entries/values but when testing ContainsInAnyOrderOnlyEntries/Values during mismatch or missing case only size or contains assertion is returned and I can't figure out where is it getting filtered out

### DIFF
--- a/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/iterable/contains/creators/impl/InAnyOrderOnlyAssertionCreator.kt
+++ b/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/iterable/contains/creators/impl/InAnyOrderOnlyAssertionCreator.kt
@@ -4,15 +4,14 @@ import ch.tutteli.atrium.assertions.Assertion
 import ch.tutteli.atrium.assertions.AssertionGroup
 import ch.tutteli.atrium.assertions.builders.assertionBuilder
 import ch.tutteli.atrium.assertions.builders.invisibleGroup
+import ch.tutteli.atrium.core.Some
 import ch.tutteli.atrium.creating.AssertionContainer
+import ch.tutteli.atrium.logic.*
 import ch.tutteli.atrium.logic.assertions.impl.LazyThreadUnsafeAssertionGroup
 import ch.tutteli.atrium.logic.creating.iterable.contains.IterableLikeContains
 import ch.tutteli.atrium.logic.creating.iterable.contains.searchbehaviours.InAnyOrderOnlySearchBehaviour
 import ch.tutteli.atrium.logic.creating.typeutils.IterableLike
-import ch.tutteli.atrium.reporting.Text
 import ch.tutteli.atrium.reporting.translating.Translatable
-import ch.tutteli.atrium.translations.DescriptionAnyAssertion.TO_BE
-import ch.tutteli.atrium.translations.DescriptionCollectionAssertion
 import ch.tutteli.atrium.translations.DescriptionIterableAssertion
 import ch.tutteli.atrium.translations.DescriptionIterableAssertion.*
 
@@ -43,22 +42,19 @@ abstract class InAnyOrderOnlyAssertionCreator<E, T : IterableLike, in SC>(
     ): AssertionGroup {
         return LazyThreadUnsafeAssertionGroup {
             val list = container.maybeSubject.fold({ mutableListOf<E?>() }) { converter(it).toMutableList() }
-            val actualSize = list.size
             val assertions = mutableListOf<Assertion>()
+            val sizeAssertion = container.collectBasedOnSubject(Some(list)) {
+                _logic
+                    .size { it }
+                    .collectAndLogicAppend { toBe(searchCriteria.size) }
+            }
 
             val mismatches = createAssertionsForAllSearchCriteria(container, searchCriteria, list, assertions)
-            val featureAssertions = createSizeFeatureAssertion(searchCriteria, actualSize)
             if (mismatches == 0 && list.isNotEmpty()) {
-                featureAssertions.add(LazyThreadUnsafeAssertionGroup {
+                assertions.add(LazyThreadUnsafeAssertionGroup {
                     createExplanatoryGroupForMismatchesEtc(list, WARNING_ADDITIONAL_ELEMENTS)
                 })
             }
-            assertions.add(
-                assertionBuilder.feature
-                    .withDescriptionAndRepresentation(DescriptionCollectionAssertion.SIZE, Text(actualSize.toString()))
-                    .withAssertions(featureAssertions)
-                    .build()
-            )
 
             val description = searchBehaviour.decorateDescription(CONTAINS)
             val summary = assertionBuilder.summary
@@ -73,12 +69,18 @@ abstract class InAnyOrderOnlyAssertionCreator<E, T : IterableLike, in SC>(
                 }
                 assertionBuilder.invisibleGroup
                     .withAssertions(
+                        sizeAssertion,
                         summary,
                         createExplanatoryGroupForMismatchesEtc(list, warningDescription)
                     )
                     .build()
             } else {
-                summary
+                assertionBuilder.invisibleGroup
+                    .withAssertions(
+                        sizeAssertion,
+                        summary
+                    )
+                    .build()
             }
         }
     }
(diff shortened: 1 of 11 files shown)


logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/iterable/contains/creators/impl/InAnyOrderOnlyAssertionCreator.kt
```diff
@@ -53,12 +53,11 @@ abstract class InAnyOrderOnlyAssertionCreator<E, T : IterableLike, in SC>(
                     createExplanatoryGroupForMismatchesEtc(list, WARNING_ADDITIONAL_ELEMENTS)
                 })
             }
-            assertions.add(
+            val sizeFeatureAssertion =
```
robstoll, line 56: Please use the `size` feature extractor from CollectingLike -> see ch.tutteli.atrium.logic.creating.maplike.contains.creators.impl.DefaultMapLikeContainsAssertions => line 194

Note that the additional entries should stay inside of `contains...` in reporting and no longer be under `size`
in the end we want to see something like:
```
expect: [1, 2]        (java.util.Arrays.ArrayList <410495873>)
◆ size: 2        (kotlin.Int <710239027>)
  ◾ to be: 1        (kotlin.Int <110771485>)
◆ contains only, in order: 
  ✔ ▶ entry 0: 1        (kotlin.Int <110771485>)
      ◾ to be: 1        (kotlin.Int <110771485>)
    ❗❗ additional entries detected: 
       ⚬ entry 1: 2        (kotlin.Int <710239027>)
```

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/IterableContainsInOrderOnlyGroupedEntriesAssertionsSpec.kt
```diff
@@ -267,17 +264,17 @@ abstract class IterableContainsInOrderOnlyGroupedEntriesAssertionsSpec(
                                 contains.exactly(1).value("$rootBulletPoint$containsInOrderOnlyGrouped:")
                                 indexFail(
                                     0, 1, listOf(1.0, 2.0),
+                                    sizeCheck(2, 2),
```
robstoll, line 267: note, if the size is correct, then it will no longer be shown the user as it is no longer part of the AssertionGroup with a SummaryAssertionGroupType (which has the effect that it shows also assertions which hold).
In short, here you should actually check that the size check does no longer show up. See ch.tutteli.atrium.specs.integration.MapContainsInAnyOrderOnlyKeyValuePairsAssertionsSpec => Line 110 as an example

robstoll (changes requested):

ivanmiklec:
Gonna work this out this weekend

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/iterable/contains/creators/impl/InAnyOrderOnlyAssertionCreator.kt
ivanmiklec: I got this result, but when encountering additional entries usually I get only size assertion printed out, unless I add size feature assertion like the old style, with only mismatches it works as expected

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/IterableContainsInOrderOnlyGroupedEntriesAssertionsSpec.kt
ivanmiklec: Managed to do that but in Grouped assertions it still prints out it out, in short i get something like this:
`◆ contains only, in order, grouped: 
  ✘ ▶ index 0..1: [null, 1.0]        (java.util.ArrayList.SubList <1355742676>)
      ◾ ▶ size: 2        (kotlin.Int <1639154810>)
          ◾ equals: 2        (kotlin.Int <1639154810>)
      ◾ contains only, in any order: 
        ✔ an element which: 
            » is: null
        ✘ an element which: 
            » is: null
        ❗❗ following elements were mismatched: 
           ⚬ 1.0        (kotlin.Double <459559002>)
  ✘ ▶ index 2..3: [null, 3.0]        (java.util.ArrayList.SubList <1533221659>)
      ◾ ▶ size: 2        (kotlin.Int <1639154810>)
          ◾ equals: 2        (kotlin.Int <1639154810>)
      ◾ contains only, in any order: 
        ✔ an element which: 
            » is less than: 5.0        (kotlin.Double <1702487214>)
        ✘ an element which: 
            » is greater than: 2.0        (kotlin.Double <1123976584>)
        ❗❗ following elements were mismatched: 
           ⚬ null)`

robstoll: That's fine for the moment. The grouped assertions use itself SummaryAssertionGroupType (thus the ✘ in front of index...) and as they are implemented currently, they show all sub-assertions in case they fail (see also https://github.com/robstoll/atrium/issues/724)

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/iterable/contains/creators/impl/InAnyOrderOnlyAssertionCreator.kt
robstoll: I leave a review comment. You need to wrap size and contains assertion in an invisible assertion group. This way they are evaluated together.

ivanmiklec:
Managed to get it working, will rebase tommorow, figured out I need to use something else than `assertionCollector` with new changes

robstoll:
There is collectBasedOnSubject on container (see DefaultMapContainsAssertions)

ivanmiklec:
There are some backwards compatibility failing checks, will investigate tommorow, also do you want the readme to be updated in this PR or later on (asking because of version releases)

robstoll:
That's OK, it's a change we accept. So what you need to do is to forgive those tests in settings.gradle.kts
Add `IterableContainsInAnyOrderOnlyElementsOfAssertionSpec.*error cases.*` to 0.14.0 and 0.15.0 I guess this should do it. You can check it by running: `BC=1 ./gradlew bcTests bbcTests`

Regarding the README, in case you are using an ubuntu, then you can run `./gradlew :readme-examples:build` which will update the examples in the readme but it's not necessary. There will be a PR created automatically with the necessary changes after I merge your changes in case you don't do it yourself.

settings.gradle.kts
```diff
@@ -19,6 +19,8 @@ buildscript {
                 or(
                     // improved reporting
                     "IterableContainsInOrderOnly.*Spec",
+                    "IterableContainsInAnyOrder.*error cases.*",
+                    "IterableContainsInAnyOrderOnlyEntriesAssertionsSpec.*nullable cases.*",
```
robstoll, line 23: Sorry, I informed you wrong, there is no need for the .* at the end as there is one already in line 32
```suggestion
                    "IterableContainsInAnyOrder.*error cases",
                    "IterableContainsInAnyOrderOnlyEntriesAssertionsSpec.*nullable cases",
```

settings.gradle.kts
```diff
@@ -60,6 +62,9 @@ buildscript {
             // forgive for bc and bbc
             ("(ch/tutteli/atrium/api/(fluent|infix)/en_GB/" +
                 or(
+                    "IterableContainsInOrderOnly.*Spec",
+                    "IterableContainsInAnyOrder.*error cases.*",
+                    "IterableContainsInAnyOrderOnlyEntriesAssertionsSpec.*nullable cases.*",
```
robstoll, line 67:
```suggestion
                    "IterableContainsInAnyOrder.*error cases",
                    "IterableContainsInAnyOrderOnlyEntriesAssertionsSpec.*nullable cases",
```

settings.gradle.kts
```diff
@@ -60,6 +62,9 @@ buildscript {
             // forgive for bc and bbc
             ("(ch/tutteli/atrium/api/(fluent|infix)/en_GB/" +
                 or(
+                    "IterableContainsInOrderOnly.*Spec",
```
robstoll, line 65: if it is not too complicated to narrow the scope a bit, then please do so. If basically every test case is failing, then you can keep it this way

robstoll (approved):

settings.gradle.kts
ivanmiklec: sure will check it, managed to narrow it to `IterableContainsInOrderOnly.*error cases `

robstoll:
@ivanmiklec nice work, thank you :+1:

Issue #636, opened 2020-10-17
improve performance of containsNoDuplicates


robstoll:
*Platform* (all, jvm, js, android): all
*Extension* (none, kotlin 1.3): none

## Code related feature
We currently use the following algorithm to filter all elements of a list which have duplicates:
```
list.count { e -> e == element } > 1
``` 
Which means in the worst case we iterate the whole list for every element in the list. I assume there are better ways to do this. Research and improve the algorithm.

TODO:
- adjust the algorithm in DefaultIterableLikeAssertions

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime, we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/master/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

jGleitz:
> I assume there are better ways to do this. 

Iterating over the collectiong while building a HashSet of the elements and adding any element that already is in the HashSet to the result collection has an expected runtime complexity of O(n) as opposed to the current O(n²) (assuming a O(1) implementation of `hashCode`/`equals`).

robstoll:
@jGleitz I did on purpose not give any further hints because it's hacktoberfest and a good issue for newbies. If you don't mind, delete your comment. You can still give the hint in case the one tackling the problem needs help 😉

jGleitz:
sorry, I didn’t get that :facepalm:. I was only trying to be helpful!

robstoll:
Always appreciated 👍

mohammaddvi:
Hi, I want to work on this issue and have a question and an idea.

robstoll:
Go ahead, what's your question?

mohammaddvi:
I`ve done actually but I want to be sure about my implementation. This method just wants to know there is duplicate or not, yes?

robstoll:
Please push what you already have, (e.g. as draft PR) it's always easier to talk about code when one sees it.

shytnik-igor:
I'll work on this

shytnik-igor:
@robstoll, What else can I do about this issue?

[...]

PR #812, opened 2021-02-15
#636 improve perfomance  filter duplicates

shytnik-igor:
I have improved the algorithm search of duplications.
We iterate over the list until the first equality of elements, where the elements' indices are not equal.

### DIFF
--- a/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
+++ b/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
@@ -176,13 +176,9 @@ class DefaultIterableLikeAssertions : IterableLikeAssertions {
     ): Assertion = LazyThreadUnsafeAssertionGroup {
         val list = transformToList(container, converter)
 
-        val duplicates = createIndexAssertions(list) { (index, element) ->
-            list.forEachIndexed { elementIndex, elementValue ->
-                if(elementValue == element && elementIndex != index) {
-                    return@createIndexAssertions true
-                }
-            }
-            return@createIndexAssertions false
+        val lookupHashSet = HashSet<E>()
+        val duplicates = createIndexAssertions(list) { (_, element) ->
+            !lookupHashSet.add(element)
         }
         createHasElementPlusFixedClaimGroup(
             list,
--- a/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
+++ b/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
@@ -176,8 +176,13 @@ class DefaultIterableLikeAssertions : IterableLikeAssertions {
     ): Assertion = LazyThreadUnsafeAssertionGroup {
         val list = transformToList(container, converter)
 
-        val duplicates = createIndexAssertions(list) { (_, element) ->
-            list.count { e -> e == element } > 1
+        val duplicates = createIndexAssertions(list) { (index, element) ->
+            list.forEachIndexed { elementIndex, elementValue ->
+                if(elementIndex != index && elementValue == element) {
+                    return@createIndexAssertions true
+                }
+            }
+            return@createIndexAssertions false
         }
         createHasElementPlusFixedClaimGroup(
             list,
--- a/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
+++ b/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
@@ -178,7 +178,7 @@ class DefaultIterableLikeAssertions : IterableLikeAssertions {
 
         val duplicates = createIndexAssertions(list) { (index, element) ->
             list.forEachIndexed { elementIndex, elementValue ->
-                if(elementIndex != index && elementValue == element) {
+                if(elementValue == element && elementIndex != index) {
                     return@createIndexAssertions true
                 }
             }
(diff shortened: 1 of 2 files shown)


logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
```diff
@@ -176,8 +176,13 @@ class DefaultIterableLikeAssertions : IterableLikeAssertions {
     ): Assertion = LazyThreadUnsafeAssertionGroup {
         val list = transformToList(container, converter)
 
-        val duplicates = createIndexAssertions(list) { (_, element) ->
-            list.count { e -> e == element } > 1
+        val duplicates = createIndexAssertions(list) { (index, element) ->
```
robstoll, line 179: We can do even better here. We need to refactor a bit tough. @jGleitz would you mind giving @shytnik-igor a few hints? Just if you have time, otherwise I'll do it

jGleitz: I can try.

@shytnik-igor, your change does indeed improve performance in some cases, because we stop searching once we have found the element. The previous implementation _always_ required a full pass over the list for every element.

Note, however, that your change is only an improvement if there are duplicates. In the expected case, which is that there are no duplicates, we still perform O(n²) operations. Even if there are duplicates, I’d expect that there are usually significantly fewer duplicates than non-duplicates. So in the grand scheme of things, your change does not improve performance _significantly_, and the worst-case performance stays at O(n²).

It is clear that we will need to check every for every list element whether it is a duplicate. So we cannot get better than O(n). However, we should be able to improve the performance of the lookup whether an element is already in the list. For example, hash tables like `HashMap` and `HashSet` have an expected lookup performance of O(1), which is, of course, much better than doing an O(n) linear search.

So what we are looking for is a solution that uses an efficient lookup data structure (like a hash table). With it, we should be able to achieve an expected worst-case performance of O(n).

I did intentionally not go into more detail, because I do not want to spoil the joy of finding the solution yourself. However, I am happy to elaborate on how I would approach this, if you like.

shytnik-igor: If I understand the idea correctly:

1. Prepare from the list a hash table for lookup.
2. Search a duplicate in the hash table.

@jGleitz  Is it right?

jGleitz: Yes, that is how I would approach it.

jGleitz: Note that it might be possible to populate the hash table while you are searching, which might be a little more elegant.

shytnik-igor: Thx for the help. I will try.

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
```diff
@@ -176,8 +176,9 @@ class DefaultIterableLikeAssertions : IterableLikeAssertions {
     ): Assertion = LazyThreadUnsafeAssertionGroup {
         val list = transformToList(container, converter)
 
+        val lookupHashSet = HashSet<E>()
         val duplicates = createIndexAssertions(list) { (_, element) ->
-            list.count { e -> e == element } > 1
+            !lookupHashSet.add(element)
```
jGleitz, line 181: elegant! :+1:

shytnik-igor: But tests fail

jGleitz:
The tests fail because they expect _all_ occurrences of duplicated objects to be reported, not just the additional occurrences.

I’d vote to keep the new behaviour and adapt the tests because I think the shorter report is better, as the longer report doesn’t contain useful information. If any additional information is needed, I’d rather add the index of the first occurrence to the messages of the later occurrences than have more report entries.

@robstoll what do you think?

robstoll:
Fine with me or even better in case you want to do a bit more @shytnik-igor, we could improve the report to something like (open for other ideas):
```
expected that subject: [1, 2, 1, 2, 3, 4, 4, 4]        (java.util.Arrays.ArrayList <114044797>)
    ◆ has not: duplicate elements
      ⚬ index 0: 1          (kotlin.Int <972231853>)
         >> duplicated by index: 2
      ⚬ index 2: 2        (kotlin.Int <525432111>)
          >> duplicated by index: 3
      ⚬ index 5: 4        (kotlin.Int <1676423953>)
          >> duplicated by index: 6, 7
```

shytnik-igor:
> Fine with me or even better in case you want to do a bit more @shytnik-igor, we could improve the report to something like (open for other ideas):
> 
> ```
> expected that subject: [1, 2, 1, 2, 3, 4, 4, 4]        (java.util.Arrays.ArrayList <114044797>)
>     ◆ has not: duplicate elements
>       ⚬ index 0: 1          (kotlin.Int <972231853>)
>          >> duplicated by index: 2
>       ⚬ index 2: 2        (kotlin.Int <525432111>)
>           >> duplicated by index: 3
>       ⚬ index 5: 4        (kotlin.Int <1676423953>)
>           >> duplicated by index: 6, 7
> ```

shytnik-igor:
> Fine with me or even better in case you want to do a bit more @shytnik-igor, we could improve the report to something like (open for other ideas):
> 
> ```
> expected that subject: [1, 2, 1, 2, 3, 4, 4, 4]        (java.util.Arrays.ArrayList <114044797>)
>     ◆ has not: duplicate elements
>       ⚬ index 0: 1          (kotlin.Int <972231853>)
>          >> duplicated by index: 2
>       ⚬ index 2: 2        (kotlin.Int <525432111>)
>           >> duplicated by index: 3
>       ⚬ index 5: 4        (kotlin.Int <1676423953>)
>           >> duplicated by index: 6, 7
> ```

Sorry, but I don't understand what I need to do now

robstoll:
No worries, I guess it's a bit too much to do the advanced reporting after a first good issue. I re-open the PR, I guess you closed it inadvertently. So what you need to do is adjust IterableExpectationsSpec => line 148. Remove the index/value pairs which no longer show up.

robstoll:
@shytnik-igor thanks for your first contribution to Atrium :tada: 

Don't forget to :star: Atrium and we would be happy if you tackle another issue:

PR #831, opened 2021-03-08
Added isEmptyDirectory to PathAssertions.kt and implemented it.

### LINKED ISSUE #766: Path.toBeAnEmptyDirectory

## Code related feature
```
expect(path).toBeAnEmptyDirectory()

//instead of

expect(path)
  .toBeADirectory()
  .feature("findFirstFile", { Files.list(this).use { it.findFirst() } }) {
    isEmpty()
  }
```

*logic*
- [ ] extend PathAssertions with a function `toBeEmptyDirectory` (see other functions as a guideline)
- [ ] implement toBeEmptyDirectory in DefaultPathAssertions.kt -- *bonus* use newDirectoryStream instead and list all files as failureHint in case the `iterator.hasNext()`

*api-fluent*
- [ ] provide a fun `toBeAnEmptyDirectory` in a new file called pathExpectations.kt (see for instance isDirectory in pashAssertions.kt as a guideline)
- [ ] add `@since 0.16.0` (adapt to current [milestone](https://github.com/robstoll/atrium/milestones)) to KDOC
- [ ] add toBeAnEmptyDirectory to PathExpectationsSpec in specs-common adopt the spec in api-fluent

*api-infix*
- [ ] provide a fun `toBe` which expects `anEmptyDirectory` in a new file called pathExpectations.kt (see `toBe aDirectory` in pathAssertions.kt as a guideline)
- [ ] add `@since 0.16.0` (adapt to current [milestone](https://github.com/robstoll/atrium/milestones)) to KDOC
- [ ] add toBeAnEmptyDirectory to PathExpectationsSpec in specs-common adopt the spec in api-infix

## Your first contribution?

mai2412:
Resolves #766. Given a path, isEmptyDirectory:
- Return isDirectory() if filesystem entry is not a directory.
- Otherwise, return assertion that directory is empty, if it's not, output any file in the directory as a hint as follows:
PathAssertionSamples > toNotBeEmptyDirectory() FAILED
    expected that subject: /var/folders/0d/4grl9hg92d1_6510b0c6sx3r0000gq/T/PathAssertionSamples16636861512034623567        (sun.nio.fs.UnixPath <760738946>)
    ◆is: an empty directory
      ⚬ directory has file: "/var/folders/0d/4grl9hg92d1_6510b0c6sx3r0000gq/T/PathAssertionSamples16636861512034623567/file <1953246438>

I named the function isEmptyDirectory instead of toBeEmptyDirectory to complement isDirectory

_logic_
* [x]  extend PathAssertions with a function `isEmptyDirectory` (see other functions as a guideline)
* [x]  implement isEmptyDirectory in DefaultPathAssertions.kt 

_api-fluent_
* [x]  provide a fun `isEmptyDirectory` in a new file called pathExpectations.kt (actually I just added it to pathAssertions as I'm not sure why it has to be in new file, let me know if this is necessary)
* [x]  add `@since 0.16.0` (adapt to current [milestone](https://github.com/robstoll/atrium/milestones)) to KDOC
* [x]  add isEmptyDirectory to PathExpectationsSpec in specs-common adopt the spec in api-fluent

_api-infix_
* [x]  provide a fun `toBe` which expects `anEmptyDirectory` in a new file called pathExpectations.kt (see `toBe aDirectory` in pathAssertions.kt as a guideline) - again I just added to pathAssertions
* [x]  add `@since 0.16.0` (adapt to current [milestone](https://github.com/robstoll/atrium/milestones)) to KDOC
* [x]  add isEmptyDirectory to PathExpectationsSpec in specs-common adopt the spec in api-infix


mai2412:
I haven't had much experience in kotlin so would appreciate any feedback.

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultPathAssertions.kt
```diff
@@ -194,4 +193,31 @@ class DefaultPathAssertions : PathAssertions {
...
+                                        .build()
+                                is Failure -> hintForIoException(it.path, it.exception)
+                            }
+                        }
+                        .withDescriptionAndRepresentation(DescriptionBasic.IS, AN_EMPTY_DIRECTORY)
+                        .build()
+                }
+        } else {
+            container.isDirectory()
+        }
```
jGleitz, line 222: If a file is not a directory, the current logic will check whether the file is a directory twice: First in the if-condition, and then when displaying the assertion created by `container.isDirectory()` to the user. This is wasteful in general, but even worse for file system operations: The file could be changed between checking the condition and testing the second assertion. In that case, users would get confusing messages.

Luckily, I think there is an easier solution here: First, you do not need to test whether the file is a directory. Instead, you can simply add the assertion for the file being a directory. If it’s not, the whole assertion will already fail. You then add the second assertion to the `AssertionContainer`, like you already do in the `let`-block. Only now you will need to be aware that the second assertion’s test may also be called if the first assertion fails. This is not too bad, though, you simply need to handle the corresponding exception.

jGleitz:
Even though you have checked “add isEmptyDirectory to PathExpectationsSpec in specs-common”, I don’t see any unit tests in the diff. Can you please provide appropriate unit tests for the `isEmptyDirectory` assertion?

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultPathAssertions.kt
```diff
@@ -194,4 +193,31 @@ class DefaultPathAssertions : PathAssertions {
                     container.resolve(entry).collect { _logicAppend { exists(NOFOLLOW_LINKS) } }
                 }
         ).build()
+
+    override fun <T : Path> isEmptyDirectory(container: AssertionContainer<T>): Assertion =
+        if (container.isDirectory().holds()) {
+            container.changeSubject.unreported {
+                it.runCatchingIo { Files.list(it).findAny() }
```
jGleitz, line 200: I think it would be more helpful to list all entries in the directory. This list can, of course, be pretty long, so maybe we should limit it to some reasonable number?

jGleitz:
@robstoll I like how @mai2412 lists the ‘bad’ files when the assertion fails. Nevertheless, I think the hints would be easier to read if the paths were relative to the directory. However, that would introduce non-standard reporting for paths. Would that be an issue for you?

mai2412:
> Even though you have checked “add isEmptyDirectory to PathExpectationsSpec in specs-common”, I don’t see any unit tests in the diff. Can you please provide appropriate unit tests for the `isEmptyDirectory` assertion?

Hey, could you point me to where the tests should reside?

jGleitz:
They should go into [`PathExpectationSpec`](https://github.com/robstoll/atrium/blob/master/misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathExpectationsSpec.kt). You will need to add a new parameter to the `PathExpectationsSpec`’s constructor, which takes the concrete implementation of the assertion. Then you can write tests using that parameter. Look at the existing code for reference.

You will need to pass your respective implementation to the constructor in the implementing types. Let the compiler errors and the existing code guide you.

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/pathAssertions.kt
```diff
@@ -447,3 +446,14 @@ infix fun <T : Path> Expect<T>.hasSameTextualContentAs(pathWithEncoding: PathWit
...
+
+/**
+ * Expects that the subject of `this` expectation (a [Path]) is an empty directory;
+ * meaning that there is a file system entry at the location the [Path] points to and that is an empty directory.
+ *
+ * @return an [Expect] for the subject of `this` expectation.
+ *
+ * @since 0.16.0
+ */
+infix fun <T : Path> Expect<T>.isEmptyDirectory(@Suppress("UNUSED_PARAMETER") aDirectory: aDirectory): Expect<T> =
```
robstoll, line 458: rename to 
```suggestion
infix fun <T : Path> Expect<T>.toBe(@Suppress("UNUSED_PARAMETER") anEmptyDirectory: anEmptyDirectory): Expect<T> =
```
and introduce anEmptyDirectory

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/pathAssertions.kt
```diff
@@ -379,3 +379,14 @@ fun <T : Path> Expect<T>.hasSameTextualContentAs(
...
 fun <T : Path> Expect<T>.hasSameBinaryContentAs(targetPath: Path): Expect<T> =
     _logicAppend { hasSameBinaryContentAs(targetPath) }
+
+/**
+ * Expects that the subject of `this` expectation (a [Path]) is an empty directory;
+ * meaning that there is a file system entry at the location the [Path] points to and that is an empty directory.
+ *
+ * @return an [Expect] for the subject of `this` expectation.
+ *
+ * @since 0.16.0
```
robstoll, line 389: Please add a sample to PathAssertionSamples and link it here via `@sample` (see for instance anyAssertions.kt)
Could you do me a favour and also link toBeASymbolicLink with the corresponding sample.

apis/infix-en_GB/atrium-api-infix-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/pathAssertions.kt
```diff
@@ -447,3 +446,14 @@ infix fun <T : Path> Expect<T>.hasSameTextualContentAs(pathWithEncoding: PathWit
...
 infix fun <T : Path> Expect<T>.hasSameBinaryContentAs(targetPath: Path): Expect<T> =
     _logicAppend { hasSameBinaryContentAs(targetPath) }
+
+/**
+ * Expects that the subject of `this` expectation (a [Path]) is an empty directory;
+ * meaning that there is a file system entry at the location the [Path] points to and that is an empty directory.
+ *
+ * @return an [Expect] for the subject of `this` expectation.
+ *
+ * @since 0.16.0
```
robstoll, line 456: same here, please add a sample

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultPathAssertions.kt
```diff
@@ -194,4 +193,31 @@ class DefaultPathAssertions : PathAssertions {
...
+                .let { expectResult ->
+                    assertionBuilder.descriptive.withTest(expectResult)
+                    { it is Success && it.value.isEmpty }
+                        .withFailureHintBasedOnDefinedSubject(expectResult) {
+                            when (it) {
+                                is Success ->
+                                    assertionBuilder.descriptive.failing
+                                        .withDescriptionAndRepresentation(
+                                            DIRECTORY_HAS_FILE,
+                                            it.value.orElseThrow().toString()
```
robstoll, line 211: leave out the toString, it is the responsibility of the ObjectFormatter how Paths are formatted:
```suggestion
                                            it.value.orElseThrow()
```

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultPathAssertions.kt
robstoll, line 200: As written in the issue description, you should use `.use{...}` in order that the stream is closed. 

IMO the current implementation is fine as first step. If you like, then improve it by showing more than just one file. maybe 7 and if there are more you use something like `...`. If you want to do this, then you need to switch from assertionBuilder.descriptive to assertionBuilder.list and then you can use assertionBuilder.representationOnly inside to list the different files.

robstoll (changes requested):

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultPathAssertions.kt
mai2412: It seems like the file could always be changed even if we have 2 assertions? What happens if the first assertion holds but by the time the second assertion's test is called, the path ceased to be a directory and we get an exception, we could either
- ignore this exception which means the user gets no error even though the path probably does not point to an empty directory
- return isDirectory() again, thought this presents the same issue as before so probably not
- build our own failing assertion and return it to the user, though I don't know how to format it in the same way as isDirectory()? If we do this I think we might as well not use isDirectory() and so that the error formatting is consistent.

Let me know if there's something I'm missing and/or which approach should I take.

jGleitz: None of are file assertions are atomic, that is true. However, the impact of parallel modifications can be more or less bad. I’d argue that the current implementation can lead to irritating results, while the implementations I have in mind and @robstoll suggested should not be a problem, because the user can understand the reporting and deduce what happened.

robstoll:
@mai2412 thanks for the tests and the samples. There are still some points which are open. Especially my comment about closing the stream by using `.use{...}` this is also the reason why the build fails

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultPathAssertions.kt
mai2412: The implementation now uses .use{...}

misc/specs/atrium-specs-jvm/src/main/kotlin/ch/tutteli/atrium/specs/integration/PathExpectationsSpec.kt
```diff
@@ -948,6 +950,54 @@ abstract class PathExpectationsSpec(
...
+        it("throws an AssertionError for a non-empty directory") withAndWithoutSymlink { maybeLink ->
+            val folder = maybeLink.create(tempFolder.newDirectory("notEmpty"))
+            maybeLink.create(tempFolder.newFile("notEmpty/a"))
+            expect {
+                expect(folder).isEmptyDirectoryFun()
+            }.toThrow<AssertionError>().message {
+                contains(expectedEmptyMessage, "${HAS.getDefault()}:")
+                contains("notEmpty/a")
+            }
+        }
```
jGleitz, line 993: Can you please add a test case where the directory only contains another (empty) directory?

jGleitz: The `toThrow` block should also contain the `containsExplanationFor(maybeLink)` part.

mai2412: I imagine if the directory contains another empty directory that is still not an empty directory correct (assertion will raise)?
If I include containsExplanationFor(maybeLink) the test fails, I'm not 100% sure why but I'm thinking it has to do with  how Files.list work. isEmptyDirectory does follow symbolic links.

jGleitz: > I imagine if the directory contains another empty directory that is still not an empty directory correct (assertion will raise)?

Yes, the assertion should fail for a directory that contains another empty directory.

> If I include containsExplanationFor(maybeLink) the test fails

Without seeing the error message, I can’t tell why the test fails. The error message should give you hints. However, I _suspect_ that your assertion does not report symlinks that have been resolved along the way. There is the helper function `explainForResolvedLink` (defined in `ch.tutteli.atrium.logic.creating.filesystem.hints.hits.kt`) which provides the functionality for explaining the symbolic link chain. Please use it for your assertion, as well. You can look at the other assertions to see how it should be used.

translations/en_GB/atrium-translations-en_GB-jvm/src/main/kotlin/ch/tutteli/atrium/translations/DescriptionPathAssertion.kt
```diff
@@ -24,6 +24,8 @@ enum class DescriptionPathAssertion(override val value: String) : StringBasedTra
     EXECUTABLE("executable"),
     A_FILE("a file"),
     A_DIRECTORY("a directory"),
+    AN_EMPTY_DIRECTORY("an empty directory"),
+    DIRECTORY_HAS("directory has"),
```
jGleitz, line 28: I feel that ‘directory contains’ would be the better wording.

robstoll: @mai2412 the build currently fails because there is no such entry defined for de_CH. I also suggest that you change this to `directory contains`
```suggestion
    DIRECTORY_CONTAINS("directory contains"),
```
And then add the two entries to translations/de_CH/atrium-translations-de_CH-jvm/src/main/kotlin/ch/tutteli/atrium/translations/DescriptionPathAssertion.kt where you can use the following:
```
 AN_EMPTY_DIRECTORY("ein leeres Verzeichnis"),
DIRECTORY_CONTAINS("Verzeichnis enthält"),
```

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultPathAssertions.kt
```diff
@@ -194,4 +193,32 @@ class DefaultPathAssertions : PathAssertions {
                     container.resolve(entry).collect { _logicAppend { exists(NOFOLLOW_LINKS) } }
                 }
         ).build()
+
+    override fun <T : Path> isEmptyDirectory(container: AssertionContainer<T>): Assertion {
+        val isDirectory = container.isDirectory()
+        if (isDirectory.holds()) {
+            return container.changeSubject.unreported {
+                Pair(it, it.runCatchingIo { Files.newDirectoryStream(it).use { stream -> stream.firstOrNull() } })
```
jGleitz, line 201: The `IoResult` returned by `runCatchingIo` has an attribute `path` that gives access to the path. Hence, it’s not necessary to create a pair with the path here. You can simply change the subject to the `IoResult`. I’d prefer that, since it simplifies the code.

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultPathAssertions.kt
```diff
@@ -194,4 +193,32 @@ class DefaultPathAssertions : PathAssertions {
...
+            }.let { expectResult ->
+                    assertionBuilder.descriptive.withTest(expectResult)
+                    { it.second is Success && (it.second as Success<Path?>).value == null }
+                        .withFailureHintBasedOnDefinedSubject(expectResult) {
```
jGleitz, line 205: I find this formatting irritating to read and would prefer to format this section like this:

```suggestion
                    assertionBuilder.descriptive.withTest(expectResult) {
                        it.second is Success && (it.second as Success<Path?>).value == null
                     }
                         .withFailureHintBasedOnDefinedSubject(expectResult) {
```

This would also be in line with the other code.

logic/atrium-logic-jvm/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultPathAssertions.kt
```diff
@@ -194,4 +193,32 @@ class DefaultPathAssertions : PathAssertions {
...
+                            explainForResolvedLink(it.first) { realPath ->
+                                when (it.second) {
+                                    is Success ->
+                                        assertionBuilder.descriptive.failing
+                                            .withDescriptionAndRepresentation(
+                                                DIRECTORY_CONTAINS,
+                                                (it.second as Success<Path?>).value!!
```
jGleitz, line 212: I think this line could be simplified to:

```suggestion
                                                (it.second as Success<Path>).value
```

jGleitz:
Thank you for the additional test cases. The PR now looks good to me except for:

 * minor nit-picks on the code (see my comments)
 * I’d prefer to report the existing path(s) relative to the subject (@robstoll you haven’t shared your thoughts on this yet, see [my comment](https://github.com/robstoll/atrium/pull/831#issuecomment-792665655))
 * I’d prefer to report the first 5-10 existing paths. Note that @robstoll [said this is optional](https://github.com/robstoll/atrium/pull/831#discussion_r589486043) and could be done in another PR. He’s right, I guess.

PR #847, opened 2021-03-27
Add samples for collectionAssertions of api-infix

### LINKED ISSUE #680: add samples for collectionAssertions of api-infix
*Platform* (all, jvm, js, android): all

## Code related feature
We also want to provide samples for the infix api

- [ ] copy CollectionAssertionsSample from fluent-api, adjust imports `...api.fluent...` to `...api.infix...` and ajdust the samples accordingly (always use the infix syntax, never the fluent).
- [ ] check that the sample in the KDoc Intellij shows looks good (ALT+Mouse-click on function name)

## Your first contribution?

szatyinadam:
Add samples for collectionAssertions of api-infix.

Fix #680 .

### DIFF
--- a/apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/collectionAssertions.kt
+++ b/apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/collectionAssertions.kt
@@ -10,6 +10,8 @@ import ch.tutteli.kbox.identity
  * @param empty Use the pseudo-keyword `empty`.
  *
  * @return an [Expect] for the subject of `this` expectation.
+ *
+ * @sample ch.tutteli.atrium.api.infix.en_GB.samples.deprecated.CollectionAssertionSamples.isEmpty
  */
 infix fun <T : Collection<*>> Expect<T>.toBe(@Suppress("UNUSED_PARAMETER") empty: empty): Expect<T> =
     _logicAppend { isEmpty(::identity) }
@@ -20,6 +22,8 @@ infix fun <T : Collection<*>> Expect<T>.toBe(@Suppress("UNUSED_PARAMETER") empty
  * @param empty Use the pseudo-keyword `empty`.
  *
  * @return an [Expect] for the subject of `this` expectation.
+ *
+ * @sample ch.tutteli.atrium.api.infix.en_GB.samples.deprecated.CollectionAssertionSamples.isNotEmpty
  */
 infix fun <T : Collection<*>> Expect<T>.notToBe(@Suppress("UNUSED_PARAMETER") empty: empty): Expect<T> =
     _logicAppend { isNotEmpty(::identity) }
@@ -30,6 +34,8 @@ infix fun <T : Collection<*>> Expect<T>.notToBe(@Suppress("UNUSED_PARAMETER") em
  * Shortcut for `size.toBe(expected)`.
  *
  * @return an [Expect] for the subject of `this` expectation.
+ *
+ * @sample ch.tutteli.atrium.api.infix.en_GB.samples.deprecated.CollectionAssertionSamples.hasSize
  */
 infix fun <T : Collection<*>> Expect<T>.hasSize(expected: Int): Expect<T> =
     size { toBe(expected) }
@@ -50,6 +56,8 @@ val <T : Collection<*>> Expect<T>.size: Expect<Int>
  * returns an [Expect] for the current subject of `this` expectation.
  *
  * @return an [Expect] for the subject of `this` expectation.
+ *
+ * @sample ch.tutteli.atrium.api.infix.en_GB.samples.deprecated.CollectionAssertionSamples.sizeFeature
  */
 infix fun <E, T : Collection<E>> Expect<T>.size(assertionCreator: Expect<Int>.() -> Unit): Expect<T> =
     _logic.size(::identity).collectAndAppend(assertionCreator)
--- a/apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/collectionAssertions.kt
+++ b/apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/collectionAssertions.kt
@@ -46,6 +46,8 @@ infix fun <T : Collection<*>> Expect<T>.hasSize(expected: Int): Expect<T> =
  * so that further fluent calls are assertions about it.
  *
  * @return The newly created [Expect] for the extracted feature.
+ *
+ * @sample ch.tutteli.atrium.api.infix.en_GB.samples.deprecated.CollectionAssertionSamples.sizeFeature
  */
 val <T : Collection<*>> Expect<T>.size: Expect<Int>
     get() = _logic.size(::identity).transform()
@@ -57,7 +59,7 @@ val <T : Collection<*>> Expect<T>.size: Expect<Int>
  *
  * @return an [Expect] for the subject of `this` expectation.
  *
- * @sample ch.tutteli.atrium.api.infix.en_GB.samples.deprecated.CollectionAssertionSamples.sizeFeature
+ * @sample ch.tutteli.atrium.api.infix.en_GB.samples.deprecated.CollectionAssertionSamples.size
  */
 infix fun <E, T : Collection<E>> Expect<T>.size(assertionCreator: Expect<Int>.() -> Unit): Expect<T> =
     _logic.size(::identity).collectAndAppend(assertionCreator)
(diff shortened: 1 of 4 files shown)


apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/samples/deprecated/CollectionAssertionSamples.kt
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    @Test
+    fun sizeFeature() {
+        expect(listOf(1, 2, 3)) size {
+            it isGreaterThan 1 // subject is now of type Int (actually 3)
+        } size {
+            it isLessThan 4 // subject is still of type Int (still 3)
+        }
```
robstoll, line 49: here we want to see the overload of size in action which does not expect an assertionCreator -lambda. It's a bit more tricky than in the api-fluent especially if we want to add all remarks. please rewrite it (and other places) to:
```suggestion
        expect(listOf(1, 2, 3)) {
            size /*                 subject is now of type Int (actually 3)
            */ isGreaterThan 1 /*   subject is still of type Int (still 3)
            */ isLessThan 2
        }
```
Btw. good catch that it is still 3 and not still 1 as written in the api-fluent example. Do you mind to fix it there as well?

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/samples/deprecated/CollectionAssertionSamples.kt
```diff
@@ -0,0 +1,83 @@
...
+            containsNot("${isGreaterThanDescr}: 4")
+        }
+    }
+
+    @Test
+    fun size() {
+        expect(listOf(1, 2, 3)) size { // subject inside this block is of type Int (actually 3)
+            it isGreaterThan 1
+            it isLessThan 4
+        }
```
robstoll, line 68: please include the rest of the sample as defined in api-fluent use the same workaround with `/* */` as shown above

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/samples/deprecated/CollectionAssertionSamples.kt
```diff
@@ -0,0 +1,83 @@
...
+
+        fails {
+            // all assertions are evaluated inside an assertion group block; for more details:
+            // https://github.com/robstoll/atrium#define-single-assertions-or-assertion-groups
+            expect(listOf(1, 2, 3)) size {
+                it isLessThan 1    // fails
+                it isGreaterThan 4 // still evaluated even though `isLessThan(1)` already fails,
+                // use `.size.` if you want a fail fast behaviour
```
robstoll, line 76: please indent further so that it is aligned with the other two comments

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/samples/deprecated/CollectionAssertionSamples.kt
```diff
@@ -0,0 +1,83 @@
...
+package ch.tutteli.atrium.api.infix.en_GB.samples.deprecated
+
+import ch.tutteli.atrium.api.infix.en_GB.*
+import ch.tutteli.atrium.api.infix.en_GB.samples.fails
+import ch.tutteli.atrium.api.verbs.internal.expect
+import ch.tutteli.atrium.translations.DescriptionComparableAssertion
+import kotlin.test.Test
+
+class CollectionAssertionSamples {
+    private val isLessThanDescr = DescriptionComparableAssertion.IS_LESS_THAN.getDefault()
```
robstoll, line 13: nice idea but one does not see the definition of this in the sample. Thus you need to repeat it in each sample instead. Also, newcomers to Atrium are probably more confused about  `DescriptionComparableAssertion.IS_LESS_THAN.getDefault()` than it helps. Thus I suggest you stick to "is less than" and "is greater than" like we have it in api-fluent

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/infix/en_GB/collectionAssertions.kt
```diff
@@ -50,6 +56,8 @@ val <T : Collection<*>> Expect<T>.size: Expect<Int>
  * returns an [Expect] for the current subject of `this` expectation.
  *
  * @return an [Expect] for the subject of `this` expectation.
+ *
+ * @sample ch.tutteli.atrium.api.infix.en_GB.samples.deprecated.CollectionAssertionSamples.sizeFeature
```
robstoll, line 60: this one should point to `CollectionAssertionSamples.size` and you forgot to add the link to the overload above

robstoll (changes requested):
@szatyinadam thanks for your work, a few improvements and we are good to merge

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/samples/deprecated/CollectionAssertionSamples.kt
szatyinadam: I'm not sure what are You asking for. Can You take a look at the updated tests?

This is the overloaded size:
```
expect(listOf(1, 2, 3)) { // subject is now of type Int (actually 3)
    size isGreaterThan 1 isLessThan 4
}
```
This is the other:
```
expect(listOf(1, 2, 3)) size {
    isGreaterThan(1) // subject inside this block is of type Int (actually 3)
} size {
    isLessThan(4) // subject inside this block is of type Int (actually 3)
}
```

robstoll: See the suggestion I have made for sizeFeature. The problem with your current implementation is, that the comment is wrong. The subject inside the block is still List. It changes after the `size` call. See my suggestion, it looks a bit bewildered with the `/* */` but its (currently) the only way to add comments per line within the infix syntax. Let me know in case this explanation is not enough.

robstoll: What I meant with use the /**/ workaround: your sample is missing the note about, subject is back to list. You can do the following:
```
} /* Here comes the comment
*/ size {
...
}
```

robstoll:
@szatyinadam I just remembered that we used another syntax in ListAssertionSamples for the comments which is superior to my suggestion. Please take a look and adjust your samples accordingly.

robstoll (dismissed):
Perfect, many thanks :+1: 
Going to wait for the build to pass.

szatyinadam:
`sizeFeature()` fails, because it seems it is using an assertion group block, and the `containsNot` message still presents. Can You help me about how to use size in a fail fast assertion style?

robstoll:
I see, that's a bummer, we currently don't allow fail fast within an assertion group block. What we can see here is a weakness of the infix-API I would say, maybe we should improve here. Anyway, go with the following for now:
```
        expect(listOf(1, 2, 3))
            .size isLessThan 1 isGreaterThan 4
            //|       |              | not reported because `isLessThan 1` already fails
            //|       | fails
            //| subject is now of type Int (actually 3)
```

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/samples/deprecated/ListAssertionSamples.kt
```diff
@@ -29,7 +29,7 @@ class ListAssertionSamples {
             // use `get index(elementIndex) { ... }` if you want that all assertions are evaluated
         } message {
             contains("index out of bounds")
-            containsNot("is less than: 2")
+            containsNot("is less than: 0")
```
robstoll, line 32: I guess this change was not intentional

robstoll (changes requested):

apis/infix-en_GB/atrium-api-infix-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/infix/en_GB/samples/deprecated/ListAssertionSamples.kt
robstoll: my bad it's actually a fix, thanks :slightly_smiling_face:

robstoll (approved):

robstoll:
@szatyinadam hm... the remaining failure is due to my suggestion to inline https://github.com/robstoll/atrium/pull/847#discussion_r602795994 I am sorry, looks like my memory of the infix API wasn't that good. Up to you if you want to change this again, I can also do it myself.

robstoll (approved):

robstoll:
@szatyinadam thanks for your first contribution to Atrium :slightly_smiling_face: :+1: 

We could use your help with other issues and don't forget to :star: Atrium if you liked it.

szatyinadam:
@robstoll Thanks for your patience, and I will definitely look for other issues 🙂

PR #874, opened 2021-04-20
renaming collectionAssertions to consistent `to + infitinive` naming schema

robstoll:


jGleitz (commented):
:+1:

PR #876, opened 2021-04-21
renaming comparableAssertions to consistent `to + infitinive` naming schema

robstoll:

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,7 +288,7 @@ The next section shows how you can define multiple assertions for the same subje
 
 ```kotlin
 // two single assertions, only first evaluated
-expect(4 + 6).isLessThan(5).isGreaterThan(10)
+expect(4 + 6).toBeLessThan(5).toBeGreaterThan(10)
 ```
 ↑ <sub>[Example](https://github.com/robstoll/atrium/tree/master/misc/tools/readme-examples/src/main/kotlin/readme/examples/MostExamplesSpec.kt#L24)</sub> ↓ <sub>[Output](#ex-single)</sub>
 <a name="ex-single"></a>
@@ -319,8 +319,8 @@ If you want that both assertions are evaluated together, then use the assertion 
 ```kotlin
 // assertion group with two assertions, both evaluated
 expect(4 + 6) {
-    isLessThan(5)
-    isGreaterThan(10)
+    toBeLessThan(5)
+    toBeGreaterThan(10)
 }
 ```
 ↑ <sub>[Example](https://github.com/robstoll/atrium/tree/master/misc/tools/readme-examples/src/main/kotlin/readme/examples/MostExamplesSpec.kt#L29)</sub> ↓ <sub>[Output](#ex-group)</sub>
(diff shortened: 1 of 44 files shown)


jGleitz (commented):
Since we have `toBeLessThan` and `toBeEqualComparingTo`, I think it should be called `toBeLessThanOrEqualTo` (and `toBeGreaterThanOrEqualTo`) to be consistent.

Since this is quite a mouth full, I propose that we change them to `toNotBeGreaterThan` and `toNotBeLessThan`, which is more compact.

robstoll:
I agree with the  ...To. I will change that.
I see the argument with the length. But most people tend to think in positive terms and especially in this case where most people think in terms of operators (I guess) it's better to have the longer form to resemble `<` `<=` `>=` `>`. Don't you think so?

jGleitz:
> But most people tend to think in positive terms and especially in this case where most people think in terms of operators.

I am quite used to using the „not greater than“ wording. But then again, I did some mathematical writing. There I quickly got used to writing ‘let x not be greater than y’ instead of ‘let x be less than or equal to y’, simply because it’s easier to read. But I don’t know whether my being familiar with it is representative.

robstoll:
I see your point. We could think about adding `notToBeGreaterThan` and `notToBeLessThan` additionally but I am not sure if it's worth the confusion it could potentially bring. At first glance I am in favour of adding it because depending on your use case it makes more sense to formulate it as / reads more naturally with `notToBeGreaterThan` than with `toBeLessThanOrEqualTo`

jGleitz:
I don’t think adding `notToBeLessThan` is a good solution. It violates [R9: One way to the goal](https://github.com/robstoll/atrium-roadmap/wiki/Requirements#r9-one-way-to-the-goal). If at all, I’d suggest that `notToBeLessThan` replaces `toBeGreaterThanOrEqualTo`. My reasoning is this:

 1. Having only one way to state the expectation is more important than having an expectation function that is 100% intuitive, from my point of view.
 2. Depending on the situation and the state of mind of the developer, it will sometimes feel more natural to write `toBeGreaterThanOrEqualTo`, and sometimes `notToBeLessThan`.
 3. There is no telling which will feel more natural more often.
 4. Given the above, there is no good reason to pick one over the other, so we should go with the shorter one (`notToBeLessThan`) because it is easier to read.

robstoll:
I thought about it a bit more. I agree regarding R9 and it's also the only reason why I did not straight away add `notToBeGreaterThan` and `notToBeLessThan`. I still considered it because I also thought about a different reporting, not just an alias. Yet, I think this is only really worth it once we have an HTML report and you would like to take it to talk about it with business people. Then it might be just better to have both variants. But that's far away... I initially thought I will have a big need for the HTML report but over these years it never was that crucial and it could be I'll never implement it. So.. yeah, I also value R9 higher.
Anyway... I made up my mind regarding which method name I want to take. I'll go with `notToBeLessThan` instead of `toBeGreaterThanOrEqualTo` but not primarily because it is shorter (that's more a nice to have IMO) but because:
- we often have `to...` and `notTo...` (e.g. toStartWith, notToStartWith) and thus this is more consistent
- what I personally never liked about `isLessThanOrEqual` is that one could believe it means `subject < expected || subject == expected` -- if the equals/compareTo contract is implemented correctly, then this does not matter. But it still has a bad aftertaste IMO. With notToBeLessThan we don't have that. At least in my mind, it is clearly a comparison related to compareTo

robstoll:
I just did the refactoring locally. Although, we have all the arguments for `notTo..`, I doubt this decision is newcomer friendly. I am pretty sure there will be a lot of people looking for a way to combine `toBeLessThan` and `toEqual` with an `or` operator which I definitely don't want to provide. Especially for newcomers or rather for programming beginners, it won't be obvious that they could search for `notTheBeGreaterThan` in such a case. I'll value [R7: Completeness](https://github.com/robstoll/atrium-roadmap/wiki/Requirements#r7-completeness) higher than R9 this time.

PR #889, opened 2021-04-28
renaming iterableAssertions to consistent `to + infitinive` naming schema

robstoll:


apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceContainsCheckers.kt
```diff
@@ -21,7 +21,7 @@ import kotlin.jvm.JvmName
  *
  * @return The newly created builder.
  * @throws IllegalArgumentException In case [times] is smaller than zero.
- * @throws IllegalArgumentException In case [times] equals to zero; use [containsNot] instead.
+ * @throws IllegalArgumentException In case [times] equals to zero; use [notToContain] instead.
```
jGleitz, line 24:
```suggestion
 * @throws IllegalArgumentException In case [times] is zero; use [notToContain] instead.
```

(or ‘equals zero’)

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceContainsCheckers.kt
```diff
@@ -39,7 +39,7 @@ fun <T : CharSequence, S : SearchBehaviour> CharSequenceContains.EntryPointStep<
...
```
jGleitz, line 42:
```suggestion
 * @throws IllegalArgumentException In case [times] is zero; use [notToContain] instead.
```

(or ‘equals zero’)

jGleitz (commented):
Looking at the code, I take issue with `toHaveNextAndAny` and `toHaveNextAndAll`.

First, when reading the function, I first did not understand what they do. My mind tried to parse them as ‘to have: (next and all {})’, which makes no sense. Only after backtracking I noticed that it’s meant to be ‘(to have next), (and all {})’.

Second, I generally expect an expectation function to ‘do one thing’. Yes, there are expectation functions that do other implied checks, but that is different from one function doing two inherently different things.

I was going to suggest that we split the two and let users drop in the `and` if they want, i.e. `toHaveNext.and.all {}`. However, I remembered that you are afraid of users using `all` without realizing that it does not imply a `toHaveNext` check. I still don’t share that fear, I think an `all` check that passes for empty collections is entirely reasonable. However, I don’t want to re-start the discussion.

So I guess all I can do is give feedback that I find `toHaveNextAndAny` irritating. I have no better solution that would not involve splitting the `toHaveNext` and `any` check, though.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/main/kotlin/ch/tutteli/atrium/api/fluent/en_GB/charSequenceContainsCheckers.kt
robstoll: thanks

robstoll:
@jGleitz thanks for the feedback

> Looking at the code, I take issue with toHaveNextAndAny and toHaveNextAndAll.

I should have mentioned in the description what I plan to do in a next PR. I realised that for `Expect<Iterable>` notToBeEmptyAnd... is misleading as it could be that Iterable had next but does not have any more because one element was already consumed by a `next()` call. Thus, I will add a notToBeEmptyAndAll/Any/None to collectionExpectations.kt in a next PR -- so far I preferred the shorter version over notToBeEmptyAndAllElements etc. due to the fact that notToBeEmptyAndNoElement would no longer contain `None` (see your comment https://github.com/robstoll/atrium-roadmap/issues/93#issuecomment-737162915).

Yet, seeing that you are confused if we leave out Element, is a bad sign. I'll see if I can come up with a better solution.

jGleitz:
> I think toHaveElementsAndAll would already be better, no?

Yes, I think so.

How will the version of `all` that does not require the collection to have elements be called like?

robstoll:
I suggest `notToHaveElementsOrAll`, `notToHaveElementsOrAny`, `notToHaveElementsOrNone`

PR #906, opened 2021-05-21
Add samples for iterableExpectations of api-fluent

wordhou:
api-fluent #651 

- [X] create IterableExpectationSamples in apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples
  - [X] add one function per function in iterableExpectations.kt (see AnyExpectationSamples)
  - [X] refer to the sample in the KDoc of the corresponding function in iterableExpectations via @sample (check out toBe in anyExpectations.kt to see how you have to refer to the sample, please add it before @since if there is this annotation

### DIFF
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableExpectationSamples.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableExpectationSamples.kt
@@ -18,7 +18,7 @@ class IterableExpectationSamples {
                 .toContain
                 .inAnyOrder // order specifier
                 .atMost(2)
-                .entry { // assertion group about the entries
+                .entry { // assertion group about the entry
                     toBeGreaterThan(2)
                 }
         }
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableExpectationSamples.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableExpectationSamples.kt
@@ -18,9 +18,9 @@ class IterableExpectationSamples {
                 .toContain
                 .inAnyOrder // order specifier
                 .atMost(2)
-                .entries({ // assertion group about the entries
+                .entry { // assertion group about the entries
                     toBeGreaterThan(2)
-                })
+                }
         }
     }
 
@@ -33,15 +33,15 @@ class IterableExpectationSamples {
         fails {
             expect(listOf(1, 8, 5))
                 .notToContain
-                .entries({
+                .entry { // assertion group about the entries
                     toBeGreaterThan(6)
-                })
+                }
         }
     }
 
     @Test
     fun toContainValues() {
-        expect(listOf("A", "B", "C")).toContain("B")
+        expect(listOf("A", "B", "C")).toContain("B", "C")
 
         fails {
             expect(setOf(4, 2, 1)).toContain(3)
@@ -61,7 +61,7 @@ class IterableExpectationSamples {
     @Test
     fun toContainAssertions() {
         expect(listOf(1, 2, 2, 4))
-            .toContain(
+            .toContain( // multiple assertion group entries are evaluated independently
                 { toBeLessThan(2) },
                 { toBeGreaterThan(3) }
             )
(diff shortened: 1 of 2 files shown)


apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableExpectationSamples.kt
```diff
@@ -0,0 +1,241 @@
...
+            .values("A", "B")
+
+        fails {
+            expect(listOf(5, 3, 2, 2, 4))
+                .toContain
+                .inAnyOrder // order specifier
+                .atMost(2)
+                .entries({ // assertion group about the entries
+                    toBeGreaterThan(2)
+                })
```
robstoll, line 23: Fine like this, you could also rewrite it to 
```suggestion
                .entry { // assertion group about the entries
                    toBeGreaterThan(2)
                }
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableExpectationSamples.kt
```diff
@@ -0,0 +1,241 @@
...
+            .toContain(
+                { toBeLessThan(2) },
+                { toBeGreaterThan(3) }
+            )
+
+        fails {
+            expect(listOf(1, 2, 2, 4))
+                .toContain(
+                    { toEqual(3) }, // fails because no element in the list equals 3
+                    { toEqual(2) } // this assertion is not checked
```
robstoll, line 73: We generally try to align text under each other in samples. But that's not the main point, toContain checks all given entries, toEqual(2) holds, that's the reason why it is not reported. Thus change it to
```suggestion
                    { toEqual(3) }, // fails because no element in the list equals 3
                    { toEqual(5) } //  still evaluated and fails as well
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableExpectationSamples.kt
```diff
@@ -0,0 +1,241 @@
...
+        expect(listOf(null)).toContainExactly(null)
+
+        fails {
+            expect(listOf("A", "B")).toContainExactly {
+                toEqual("A")
+            }
+        }
+
+        fails {
+            expect(listOf("A", "B")).toContainExactly(null)
```
robstoll, line 107: I suggest you change this as follows (it is only because of the missing feature of lower bounds in Kotlin that `null` is allowed here):
```suggestion
            expect(listOf("A", null)).toContainExactly(null)
```

robstoll (changes requested):
@wordhou thanks for the samples. A few minor things and we are good to merge :slightly_smiling_face:

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableExpectationSamples.kt
wordhou: Good catch! I was trying to write two failing assertions and missed that.

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableExpectationSamples.kt
```diff
@@ -18,9 +18,9 @@ class IterableExpectationSamples {
                 .toContain
                 .inAnyOrder // order specifier
                 .atMost(2)
-                .entries({ // assertion group about the entries
+                .entry { // assertion group about the entries
```
robstoll, line 21:
```suggestion
                .entry { // assertion group about the entry
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-common/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableExpectationSamples.kt
```diff
@@ -33,15 +33,15 @@ class IterableExpectationSamples {
         fails {
             expect(listOf(1, 8, 5))
                 .notToContain
-                .entries({
+                .entry { // assertion group about the entries
```
robstoll, line 36:
```suggestion
                .entry { // assertion group about the entry
```

robstoll (changes requested):
two subtle things, I'll merge the suggestions in case you have enabled the feature

robstoll:
@wordhou thanks for your first contribution to Atrium :+1: 
Going to review your second in a bit :slightly_smiling_face:

PR #930, opened 2021-06-01
Issue/path expectations sample 919

kacmacuna:
- [x]  add one function per function in pathExpectations.kt and non deprectated function in pathAssertions to PathExpectationSamples (see PathExpectationSamples)

- [x] for functions having two overloads where one expects an assertionCreator in addition -- add the suffix Feature to the function name of the one not expecting assertionCreator (see ListExpectationSamples)

- [x]  refer to the sample in the KDoc of the corresponding function in pathExpectations and pathAssertions via @sample (check out toEqual in anyExpectations.kt to see how you have to refer to the sample)

### DIFF
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
@@ -6,7 +6,6 @@ import ch.tutteli.niok.*
 import java.nio.file.Files
 import java.nio.file.Paths
 import kotlin.test.Test
-import kotlin.test.fail
 
 class PathExpectationSamples {
 
@@ -32,14 +31,14 @@ class PathExpectationSamples {
         val dir = tempDir.newDirectory(TEST_DIR)
         expect(dir).toBeAnEmptyDirectory()
 
-        dir.newFile("test.txt")
+        dir.newFile("$TEST_FILE.txt")
         fails {
             expect(dir).toBeAnEmptyDirectory()
         }
     }
 
     @Test
-    fun toStartWithTestDir() {
+    fun toStartWith() {
         val dir = tempDir.newDirectory(TEST_DIR)
 
         expect(dir.last()).toStartWith(Paths.get(TEST_DIR))
@@ -50,7 +49,7 @@ class PathExpectationSamples {
     }
 
     @Test
-    fun notToStartWithTestDir() {
+    fun notToStartWith() {
         val dir = tempDir.newDirectory(TEST_DIR)
 
         expect(dir.last()).notToStartWith(Paths.get(INVALID_DIR))
@@ -61,7 +60,7 @@ class PathExpectationSamples {
     }
 
     @Test
-    fun toEndWithTestDir() {
+    fun toEndWith() {
         val dir = tempDir.newDirectory(TEST_DIR)
 
         expect(dir).toEndWith(Paths.get(TEST_DIR))
@@ -72,7 +71,7 @@ class PathExpectationSamples {
     }
 
     @Test
-    fun notToEndWithTestDir() {
+    fun notToEndWith() {
         val dir = tempDir.newDirectory(TEST_DIR)
 
         expect(dir).notToEndWith(Paths.get(INVALID_DIR))
@@ -83,7 +82,7 @@ class PathExpectationSamples {
     }
 
     @Test
-    fun toExistTestDir() {
+    fun toExist() {
         val dir = tempDir.newDirectory(TEST_DIR)
 
         expect(dir).toExist()
@@ -94,7 +93,7 @@ class PathExpectationSamples {
     }
 
     @Test
-    fun notToExistTestDir() {
+    fun notToExist() {
         val dir = tempDir.newDirectory(TEST_DIR)
 
         expect(Paths.get(INVALID_DIR)).notToExist()
@@ -184,7 +183,7 @@ class PathExpectationSamples {
     }
 
     @Test
-    fun toHaveDirectoryEntries() {
+    fun toHaveTheDirectoryEntries() {
         val dir = tempDir.newDirectory(TEST_DIR)
         val file = dir.newFile(TEST_FILE)
 
@@ -199,12 +198,14 @@ class PathExpectationSamples {
     }
 
     @Test
-    fun toHaveSameTextualContentAs() {
+    fun toHaveTheSameTextualContentAs() {
+        val writtenTextInFile = "test_test"
+
         val notEmptyFilePath = tempDir.newFile("${TEST_FILE}_1")
-        notEmptyFilePath.toFile().writeText("test_test")
+        notEmptyFilePath.toFile().writeText(writtenTextInFile)
 
         val expectedFilePath = tempDir.newFile("${TEST_FILE}_2")
-        expectedFilePath.toFile().writeText("test_test")
+        expectedFilePath.toFile().writeText(writtenTextInFile)
 
         val emptyFilePath = tempDir.newFile("${TEST_FILE}_3")
 
@@ -234,28 +235,30 @@ class PathExpectationSamples {
 
     @Test
     fun testPathExtensionFeature() {
-        val dir = tempDir.newDirectory(TEST_DIR)
+        val extension = "txt"
+        val dir = tempDir.newFile("$TEST_FILE.$extension")
 
-        expect(dir).extension.notToBeEmpty().toEndWith(TEST_DIR)
+        expect(dir).extension.notToBeEmpty().toEqual(extension)
 
         fails {
-            expect(dir).extension.toBeEmpty().toEndWith(INVALID_DIR)
+            val invalidExtension = "txtt"
+            expect(dir).extension.toEqual(invalidExtension)
         }
     }
 
     @Test
     fun testPathExtension() {
-        val dir = tempDir.newDirectory(TEST_DIR)
+        val extension = "txt"
+        val dir = tempDir.newDirectory("test.$extension")
 
         expect(dir).extension {
             notToBeEmpty()
-            toEndWith(TEST_DIR)
+            toEqual(extension)
         }
 
         fails {
             expect(dir).extension {
-                toBeEmpty()
-                toEndWith(INVALID_DIR)
+                toEqual("txtt")
             }
         }
     }
@@ -319,20 +322,20 @@ class PathExpectationSamples {
 
     @Test
     fun testParentFeature() {
-        val dir = tempDir.newFile("${TEST_DIR}_1")
-        val dir2 = tempDir.newFile("${TEST_DIR}_2")
+        val dir = tempDir.newDirectory("${TEST_DIR}_1")
+        val dir2 = tempDir.newDirectory("${TEST_DIR}_2")
 
         expect(dir).parent.toEqual(dir2.parent)
 
         fails {
-            expect(dir).parent.toEqual(dir2.newFile(TEST_DIR))
+            expect(dir).parent.toEqual(dir2.newDirectory(TEST_DIR))
         }
     }
 
     @Test
     fun testParent() {
-        val dir = tempDir.newFile("${TEST_DIR}_1")
-        val dir2 = tempDir.newFile("${TEST_DIR}_2")
+        val dir = tempDir.newDirectory("${TEST_DIR}_1")
+        val dir2 = tempDir.newDirectory("${TEST_DIR}_2")
 
         expect(dir).parent {
             toEqual(dir2.parent)
@@ -340,7 +343,36 @@ class PathExpectationSamples {
 
         fails {
             expect(dir).parent {
-                toEqual(dir2.newFile(TEST_DIR))
+                toEqual(dir2.newDirectory(TEST_DIR))
+            }
+        }
+    }
+
+    @Test
+    fun testResolveFeature() {
+        val dir = tempDir.newDirectory(TEST_DIR)
+        val fileInDir = dir.newFile("$TEST_FILE.txt")
+
+        expect(dir).resolve("$TEST_FILE.txt").toEqual(fileInDir)
+
+        fails {
+            expect(dir).resolve("$TEST_FILE.ttt").toEqual(fileInDir)
+        }
+    }
+
+    @Test
+    fun testResolve() {
+        val dir = tempDir.newDirectory(TEST_DIR)
+        val fileInDir = dir.newFile("$TEST_FILE.txt")
+
+        expect(dir).resolve("$TEST_FILE.txt") {
+            toEqual(fileInDir)
+            toExist()
+        }
+
+        fails {
+            expect(dir).resolve("$TEST_FILE.ttt") {
+                toEqual(fileInDir)
             }
         }
     }
--- a/apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
+++ b/apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
@@ -1,12 +1,12 @@
 package ch.tutteli.atrium.api.fluent.en_GB.samples
 
-import ch.tutteli.atrium.api.fluent.en_GB.toBeASymbolicLink
-import ch.tutteli.atrium.api.fluent.en_GB.toBeAnEmptyDirectory
+import ch.tutteli.atrium.api.fluent.en_GB.*
 import ch.tutteli.atrium.api.verbs.internal.expect
-import ch.tutteli.niok.newDirectory
-import ch.tutteli.niok.newFile
+import ch.tutteli.niok.*
 import java.nio.file.Files
+import java.nio.file.Paths
 import kotlin.test.Test
+import kotlin.test.fail
 
 class PathExpectationSamples {
 
@@ -29,7 +29,7 @@ class PathExpectationSamples {
 
     @Test
     fun toBeAnEmptyDirectory() {
-        val dir = tempDir.newDirectory("dir")
+        val dir = tempDir.newDirectory(TEST_DIR)
         expect(dir).toBeAnEmptyDirectory()
 
         dir.newFile("test.txt")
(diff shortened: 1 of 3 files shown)


apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -29,12 +28,362 @@ class PathExpectationSamples {
...
+        }
+    }
+
+    private companion object {
+        private const val TEST_DIR = "test_dir"
+        private const val TEST_FILE = "test_file"
+        private const val INVALID_DIR = "invalid_dir"
+
+        private const val RELATIVE_DIR = "relative_dir"
+        private const val ABSOLUTE_DIR = "absolute_dir"
```
robstoll, line 386: this does not look like an absolute dir

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -29,12 +28,362 @@ class PathExpectationSamples {
...
+
+        fails {
+            expect(dir).resolve("$TEST_FILE.ttt") {
+                toEqual(fileInDir)
+            }
+        }
+    }
+
+    private companion object {
+        private const val TEST_DIR = "test_dir"
```
robstoll, line 381: IMO sample files build an exception to the rule of reducing code duplication. Here I would not introduce these constants because the user which looks at the KDOC does not see that they stand for which makes understanding the samples harder. Please inline all those constants

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -29,12 +28,362 @@ class PathExpectationSamples {
...
         fails {
             expect(dir).toBeAnEmptyDirectory()
         }
     }
+
+    @Test
+    fun toStartWith() {
+        val dir = tempDir.newDirectory(TEST_DIR)
+
+        expect(dir.last()).toStartWith(Paths.get(TEST_DIR))
```
robstoll, line 44: This does not show the toStartWith functionality nicely -- in the end your check does the following: `expect(Paths.get("test_dir")).toStartWith(Paths.get("test_dir"))`.
I suggest you rewrite it to something like the following. 
```suggestion
        expect(dir).toStartWith(dir.parent)
```
Same same for notToStartWith/ (not)toEndWith

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -29,12 +28,362 @@ class PathExpectationSamples {
...
+
+    @Test
+    fun toBeAbsolute() {
+        val dir = tempDir.newDirectory(ABSOLUTE_DIR)
+
+        expect(dir).toBeAbsolute()
+
+        fails {
+            expect(Paths.get(RELATIVE_DIR)).toBeAbsolute()
+        }
```
robstoll, line 173: I would not create a directory here as it is not necessary (also ABSOLUTE_DIR is not per se absolute, it's the newDirectory which returns an absolute path and the user might not know this). Better rewrite this to the following which makes it clearer:

```suggestion
        expect(Paths.get("/absolute/path").toBeAbsolute()

        fails {
            expect(Paths.get("relative/path")).toBeAbsolute()
        }
```
same same for toBeRelative

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -29,12 +28,362 @@ class PathExpectationSamples {
...
+            expect(tempDir.newDirectory(ABSOLUTE_DIR)).toBeRelative()
+        }
+    }
+
+    @Test
+    fun toHaveTheDirectoryEntries() {
+        val dir = tempDir.newDirectory(TEST_DIR)
+        val file = dir.newFile(TEST_FILE)
+
+        expect(dir).toHaveTheDirectoryEntries(TEST_FILE)
```
robstoll, line 190: can you please create a second file `test_file2` and add the following line in addition:
```suggestion
        expect(dir).toHaveTheDirectoryEntries(TEST_FILE, "test_file2" /*, ...*/)
```
This way we show the user that he can also pass multiple files

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -29,12 +28,362 @@ class PathExpectationSamples {
...
+        }
+
+    }
+
+    @Test
+    fun toHaveTheSameTextualContentAs() {
+        val writtenTextInFile = "test_test"
+
+        val notEmptyFilePath = tempDir.newFile("${TEST_FILE}_1")
+        notEmptyFilePath.toFile().writeText(writtenTextInFile)
```
robstoll, line 205: fine like this but you could also rewrite it to the following (using niok's writeText)
```suggestion
        notEmptyFilePath.writeText(writtenTextInFile)
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -29,12 +28,362 @@ class PathExpectationSamples {
...
+
+        val expectedFilePath = tempDir.newFile("${TEST_FILE}_2")
+        expectedFilePath.toFile().writeText(writtenTextInFile)
```
robstoll, line 208:
```suggestion
        expectedFilePath.writeText(writtenTextInFile)
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -29,12 +28,362 @@ class PathExpectationSamples {
...
+
+        fails {
+            expect(emptyFilePath).toHaveTheSameTextualContentAs(expectedFilePath)
+        }
+    }
+
+    @Test
+    fun toHaveTheSameBinaryContentAs() {
+        val notEmptyFilePath = tempDir.newFile("${TEST_FILE}_1")
+        notEmptyFilePath.toFile().writeBytes(byteArrayOf(1, 2, 3))
```
robstoll, line 222:
```suggestion
        notEmptyFilePath.writeBytes(byteArrayOf(1, 2, 3))
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -29,12 +28,362 @@ class PathExpectationSamples {
...
+
+        val expectedFilePath = tempDir.newFile("${TEST_FILE}_2")
+        expectedFilePath.toFile().writeBytes(byteArrayOf(1, 2, 3))
```
robstoll, line 225:
```suggestion
        expectedFilePath.writeBytes(byteArrayOf(1, 2, 3))
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -29,12 +28,362 @@ class PathExpectationSamples {
...
+
+        expect(notEmptyFilePath).toHaveTheSameBinaryContentAs(expectedFilePath)
+
+        fails {
+            expect(emptyFilePath).toHaveTheSameBinaryContentAs(expectedFilePath)
+        }
+    }
+
+    @Test
+    fun testPathExtensionFeature() {
```
robstoll, line 237: we don't use a `testPath` prefix in the method names, please change to:
```suggestion
    fun extensionFeature() {
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -29,12 +28,362 @@ class PathExpectationSamples {
...
+        expect(dir).extension.notToBeEmpty().toEqual(extension)
+
+        fails {
+            val invalidExtension = "txtt"
+            expect(dir).extension.toEqual(invalidExtension)
+        }
+    }
+
+    @Test
+    fun testPathExtension() {
```
robstoll, line 250:
```suggestion
    fun extension() {
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -29,12 +28,362 @@ class PathExpectationSamples {
...
+
+        fails {
+            expect(dir).extension {
+                toEqual("txtt")
+            }
+        }
+    }
+
+    @Test
+    fun testFileNameFeature() {
```
robstoll, line 267:
```suggestion
    fun fileNameFeature() {
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -29,12 +28,362 @@ class PathExpectationSamples {
...
+
+        expect(dir).fileName.toEndWith(TEST_DIR).notToStartWith(INVALID_DIR).notToBeBlank()
+
+        fails {
+            expect(dir).fileName.toEndWith(INVALID_DIR).toStartWith(TEST_DIR)
+        }
+    }
+
+    @Test
+    fun testFileName() {
```
robstoll, line 278:
```suggestion
    fun fileName() {
```

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -29,12 +28,362 @@ class PathExpectationSamples {
...
+            expect(emptyFilePath).toHaveTheSameBinaryContentAs(expectedFilePath)
+        }
+    }
+
+    @Test
+    fun testPathExtensionFeature() {
+        val extension = "txt"
+        val dir = tempDir.newFile("$TEST_FILE.$extension")
+
+        expect(dir).extension.notToBeEmpty().toEqual(extension)
```
robstoll, line 241: please add the comment about the subject change (see e.g. ListExpectationSamples -> getFeature)
```suggestion
        expect(dir)
        .extension // subject is now of type String (actually "txt")
        .toEqual(extension)
```
Same same for other functions further below

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -29,12 +28,362 @@ class PathExpectationSamples {
...
+
+        fails {
+            val invalidExtension = "txtt"
+            expect(dir).extension.toEqual(invalidExtension)
```
robstoll, line 245: Please have a look at e.g. ListExpectationSamples -> getFeature. It would be nice if you could include the comment about the fail fast behaviour and about the subject change, something like
```suggestion
            expect(dir)
              .extension         // subject is now of type String (actually "txt")
              .toStartWith("jp") // fails because extension starts with `tx`
              .toEndWith("eg")   // not reported
              //                    use `extension { ... }` if you want that all expectations are evaluated
```

robstoll (changes requested):

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
kacmacuna: `Paths.get("/absolute/path").isAbsolute` returns false.
We could use `FileSystemView.getFileSystemView().homeDirectory.toPath().resolve("foo")` instead and maybe even refactor it as method inside of test class or somewhere else.
`private fun getAbsolutePath(name: String) = FileSystemView.getFileSystemView().homeDirectory.toPath().resolve(name)`

robstoll: I guess you are using Windows. On a unix system it returns true. I figured this was abstracted away by Java in nio but apparently not (learned something new :slightly_smiling_face:). In this case you need to write it as follows:
```kotlin
    // on unix like systems / and \ on Windows
    val s = FileSystems.getDefault().getSeparator()
    expect(Paths.get("${s}absolute${s}path").toBeAbsolute()

    fails {
        expect(Paths.get("relative${s}path")).toBeAbsolute()
    }
```

kacmacuna: on windows it still returns false, but "\\absolute\path" works fine.

kacmacuna: I investigated further and utility class WindowsPathParser decides in this case that "\absolute\path" is relative.

robstoll: That's a pity, so probably we ned to prefix the path with `C:` in case of Windows. Can you quickly check if this would work?

kacmacuna: C:\absolute\path returns true

kacmacuna: Will this work?
Screenshot_43

robstoll: I guess it will but it is somehow also a bit too much for a simple example (which should also be suitable for beginners).
I therefore suggest we change it to the following which takes up less space so the most relative thing:
```
val s = FileSystems.getDefault().separator
val prefix = if (s == "\\") "C:" else ""
expect(Paths.get("$prefix${s}absolute${s}path")).toBeAbsolute()
```
What do you think?

kacmacuna: yup I understand your point, I will go with your solution

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -244,11 +246,16 @@ class PathExpectationSamples {
...
+            .toEqual(extension)
+            .notToEndWith("jpg")
 
         fails {
-            val invalidExtension = "txtt"
-            expect(dir).extension.toEqual(invalidExtension)
+            expect(dir).extension // subject is now of type String (actually "txt")
+                .toEqual("txtt") // fails because it doesn't equal to "txtt"
+                .toEndWith("jpg") // not reported
+            //                              use `.extension` if you want that all expectations are evaluated
```
robstoll, line 258: Can you please align the comments, makes it a bit more pleasant to read (also further below)
```suggestion
            expect(dir).extension  // subject is now of type String (actually "txt")
                .toEqual("txtt")   // fails because it doesn't equal to "txtt"
                .toEndWith("jpg")  // not reported
            //                        use `.extension` if you want that all expectations are evaluated
```

robstoll (changes requested):

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -252,10 +252,10 @@ class PathExpectationSamples {
...
 
         fails {
-            expect(dir).extension // subject is now of type String (actually "txt")
-                .toEqual("txtt") // fails because it doesn't equal to "txtt"
-                .toEndWith("jpg") // not reported
-            //                              use `.extension` if you want that all expectations are evaluated
+            expect(dir).extension //            subject is now of type String (actually "txt")
+                .toEqual("txtt") //     fails because it doesn't equal to "txtt"
+                .toEndWith("jpg") //    not reported
+            //                                  use `.extension` if you want that all expectations are evaluated
```
robstoll, line 258: That does not look well aligned. also, please align not only the text but also the `//` the only exception to that rule is if `//` starts on a new line (without prefixed code). The reason for this is that Intellij would move the // to the beginning of the indentation when formatting.
So please use the following approach:
```suggestion
            expect(dir).extension  // subject is now of type String (actually "txt")
                .toEqual("txtt")   // fails because it doesn't equal to "txtt"
                .toEndWith("jpg")  // not reported
            //                        use `.extension` if you want that all expectations are evaluated
```

robstoll (changes requested):

kacmacuna:
everything is correctly formatted on local AS but github messes it up for some reason

apis/fluent-en_GB/atrium-api-fluent-en_GB-jvm/src/test/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/PathExpectationSamples.kt
```diff
@@ -252,10 +252,10 @@ class PathExpectationSamples {
             .notToEndWith("jpg")
 
         fails {
-            expect(dir).extension //            subject is now of type String (actually "txt")
-                .toEqual("txtt") //     fails because it doesn't equal to "txtt"
-                .toEndWith("jpg") //    not reported
-            //                                  use `.extension` if you want that all expectations are evaluated
+            expect(dir).extension         // subject is now of type String (actually "txt")
```
robstoll: Could it be, that you have enabled parameter name hints in Intellij? The alignment is still not right (as you can see here)

kacmacuna: Think that is outdated version, this is new one
Screenshot_45

robstoll: hmm... starts to be bothersome really. Looks all good if I go to https://github.com/robstoll/atrium/pull/930/files
Doesn't look good if I come from the github notification where one should see only the new differences:
https://github.com/robstoll/atrium/pull/930/files/5dbe7c8c8db13091cbb5a22f6b669eff86f2a893..eb3113721fb5c3ba2eae5834bb73e7be378f57f1#diff-bec3c79f9b9bc972773563affa67021ede5661d57d9d9968bb81f8cc95af3780

robstoll:
@kacmacuna thanks for your third contribution to Atrium :tada:

PR #931, opened 2021-06-03
Make notToContain less verbose and show mismatched indices

### LINKED ISSUE #722: less verbose containsNot reporting and more hints

## Code related feature
Instead of 
```
expected that subject: [1, 2, 3, 4]        (java.util.Arrays.ArrayList <1234789>)
◆ does not contain: 
  ⚬ an element which: 
      » is greater than: 2        (kotlin.Int <1234789>)
    ✘ ▶ number of such entries: 2
        ◾ is: 0        (kotlin.Int <1234789>)
    ✔ ▶ has at least one element: true
        ◾ is: true
```
We would like to see (similar to `all`)
```
expected that subject: [1, 2, 3, 4]        (java.util.Arrays.ArrayList <1234789>)
◆ does not contain: 
  ⚬ an element which: 
      » is greater than: 2        (kotlin.Int <1234789>)
     ❗❗ following entries were mismatched: 
        ⚬ index 3: 3        (kotlin.Int <1234789>)
        ⚬ index 4: 4        (kotlin.Int <1234789>)
```
Also the 
```
▶ has at least one element: true
        ◾ is: true
```
Should be moved out of `does not contain` (and hence only be shown if failing) in the spirit of #305
Also, we should not re-invent the wheel and use the existing `hasNext` implementation instead of an own `has at least one element`

In case of an empty Iterable, we should still show what the user intended to search for IMO. Something like:
```
expected that subject: []        (kotlin.collections.EmptySet <1837188010>)
◆ has: a next element
  » does not contain:
    ⚬ an element which:
       » is greater than: 2        (kotlin.Int <1234789>)
```
IMO also fine if we show `✘ ▶ number of such entries: ...` in addition if this is simpler (which is the case most likely)

Without this extra information it could be confusing for a user

wordhou:
Work in progress pull request. Resolves #722.

- [X] Modify `ContainsAssertionCreator` so that subclasses can decorate the assertion
- [X] Modify `InAnyOrderEntriesAssertionCreator` in the `NotSearchBehavior` case:
  - [X] Remove feature assertion about the number of occurrences
  - [X] Add explanatory group with the mismatches
  - [X] Use `_logic.hasNext` to decorate the parent assertion
  - [X] Modify specs
- [x] Refactor `expect(IterableLike).all` to use new methods in `containsHelpers`.
- [x] Modify `InAnyOrderValuesAssertionCreator` and/or `ContainsObjectsAssertionCreator` in the `NotSearchBehavior` case:
  - [x] Decorate parent assertion with `_logic.hasNext`
  - [x] Remove feature assertion about the number of occurrences
  - [x] Add explanatory group with mismatches
  - [x] Modify specs
- [x] Forgive backward compatability for changing specs

### DIFF
--- a/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/basic/contains/creators/impl/ContainsObjectsAssertionCreator.kt
+++ b/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/basic/contains/creators/impl/ContainsObjectsAssertionCreator.kt
@@ -5,7 +5,6 @@ import ch.tutteli.atrium.assertions.AssertionGroup
 import ch.tutteli.atrium.assertions.builders.assertionBuilder
 import ch.tutteli.atrium.creating.AssertionContainer
 import ch.tutteli.atrium.logic.creating.basic.contains.Contains
-import ch.tutteli.atrium.logic.creating.iterable.contains.searchbehaviours.NotSearchBehaviour
 import ch.tutteli.atrium.reporting.translating.Translatable
 
 /**
--- a/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/basic/contains/creators/impl/ContainsObjectsAssertionCreator.kt
+++ b/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/basic/contains/creators/impl/ContainsObjectsAssertionCreator.kt
@@ -2,10 +2,10 @@ package ch.tutteli.atrium.logic.creating.basic.contains.creators.impl
 
 import ch.tutteli.atrium.assertions.Assertion
 import ch.tutteli.atrium.assertions.AssertionGroup
-import ch.tutteli.atrium.assertions.AssertionGroupType
 import ch.tutteli.atrium.assertions.builders.assertionBuilder
 import ch.tutteli.atrium.creating.AssertionContainer
 import ch.tutteli.atrium.logic.creating.basic.contains.Contains
+import ch.tutteli.atrium.logic.creating.iterable.contains.searchbehaviours.NotSearchBehaviour
 import ch.tutteli.atrium.reporting.translating.Translatable
 
 /**
@@ -32,7 +32,7 @@ abstract class ContainsObjectsAssertionCreator<T : Any, TT : Any, in SC, S : Con
     checkers: List<C>
 ) : ContainsAssertionCreator<T, TT, SC, C>(searchBehaviour, checkers) {
 
-    final override fun searchAndCreateAssertion(
+    override fun searchAndCreateAssertion(
         multiConsumableContainer: AssertionContainer<TT>,
         searchCriterion: SC,
         featureFactory: (Int, Translatable) -> AssertionGroup
@@ -40,11 +40,10 @@ abstract class ContainsObjectsAssertionCreator<T : Any, TT : Any, in SC, S : Con
         val count = search(multiConsumableContainer, searchCriterion)
         val featureAssertion = featureFactory(count, descriptionNumberOfOccurrences)
 
-        val inAnyOrderAssertion =  assertionBuilder.list
+        return assertionBuilder.list
             .withDescriptionAndRepresentation(groupDescription, searchCriterion)
             .withAssertion(featureAssertion)
             .build()
-        return decorateInAnyOrderAssertion(inAnyOrderAssertion, multiConsumableContainer)
     }
 
     /**
--- a/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/basic/contains/creators/impl/ContainsObjectsAssertionCreator.kt
+++ b/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/basic/contains/creators/impl/ContainsObjectsAssertionCreator.kt
@@ -40,10 +40,11 @@ abstract class ContainsObjectsAssertionCreator<T : Any, TT : Any, in SC, S : Con
         val count = search(multiConsumableContainer, searchCriterion)
         val featureAssertion = featureFactory(count, descriptionNumberOfOccurrences)
 
-        return assertionBuilder.customType(getAssertionGroupType())
+        val inAnyOrderAssertion =  assertionBuilder.list
             .withDescriptionAndRepresentation(groupDescription, searchCriterion)
-            .withAssertions(decorateAssertion(multiConsumableContainer, featureAssertion))
+            .withAssertion(featureAssertion)
             .build()
+        return decorateInAnyOrderAssertion(inAnyOrderAssertion, multiConsumableContainer)
     }
 
     /**
@@ -55,11 +56,6 @@ abstract class ContainsObjectsAssertionCreator<T : Any, TT : Any, in SC, S : Con
      * Provides the translation for [AssertionGroup.description]
      */
     protected abstract val groupDescription: Translatable
-
-    /**
-     * Provides the [AssertionGroupType] for the resulting [AssertionGroup].
-     */
-    protected abstract fun getAssertionGroupType(): AssertionGroupType
 
 
     /**
@@ -73,9 +69,4 @@ abstract class ContainsObjectsAssertionCreator<T : Any, TT : Any, in SC, S : Con
      * @return The number of times the [searchCriterion] matched in the subject of this expectation.
      */
     protected abstract fun search(multiConsumableContainer: AssertionContainer<TT>, searchCriterion: SC): Int
-
-    /**
-     * Either return the given [featureAssertion] as [List] or add further assertions.
-     */
-    abstract fun decorateAssertion(container: AssertionContainer<TT>, featureAssertion: Assertion): List<Assertion>
 }
--- a/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/iterable/contains/creators/impl/InAnyOrderEntriesAssertionCreator.kt
+++ b/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/iterable/contains/creators/impl/InAnyOrderEntriesAssertionCreator.kt
@@ -3,6 +3,7 @@ package ch.tutteli.atrium.logic.creating.iterable.contains.creators.impl
 import ch.tutteli.atrium.assertions.Assertion
 import ch.tutteli.atrium.assertions.AssertionGroup
 import ch.tutteli.atrium.assertions.builders.assertionBuilder
+import ch.tutteli.atrium.assertions.builders.fixedClaimGroup
 import ch.tutteli.atrium.assertions.builders.invisibleGroup
 import ch.tutteli.atrium.core.None
 import ch.tutteli.atrium.core.getOrElse
@@ -16,12 +17,15 @@ import ch.tutteli.atrium.logic.creating.iterable.contains.IterableLikeContains
 import ch.tutteli.atrium.logic.creating.iterable.contains.searchbehaviours.InAnyOrderSearchBehaviour
 import ch.tutteli.atrium.logic.creating.iterable.contains.searchbehaviours.NotSearchBehaviour
 import ch.tutteli.atrium.logic.creating.typeutils.IterableLike
+import ch.tutteli.atrium.logic.hasNext
 import ch.tutteli.atrium.logic.impl.allCreatedAssertionsHold
 import ch.tutteli.atrium.logic.impl.createExplanatoryAssertionGroup
-import ch.tutteli.atrium.logic.impl.createHasElementAssertion
+import ch.tutteli.atrium.logic.impl.createExplanatoryGroupForMismatches
+import ch.tutteli.atrium.logic.impl.createIndexAssertions
 import ch.tutteli.atrium.reporting.translating.Translatable
 import ch.tutteli.atrium.translations.DescriptionIterableAssertion
 import ch.tutteli.atrium.translations.DescriptionIterableAssertion.AN_ELEMENT_WHICH
+import ch.tutteli.kbox.identity
 
 /**
  * Represents a creator of a sophisticated `contains` assertions for [Iterable] where an expected entry can appear
@@ -62,7 +66,7 @@ class InAnyOrderEntriesAssertionCreator<E : Any, T : IterableLike>(
         return if (searchBehaviour is NotSearchBehaviour) {
             assertionBuilder.invisibleGroup
                 .withAssertions(
-                    createHasElementAssertion(list),
+                    multiConsumableContainer.hasNext(::identity),
                     inAnyOrderAssertion
                 )
                 .build()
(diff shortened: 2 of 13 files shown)


logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/basic/contains/creators/impl/ContainsObjectsAssertionCreator.kt
```diff
@@ -32,17 +32,17 @@ abstract class ContainsObjectsAssertionCreator<T : Any, TT : Any, in SC, S : Con
     checkers: List<C>
 ) : ContainsAssertionCreator<T, TT, SC, C>(searchBehaviour, checkers) {
 
-    final override fun searchAndCreateAssertion(
+    override fun searchAndCreateAssertion(
```
wordhou, line 35: I don't know if this is ideal, but I overrode `searchAndCreateAssertion` in the `InAnyOrderValuesAssertionCreator` to handle the case we're changing.

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
```diff
@@ -134,22 +122,6 @@ class DefaultIterableLikeAssertions : IterableLikeAssertions {
...
-        .mapWithIndex()
-        .filter { predicate(it) }
-        .map { (index, element) ->
-            assertionBuilder.createDescriptive(
-                TranslatableWithArgs(DescriptionIterableAssertion.INDEX, index),
-                element,
-                falseProvider
-            )
-        }
-        .toList()
```
wordhou, line 151: I saw no reason to rewrite this logic so I moved this method into `containsHelpers.kt`

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/basic/contains/creators/impl/ContainsObjectsAssertionCreator.kt
robstoll: Maybe not but I wonder, what was the reason that you did it? was decorateInAnyOrderAssertion not good enough?

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/iterable/contains/creators/impl/InAnyOrderEntriesAssertionCreator.kt
```diff
@@ -92,26 +123,23 @@ class InAnyOrderEntriesAssertionCreator<E : Any, T : IterableLike>(
         return group to count
     }
 
-    //TODO 0.18.0 check if this is still state of the art to add a hint that no assertion was created
```
robstoll, line 95: please re-add the TODO, it's something I still need to check in the next version

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/iterable/contains/creators/impl/InAnyOrderEntriesAssertionCreator.kt
```diff
@@ -92,26 +123,23 @@ class InAnyOrderEntriesAssertionCreator<E : Any, T : IterableLike>(
...
     @Suppress("DEPRECATION" /* OptIn is only available since 1.3.70 which we cannot use if we want to support 1.2 */)
     @UseExperimental(ExperimentalComponentFactoryContainer::class)
-    private fun addEmptyAssertionCreatorLambdaIfNecessary(
+    private fun createEmptyAssertionHintIfNecessary(
         container: AssertionContainer<*>,
-        assertions: MutableList<Assertion>,
         searchCriterion: (Expect<E>.() -> Unit)?,
         count: Int
-    ) {
+    ): Assertion? {
```
robstoll, line 132: better to have a side-effect free function :+1:

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/basic/contains/creators/impl/ContainsObjectsAssertionCreator.kt
```diff
@@ -2,10 +2,10 @@ package ch.tutteli.atrium.logic.creating.basic.contains.creators.impl
 
 import ch.tutteli.atrium.assertions.Assertion
 import ch.tutteli.atrium.assertions.AssertionGroup
-import ch.tutteli.atrium.assertions.AssertionGroupType
 import ch.tutteli.atrium.assertions.builders.assertionBuilder
 import ch.tutteli.atrium.creating.AssertionContainer
 import ch.tutteli.atrium.logic.creating.basic.contains.Contains
+import ch.tutteli.atrium.logic.creating.iterable.contains.searchbehaviours.NotSearchBehaviour
```
robstoll, line 8: remove again (not used)
```suggestion
```

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/iterable/contains/creators/impl/InAnyOrderValuesAssertionCreator.kt
```diff
@@ -1,16 +1,24 @@
...
 import ch.tutteli.atrium.core.getOrElse
 import ch.tutteli.atrium.creating.AssertionContainer
 import ch.tutteli.atrium.logic.creating.basic.contains.creators.impl.ContainsObjectsAssertionCreator
 import ch.tutteli.atrium.logic.creating.iterable.contains.IterableLikeContains
 import ch.tutteli.atrium.logic.creating.iterable.contains.searchbehaviours.InAnyOrderSearchBehaviour
 import ch.tutteli.atrium.logic.creating.iterable.contains.searchbehaviours.NotSearchBehaviour
 import ch.tutteli.atrium.logic.creating.typeutils.IterableLike
+import ch.tutteli.atrium.logic.hasNext
+import ch.tutteli.atrium.logic.impl.createExplanatoryGroupForMismatches
 import ch.tutteli.atrium.logic.impl.createHasElementAssertion
```
robstoll, line 17: could be removed
```suggestion
```

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/iterable/contains/creators/impl/InAnyOrderValuesAssertionCreator.kt
```diff
@@ -38,31 +46,52 @@ class InAnyOrderValuesAssertionCreator<SC, T : IterableLike>(
...
+                    .withListType
+                    .withClaim(mismatches.isEmpty())
+                    .withDescriptionAndRepresentation(groupDescription, searchCriterion)
+                    .withAssertion(createExplanatoryGroupForMismatches(mismatches))
+                    .build()
+            } else {
+                assertionBuilder.invisibleGroup
+                    .withAssertion(
+                        assertionBuilder.createDescriptive(groupDescription, searchCriterion, falseProvider)
+                    ).build()
```
robstoll, line 74: I may be mistaken but I think we this way we don't show to the user the right reason why the expectation failed. For instance, if one writes `expect(setOf<Int>()).notToContain(2)`. Then he would see roughly the following:
```
expected that subject: []        (kotlin.collections.EmptySet <1837188010>)
◆ does not contain:
  ⚬ an element which equals: 2
```
Which is actually a valid statement and should not fail as such (2 is not in the empty list). Thus we should show the following:
```
expected that subject: []        (kotlin.collections.EmptySet <1837188010>)
◆ has next: true
  » does not contain:
    ⚬ an element which equals: 2         (kotlin.Int <1234789>)
```
This way it is clear that `notToContain` is actually `toHaveElementsAndNone`


*Edit*:  We should really refactor this class, the decorating approach isn't any better than the overriding approach laughing:. I see now that we add the hasNext assertion in decorateInAnyOrderAssertion. It's kind of a hack that you use a failing descriptive assertion here (because this part holds) and I therefore suggest that you write a comment here so that the next person sees it immediately. 
```suggestion
                assertionBuilder.invisibleGroup
                    .withAssertion(
                        // TODO 0.18.0 this is a hack so that we see the searchCriterion in the reporting. Refactor so that it is an explanatoryAssertion instead
                        assertionBuilder.createDescriptive(groupDescription, searchCriterion, falseProvider)
                    ).build()
```
Or even better, remove the `if` here and place it in decorateInAnyOrderAssertion where you can turn the `inAnyOrderAssertion` into an explanatory assertion group

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/iterable/contains/creators/impl/InAnyOrderEntriesAssertionCreator.kt
```diff
@@ -68,18 +87,30 @@ class InAnyOrderEntriesAssertionCreator<E : Any, T : IterableLike>(
...
+            }
+            assertions.add(createExplanatoryGroupForMismatches(mismatches))
+            val hasNext = multiConsumableContainer.hasNext(::identity).holds()
+            return assertionBuilder.fixedClaimGroup
+                .withListType
+                .withClaim(mismatches.isEmpty() && hasNext && emptyAssertionHint == null)
+                .withDescriptionAndEmptyRepresentation(AN_ELEMENT_WHICH)
+                .let {
+                    if (hasNext) it.withAssertions(assertions)
+                    else it.withAssertion(explanatoryGroup)
```
robstoll, line 105: I might be wrong but I think this `if` shouldn't be necessary or rather is wrong. I guess you don't show that the assertionCreator-Lambda was empty in case the iterable does not hasNext. Please check, what happens if one writes (please add a test-case):
```
expect(listOf<int>()).notToContain { }
```

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/iterable/contains/creators/impl/InAnyOrderValuesAssertionCreator.kt
```diff
@@ -38,31 +46,52 @@ class InAnyOrderValuesAssertionCreator<SC, T : IterableLike>(
...
-            )
+            assertionBuilder.invisibleGroup.withAssertions(
+                multiConsumableContainer.hasNext(::identity),
+                inAnyOrderAssertion
+            ).build()
         } else {
-            listOf(featureAssertion)
+            inAnyOrderAssertion
         }
     }
```
robstoll, line 96: Follow-up to the above comment. Something like the following here:
```kotlin
    return if (searchBehaviour is NotSearchBehaviour) {
            val hasNext = multiConsumableContainer.hasNext(::identity)
            assertionBuilder.invisibleGroup.withAssertions(
                hasNext,
                if (hasNext.holds()) {
                    inAnyOrderAssertion
                } else {
                    assertionBuilder.explanatoryGroup.withDefaultType.withAssertion(inAnyOrderAssertion).build()
                }
            ).build()

        } else {
            inAnyOrderAssertion
        }
```

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/IterableToContainEntriesSpecBase.kt
```diff
@@ -23,14 +24,25 @@ abstract class IterableToContainEntriesSpecBase(
         val anElementWhich = DescriptionIterableAssertion.AN_ELEMENT_WHICH.getDefault()
         val toBeLessThanDescr = DescriptionComparableAssertion.IS_LESS_THAN.getDefault()
         val toBeGreaterThanDescr = DescriptionComparableAssertion.IS_GREATER_THAN.getDefault()
+        val hasDescriptionBasic = DescriptionBasic.HAS.getDefault()
+        val nextElement = DescriptionIterableAssertion.NEXT_ELEMENT.getDefault()
+        val mismatches = DescriptionIterableAssertion.WARNING_MISMATCHES.getDefault()
```
robstoll, line 29: shouldn't be necessary as it is already defined in the super-class companion IterableToContainSpecBase

robstoll (changes requested):

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/iterable/contains/creators/impl/InAnyOrderEntriesAssertionCreator.kt
wordhou: I agree, that `if` is not necessary here.

I took a look at what happens with that test case and I think we've found a bug. In that test case, the hint for the empty `assertionCreator` is actually getting added twice, once from `createExplanatoryAssertionGroup` in `containsHelpers`, and once more from `createEmptyAssertionHintIfNecessary`. There's some other strange things happening with the empty assertion hints but I checked and these same things are happening on master, so I'll create another branch to investigate these issues.

misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/IterableToContainEntriesSpecBase.kt
wordhou: I moved `hasDescriptionBasic` and `nextElement` up to the `IterableToContainSpecBase` companion.

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/iterable/contains/creators/impl/InAnyOrderValuesAssertionCreator.kt
wordhou: I took another stab at this, took out the hacks, and simplified the logic. I think earlier I wasn't understanding how to use the explanatoryGroup.

I agree that a refactor may be the best thing for `ContainsObjectsAssertionCreator` and its children. For now it seems like we need both the `decorateInAnyOrderAssertion` as well as the override for `searchAndCreateAssertion` here.

settings.gradle.kts
```diff
@@ -153,7 +159,10 @@ buildscript {
                 "CharSequenceContains.*Spec.*points to containsNot",
                 "IterableContains.*Spec.*points to containsNot",
                 // we improved reporting for containsNoDuplicates
-                "IterableExpectationsSpec.*`(containsNoDuplicates|contains noDuplicates)`"
+                "IterableExpectationsSpec.*`(containsNoDuplicates|contains noDuplicates)`",
+                // we improved reporting for notToContain
```
robstoll, line 163:
```suggestion
                // we improved reporting for notToContain with 0.17.0
```

robstoll (approved):

robstoll:
@wordhou great work, thank you very much for the PR :+1:

PR #938, opened 2021-06-11
Fail explanatory assertion group on empty assertionCreator

### LINKED ISSUE #933: Empty assertionCreator lambda warning is added twice
*Affected Version*: 1.7
*API* (fluent-en_GB, infix-en_GB): both
*Platform* (jvm, js): logic

```kotlin
expect(oneToSeven()).notToContainFun({})
```
produces
```
"expected that subject: kotlin.sequences.SequencesKt___SequencesKt$asIterable$$inlined$Iterable$1@5453dabb        (kotlin.sequences.SequencesKt___SequencesKt$asIterable$$inlined$Iterable$1 <1414781627>)
    ◆ does not contain: 
      ⚬ an element which: 
          » at least one assertion defined: false
              » You forgot to define assertions in the assertionCreator-lambda
              » Sometimes you can use an alternative to `{ }` For instance, instead of `toThrow<..> { }` you should use `toThrow<..>()`
        ✔ ▶ number of such entries: 0
            ◾ is: 0        (kotlin.Int <449752083>)
        ✔ ▶ has at least one element: true
            ◾ is: true
        ✘ at least one assertion defined: false
            » You forgot to define assertions in the assertionCreator-lambda
            » Sometimes you can use an alternative to `{ }` For instance, instead of `toThrow<..> { }` you should use `toThrow<..>()`"
```
and
```kotlin
expect(oneToSeven()).notToContainFun({}, {toBeLessThan(3.0)})
```
produces
```
expected that subject: kotlin.sequences.SequencesKt___SequencesKt$asIterable$$inlined$Iterable$1@1b2be2e2        (kotlin.sequences.SequencesKt___SequencesKt$asIterable$$inlined$Iterable$1 <455860962>)
    ◆ does not contain: 
      ⚬ an element which: 
          » at least one assertion defined: false
              » You forgot to define assertions in the assertionCreator-lambda
              » Sometimes you can use an alternative to `{ }` For instance, instead of `toThrow<..> { }` you should use `toThrow<..>()`
        ✔ ▶ number of such entries: 0
            ◾ is: 0        (kotlin.Int <449752083>)
        ✔ ▶ has at least one element: true
            ◾ is: true
        ✘ at least one assertion defined: false
            » You forgot to define assertions in the assertionCreator-lambda
            » Sometimes you can use an alternative to `{ }` For instance, instead of `toThrow<..> { }` you should use `toThrow<..>()`
      ⚬ an element which: 
          » is less than: 3.0        (kotlin.Double <888510275>)
        ✘ ▶ number of such entries: 2
            ◾ is: 0        (kotlin.Int <449752083>)
        ✔ ▶ has at least one element: true
            ◾ is: true
```

## Describe the bug
On IterableLike `notToContain.entry/entries` the empty assertionCreator lambda is getting added twice. In `InAnyOrderEntriesAssertionCreator` it appears to be getting added once in `addEmptyAssertionCreatorLambdaIfNeccessary`and once in `createExplanatoryAssertionGroup`.

I added a test in this branch to take a look at this issue:
https://github.com/wordhou/atrium/blob/empty-assertion-creator-issues/misc/specs/atrium-specs-common/src/main/kotlin/ch/tutteli/atrium/specs/integration/IterableNotToContainEntriesExpectationsSpec.kt#L106

wordhou:
Resolves #933. Whenever an assertionCreator is taken as an argument of an expectation, an explanatory group is created that describes the assertions created. If no assertions are created, then an empty lambda failure hint is given. This PR makes the explanatory group fail when the empty lambda failure hint is given, and then removes the `createEmptyAssertionHintIfNecessary` from the assertion creators.

### DIFF
--- a/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/collectors/AssertionsOptionExplantoryExtensions.kt
+++ b/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/collectors/AssertionsOptionExplantoryExtensions.kt
@@ -16,6 +16,8 @@ import ch.tutteli.atrium.logic.collectForCompositionBasedOnSubject
  * Collects the assertions [assertionCreator] creates and uses them as [AssertionGroup.assertions].
  *
  * //TODO 0.18.0 in case we somehow incorporate the current container in AssertionsOptions, then remove container as parameter
+ *
+ * TODO 1.0.0 at the latest: use type ExplanatoryGroup.FinalStep when ExplanatoryAssertionGroupFinalStep is removed
  */
 @Suppress("DEPRECATION" /* OptIn is only available since 1.3.70 which we cannot use if we want to support 1.2 */)
 @UseExperimental(ExperimentalComponentFactoryContainer::class)
--- a/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/collectors/AssertionsOptionExplantoryExtensions.kt
+++ b/logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/collectors/AssertionsOptionExplantoryExtensions.kt
@@ -3,9 +3,13 @@ package ch.tutteli.atrium.logic.creating.collectors
 import ch.tutteli.atrium.assertions.AssertionGroup
 import ch.tutteli.atrium.assertions.ExplanatoryAssertionGroupType
 import ch.tutteli.atrium.assertions.builders.AssertionsOption
+import ch.tutteli.atrium.assertions.builders.ExplanatoryAssertionGroupFinalStep
+import ch.tutteli.atrium.core.None
 import ch.tutteli.atrium.core.Option
 import ch.tutteli.atrium.creating.AssertionContainer
+import ch.tutteli.atrium.creating.CollectingExpect
 import ch.tutteli.atrium.creating.Expect
+import ch.tutteli.atrium.creating.ExperimentalComponentFactoryContainer
 import ch.tutteli.atrium.logic.collectForCompositionBasedOnSubject
 
 /**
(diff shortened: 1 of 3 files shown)


logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/impl/containsHelpers.kt
```diff
@@ -46,7 +50,14 @@ internal fun <E : Any> createExplanatoryAssertionGroup(
             //TODO 0.18.0 looks a lot like toBeNullIfNullGiven
             if (assertionCreatorOrNull != null) {
                 // we don't use a subject, we will not show it anyway
-                it.collectAssertions(container, None, assertionCreatorOrNull)
+                val collectingExpect = CollectingExpect<E>(None, container.components)
+                // not using addAssertionsCreatedBy on purpose so that we don't append a failing assertion
+                collectingExpect.assertionCreatorOrNull()
+                if(collectingExpect.getAssertions().isEmpty()) {
+                    it.collectAssertions(container, None, assertionCreatorOrNull).failing
```
wordhou, line 57: Would it be better to make `collectAssertions(...)` fail the explanatory group when the assertionCreator doesn't create any assertions?

robstoll: sounds like a reasonable thing, that an explanatory group can fail is rather a new thing

wordhou: I realize I'm not actually sure how to modify the builder to accomplish this. 

I'm looking at `AssertionOptionExplanatoryExtensions.kt` and I'm not sure how to get the extension method `.collectAssertions` to produce a failing final step.

My naive attempt:
```kotlin
fun <T, G : ExplanatoryAssertionGroupType, R> AssertionsOption<G, R>.collectAssertions(
    container: AssertionContainer<*>,
    maybeSubject: Option<T>,
    assertionCreator: Expect<T>.() -> Unit
): R {
    val collectingExpect = CollectingExpect<T>(None, container.components)
    // not using addAssertionsCreatedBy on purpose so that we don't append a failing assertion
    collectingExpect.assertionCreator()
    return withAssertions(container.collectForCompositionBasedOnSubject(maybeSubject, assertionCreator))
        .let {
            if(collectingExpect.getAssertions().isEmpty()) it.failing
            else it
        }
}
```
has issues because the return type is the generic type `R` and it doesn't recognize `.failing`. I'm not really understanding here how the builder pattern is implemented here with the generic `AssertionsOption<G,R>` type.

robstoll: You need to restrict R with an upper bound of ExplanatoryGroup.FinalStep.kt in AssertionOptionExplanatoryExtensions.kt.

wordhou: With the above type signature the compiler complained about two things:
* It expects the method to return `R` but the inferred type is `ExplanatoryGroup.FinalStep`
* Elsewhere in the code, in other places, it complains about the upper bound `R : ExplanatoryGroup.FinalStep`, saying `ExplanatoryAssertionGroupFinalStep` isn't a subtype of `ExplanatoryGroup.FinalStep`.

I had to use the following to satisfy the compiler.
```kotlin
fun <T, G : ExplanatoryAssertionGroupType, R: ExplanatoryAssertionGroupFinalStep> AssertionsOption<G, R>.collectAssertions(
    container: AssertionContainer<*>,
    maybeSubject: Option<T>,
    assertionCreator: Expect<T>.() -> Unit
): ExplanatoryAssertionGroupFinalStep {
  // ...
}
```

robstoll:
@wordhou sorry, I am quite busy at the moment, going to look into it hopefully soon

logic/atrium-logic-common/src/main/kotlin/ch/tutteli/atrium/logic/creating/collectors/AssertionsOptionExplantoryExtensions.kt
```diff
@@ -3,18 +3,35 @@ package ch.tutteli.atrium.logic.creating.collectors
 import ch.tutteli.atrium.assertions.AssertionGroup
 import ch.tutteli.atrium.assertions.ExplanatoryAssertionGroupType
 import ch.tutteli.atrium.assertions.builders.AssertionsOption
+import ch.tutteli.atrium.assertions.builders.ExplanatoryAssertionGroupFinalStep
```
robstoll, line 6: this deprecation needs to be suppressed as well (add deprecation to `@file`) this should fix the build

robstoll (approved):

PR #955, opened 2021-07-13
Show more than first line when using mocha test runner

### LINKED ISSUE #691: show more than first line when using mocha test runner
*Platform* (all, jvm, js): js

## Code related feature
It seems like mocha or maybe also the gradle plugin of Jetbrains is configured per default in a way that we see only the first line of an error in the terminal. We see the full message in the resulting gradle test report html though.

It would be nice, if we would see the whole message and not just the first line.
I am totally not sure how to do this, can only give the following hint, that it might be we can configure mocha somehow via the following (in samples/js/build.gradle)

```
kotlin {
    js {
        nodejs{
            testTask{
                useMocha{
                        // config here
                }
            }
        }
    }
}
```
I suggest you let /samples/js/src/test/kotlin/org/atriumlib/samples/mocha/SampleJsTest.kt fail on purpose (e.g. set `toBe(2)`) and run the sample afterwards:
```
cd samples/js
./gradlew buil
```

wordhou:
Resolves #691, adding settings to the `build.gradle.kts` to show the full exception when an assertion fails.

### DIFF
--- a/samples/js/build.gradle.kts
+++ b/samples/js/build.gradle.kts
@@ -1,6 +1,8 @@
 // Example project to show how to use Atrium in combination with mocha
 // For more information on how to setup Atrium for a JS project -> https://github.com/robstoll/atrium#js
 
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat
+
 // for infix-api -> change to 'atrium-infix-en_GB-js'
 val atriumApi = "atrium-fluent-en_GB-js"
 val atriumVersion = "0.16.0"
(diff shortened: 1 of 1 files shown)


samples/js/build.gradle.kts
```diff
@@ -28,6 +30,12 @@ version = "0.0.1"
 
 kotlin {
     js {
-        nodejs()
+        nodejs {
+            testTask {
+                testLogging {
+                    exceptionFormat = TestExceptionFormat.FULL // Show full exception when an assertion fails
```
wordhou, line 36: We could also add in the following options for explanation sake and to be consistent with the other sample `build.gradle.kts`.

```
    testLogging {
        exceptionFormat = TestExceptionFormat.FULL
        showExceptions = true        // defaults to true
        showCauses = true            // defaults to true
        showStackTraces = true       // defaults to true
    }
```

robstoll: good idea, let's do this

Issue #1039, opened 2021-10-29
add samples for Map.size to api-fluent and api-infix


robstoll:
*Platform* (all, jvm, js): all
*Extension* (none, kotlin 1.3): none

## Code related feature
See other samples in MapFeatureExtractorSamples (e.g. valuesFeature and values) we want the same for the function size

- [ ] add sizeFeature and size to MapFeatureExtractorSamples in api-fluent
- [ ] add `@sample` to the corresponding functions in mapFeatureExtractors.kt in api-fluent
- [ ] add sizeFeature and size to MapFeatureExtractorSamples in api-infix
- [ ] add `@sample` to the corresponding functions in mapFeatureExtractors.kt in api-infix

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime, we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/main/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

ashwini-desai:
@robstoll : I'll work on this

ashwini-desai:
@robstoll : Raised a PR

PR #1041, opened 2021-10-31
Add samples for Map.size

ashwini-desai:
PR for [this](https://github.com/robstoll/atrium/issues/1039) issue.


robstoll:
@ashwini-desai I am already merging so that this PR still counts for hacktoberfest in case you take part. 
Would be nice if you could align the comments so that they are one another. You can do this in a second PR.

In this sense, thanks for your first contribution to Atrium 🙂👍

PR #1119, opened 2022-04-07
introduce InAnyOrderOnlyReportingOptions

robstoll:
On one hand for contains.inAnyOrder but on the other also for
contains.inOrder.only.grouped.inAnyOrder. There both should be
configurable, the grouped reporting as well as the in-group reporting
options

moreover:
- deprecate InOrderOnlyReportingOptions.numberOfElementsInSummary,
  instead provide showOnlyFailingIfMoreExpectedElementsThan
- only run js test tasks in CI (as they currently re-run even if they
  are successful)


robstoll:
also add reportOptions to inAnyOrder.only not yet entirely done, there are some missing specs for mapLike.entriesOf

Issue #1161, opened 2022-06-27
rewrite misc/tools/atrium-samples-test/build.gradle to build.gradle.kts


robstoll:
*Platform* (all, jvm, js): all
*Extension* (none, kotlin 1.3): none

## Code related feature
Would be nicer to have the build file in Kotlin instead of groovy

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime, we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/main/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

Procrasteenator:
Hi, I would like to work on this issue but I am a beginner and haven't worked on writing Gradle files before, it would be great if you could guide me. 
Thank You

robstoll:
@Procrasteenator maybe not the best issue to start with in this case. Take a look at 1046. 
If you want to continue with this one, then read https://kotlinlang.org/docs/multiplatform-add-dependencies.html and similar sites and compare groovy and Kotlin.

agataporwit:
I have some Kotlin experience and would like to challange myself. Would that be a good place to start ? I am open to any suggestions.

robstoll:
@agataporwit give it a go, you will quickly find out if your list or not I guess. Ask if you need help 🙂

robstoll:
@agataporwit There is already a PR for this issue by @stexxe which I just merged. 
I hope you have not yet started with your work. In case you have, then you might be interested to compare your work with that of stexxe's see https://github.com/robstoll/atrium/pull/1163
There are still other good first issues:
https://github.com/robstoll/atrium/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22

Stexxe:
I'll work on this

PR #1163, opened 2022-07-06
Migrate build file from Groovy to Kotlin DSL of samples-composite project

Stexxe:
Fixes https://github.com/robstoll/atrium/issues/1161


robstoll (approved):

robstoll:
@Stexxe thanks for your first contribution to Atrium :slightly_smiling_face: :+1: 

Please, next time, add the comment `I'll work on this` to the corresponding issue before you start working on it. There was already someone else working on it which is kind of a pity. Anyway... how about a next one?

Stexxe:
@robstoll, I don't disturb someone else's work, and it's up to you to decide to merge my pull request or to wait for another one. I don't like to comment like this because, in the end, I may not finish it, so in my opinion, it's better to just deliver a solution, especially for small tasks like this. I can write `I'll work on this` just before I create a PR if you want.

robstoll:
@Stexxe I see your point. Would be good if you could write a comment, only this way I can assign the issue to you. You might want to do this also for #1161 so that I can assign it to you

Issue #1129, opened 2022-04-24
add tests for reportOptions to MapToContainInOrderOnlyKeyValueExpectationsSpec


robstoll:
*Platform* (all, jvm, js): all
*Extension* (none, kotlin 1.3): none

## Code related feature

Following the things you need to do:

- [ ] see TODOs in MapToContainInOrderOnlyKeyValueExpectationsSpec 

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime, we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/main/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

simonNozaki:
Hi @robstoll ,

I'll work on this. I'm new to open source projects, so would you support me where to start?

robstoll:
@simonNozaki sure, I guess you had already a look at [Your first code contribution](https://github.com/robstoll/atrium/blob/main/.github/CONTRIBUTING.md#your-first-code-contribution). Note that adding a sample would probably be an easier issue to start with and as you are new, I recommend you work on one of those:
https://github.com/robstoll/atrium/issues?q=is%3Aopen+label%3A%22good+first+issue%22+sample

simonNozaki:
@robstoll 
Thank you for your replying!

>  Note that adding a sample would probably be an easier issue to start with and as you are new
Yes, so I'll work on this:
https://github.com/robstoll/atrium/issues/1045

simonNozaki:
@robstoll ,
Thanks to assign me to this issue. Can I add some methods or utilities to the base class of `MapToContainInOrderOnlyKeyValueExpectationsSpec` , `MapLikeToContainFormatSpecBase` ?

`IterableToContainSpecBase` (the base class of `IterableToContainInOrderOnlyValuesExpectationsSpec`) has those, so I suppose that it's better to match its example.

robstoll:
Sure go ahead

simonNozaki:
Hi, @robstoll 

[...]

PR #1171, opened 2022-07-15
Feature/add specs maptocontain

simonNozaki:
# Outline
Add some specs to `MapToContainInOrderOnlyKeyValueExpectationsSpec` .

I made those reference to `IterableToContainInOrderOnlyValuesExpectationsSpec` .

# Issue
#1129 

### DIFF
--- a/misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
+++ b/misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
@@ -268,11 +268,36 @@ abstract class MapToContainInOrderOnlyKeyValueExpectationsSpec(
                     }.toThrow<AssertionError> {
                         message {
                             toContainSize(2, 3)
+                            elementSuccess(0, "1", "a", "1")
+                            elementFailing(0, "2", "a", "1")
                             elementOutOfBound(3, "d", "4")
-                            elementSuccess(0, "a", "a", "1")
                         }
                     }
                 }
+                it("shows only failing with report option `showOnlyFailingIfMoreExpectedElementsThan(2)` because there are 3") {
+                    expect {
+                        expect(map).toContainFun(
+                            keyValue("a") {toEqual(1)},
+                            keyValue("b") {toEqual(2)},
+                            keyValue("c") {toEqual(3)},
+                            report = { showOnlyFailingIfMoreExpectedElementsThan(2) }
+                        )
+                    }.toThrow<AssertionError> {
+                        message {
+                            toContainSize(2, 3)
+                            elementSuccess(0, "1", "a", "1")
+                            elementFailing(0, "2", "a", "1")
+                            elementOutOfBound(3, "d", "4")
+                        }
+                    }
+                }
+                it("shows summary with report option `showOnlyFailingIfMoreExpectedElementsThan(2)` because there are 2") {
+
+                }
+                it("shows summary without report option if there are 10 expected elements because default for showOnlyFailingIfMoreExpectedElementsThan is 10") {
+
+                }
+                it("shows only failing without report option if there are 11 expected elements because default for showOnlyFailingIfMoreExpectedElementsThan is 10") {
             }
         }
     }
--- a/misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
+++ b/misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
@@ -7,7 +7,6 @@ import ch.tutteli.atrium.logic.creating.iterablelike.contains.reporting.InOrderO
 import ch.tutteli.atrium.specs.*
 import ch.tutteli.atrium.specs.integration.IterableToContainSpecBase.Companion.nonNullableCases
 import ch.tutteli.atrium.specs.integration.IterableToContainSpecBase.Companion.nullableCases
-import org.spekframework.spek2.style.specification.Suite
 
 abstract class MapToContainInOrderOnlyKeyValueExpectationsSpec(
     keyWithValueAssertions: MFun3<String, Int, Expect<Int>.() -> Unit, InOrderOnlyReportingOptions.() -> Unit>,
@@ -257,6 +256,24 @@ abstract class MapToContainInOrderOnlyKeyValueExpectationsSpec(
 
         context("report options") {
             //TODO #1129 add tests, see IterableToContainInOrderOnlyValuesExpectationsSpec -> report options
+            context("") {
+                it("shows only failing with report option `showOnlyFailing`") {
+                    expect {
+                        expect(map).toContainFun(
+                            keyValue("a") {toEqual(1)},
+                            keyValue("b") {toEqual(2)},
+                            keyValue("c") {toEqual(3)},
+                            report = { showOnlyFailing() }
+                        )
+                    }.toThrow<AssertionError> {
+                        message {
+                            toContainSize(2, 3)
+                            elementOutOfBound(3, "d", "4")
+                            elementSuccess(0, "a", "a", "1")
+                        }
+                    }
+                }
+            }
         }
     }
 
--- a/misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
+++ b/misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
@@ -260,9 +260,9 @@ abstract class MapToContainInOrderOnlyKeyValueExpectationsSpec(
                 it("shows only failing with report option `showOnlyFailing`") {
                     expect {
                         expect(map).toContainFun(
-                            keyValue("a") {toEqual(1)},
-                            keyValue("b") {toEqual(2)},
-                            keyValue("c") {toEqual(3)},
+                            keyValue("a") { toEqual(1) },
+                            keyValue("b") { toEqual(2) },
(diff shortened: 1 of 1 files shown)


misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
```diff
@@ -257,6 +256,87 @@ abstract class MapToContainInOrderOnlyKeyValueExpectationsSpec(
...
+                        expect(map).toContainFun(
+                            keyValue("a") { toEqual(1) },
+                            keyValue("b") { toEqual(2) },
+                            keyValue("c") { toEqual(3) },
+                            report = { showOnlyFailing() }
+                        )
+                    }.toThrow<AssertionError> {
+                        message {
+                            toContainSize(2, 3)
+                            elementSuccess(0, "1", "a", "1")
```
robstoll, line 271: > shows only failing with report option

You are not expecting the contrary here. You should check that successful elements are not in the error report

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
```diff
@@ -257,6 +256,87 @@ abstract class MapToContainInOrderOnlyKeyValueExpectationsSpec(
...
+                    expect {
+                        expect(map).toContainFun(
+                            keyValue("a") { toEqual(1) },
+                            keyValue("b") { toEqual(2) },
+                            keyValue("c") { toEqual(3) },
+                            report = { showOnlyFailingIfMoreExpectedElementsThan(2) }
+                        )
+                    }.toThrow<AssertionError> {
+                        message {
+                            elementSuccess(0, "1", "a", "1")
```
robstoll, line 287: same same

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
```diff
@@ -257,6 +256,87 @@ abstract class MapToContainInOrderOnlyKeyValueExpectationsSpec(
...
+                it("shows summary with report option `showOnlyFailingIfMoreExpectedElementsThan(2)` because there are 2") {
+                    expect {
+                        expect(map).toContainFun(
+                            keyValue("a") { toEqual(1) },
+                            report = { showOnlyFailingIfMoreExpectedElementsThan(1) }
+                        )
+                    }.toThrow<AssertionError> {
+                        message {
+                            toContainSize(2, 1)
+                            elementSuccess(0, "1", "a", "1")
```
robstoll, line 303: I guess you got it the wrong way, showOnlyFailingIfMoreExpectedElementsThan(2) is not met, hence it should show all successful expectations and all failing expectations

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
```diff
@@ -257,6 +256,87 @@ abstract class MapToContainInOrderOnlyKeyValueExpectationsSpec(
...
+                        }
+                    }
+                }
+                it("shows summary without report option if there are 2 expected elements because default for showOnlyFailingIfMoreExpectedElementsThan is 2") {
```
robstoll, line 307: default is 10 not 2. you need to state 10 expectations to be sure that it still shows the summary

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
```diff
@@ -257,6 +256,87 @@ abstract class MapToContainInOrderOnlyKeyValueExpectationsSpec(
...
+                    }.toThrow<AssertionError> {
+                        message {
+                            elementSuccess(0, "1", "a", "1")
+                            elementFailing(0, "2", "a", "1")
+                            elementOutOfBound(3, "d", "4")
+                            toContainSize(2, 3)
+                        }
+                    }
+                }
+                it("shows only failing without report option if there are 3 expected elements because default for showOnlyFailingIfMoreExpectedElementsThan is 2") {
```
robstoll, line 323: same as above, the default is 10, so you should make 11 expectations about entries to see if only the failing are shown

robstoll (changes requested):

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
simonNozaki: Hi, @robstoll 

Thank you for reviewing. Asking questions you once before, how can I check that successful elements are not in the error report? I could not find the way to check specs I wrote...

robstoll: Sorry, did not get that this question was not solved yet. I remember that you were asking whether you can copy methods from IterableToContainsSepcBase to MapLikeToContainSpecBase. So yes, you totally can. You can for instance copy notToContainElement from there and adopt to map. I suggest we name the method `notToContainEntry` and it would suffice if it takes the key i.e it would look like follows:  
```
fun Expect<String>.notToContainEntry(key: String): Expect<String> = ...
```
I left out the body on purpose, let me know if you don't want to figure this out yourself, I can also provide it.

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
simonNozaki: @robstoll 
> default is 10 not 2
I could not understand this... Why is the default value 10, not 2? Test data for spec is optional, I thought.

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
simonNozaki: >  I suggest we name the method notToContainEntry and it would suffice if it takes the key i.e it would look like follows:

Thanks to provide some example.  I understood that I could write expectation by `toContainFun` , but sould I create something like `notToContainEntry` ?

robstoll: Yes, you should create notToContainEntry in MapLikeToContainSpecBase

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
```diff
@@ -257,6 +256,87 @@ abstract class MapToContainInOrderOnlyKeyValueExpectationsSpec(
...
+                            elementOutOfBound(3, "d", "4")
+                            toContainSize(2, 3)
+                        }
+                    }
+                }
+                it("shows summary with report option `showOnlyFailingIfMoreExpectedElementsThan(2)` because there are 2") {
+                    expect {
+                        expect(map).toContainFun(
+                            keyValue("a") { toEqual(1) },
+                            report = { showOnlyFailingIfMoreExpectedElementsThan(1) }
```
robstoll, line 298: Didn't see this before, you should do what you describe in `it` i.e. pass 2 and not 1 to showOnlyFailingIfMoreExpectedElementsThan
```suggestion
                            report = { showOnlyFailingIfMoreExpectedElementsThan(2) }
```
Accordingly, you need to add two entry expectations to `toContainFun` to meet the criteria

---

Alternatively you can rewrite the `it` to:
```
it("shows summary with report option `showOnlyFailingIfMoreExpectedElementsThan(1)` because there are 1") {
```

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
```diff
@@ -257,6 +256,87 @@ abstract class MapToContainInOrderOnlyKeyValueExpectationsSpec(
 
         context("report options") {
             //TODO #1129 add tests, see IterableToContainInOrderOnlyValuesExpectationsSpec -> report options
+            context("") {
```
robstoll, line 259: describe here what map is the subject, i.e. use 
```suggestion
            context("map $map") {
```

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/MapToContainInOrderOnlyKeyValueExpectationsSpec.kt
robstoll: `report = {...}` is optional exactly and you did not specify it here. If you leave it out then the default kicks in which is kind of:
`report = { showOnlyFailingIfMoreExpectedElementsThan(10) }`

robstoll:
@simonNozaki closing this, will take your commits as basis for a new PR. Thanks for the pre-work

PR #1203, opened 2022-10-09
WIP PR - attempted import in comparableExpectations.kt

JordanllHarper:


apis/fluent-en_GB/atrium-api-fluent-en_GB/src/commonMain/kotlin/ch/tutteli/atrium/api/fluent/en_GB/comparableExpectations.kt
```diff
@@ -2,7 +2,7 @@ package ch.tutteli.atrium.api.fluent.en_GB
 
 import ch.tutteli.atrium.creating.Expect
 import ch.tutteli.atrium.logic.*
-
+import java.time.chrono.ChronoLocalDateTime //attempted import
```
robstoll, line 5: That's the wrong way round. You should add the new overloads in /apis/fluent-en_GB/atrium-api-fluent-en_GB/src/jvmMain/kotlin/ch/tutteli/atrium/api/fluent/en_GB/chronoZonedDateTimeExpectations.kt  (note the **jvmMain**). You cannot import java... in commonMain

JordanllHarper: Ah right of course. Thank you so much for the help!

PR #1233, opened 2022-10-25
Migrate files from DescribeSpec to FunSpec

Dark-Knight11:
js target doesn't support Describe Spec of Kotest hence we have to migrate the spek to Fun spec before enabling js target again. 
I am migrating the already migrated DescribeSpec files to FunSpec.


robstoll:
@Dark-Knight11 looks good IMO, do you want to do something in addition (as you named it WIP which stands for work in progress)? Otherwise we can merge it :+1:

Dark-Knight11:
yeah I'm working on migration of other files as well.
We can merge it if required. I'll do other files in separate PR.

PR #1235, opened 2022-10-27
Migrate DescriptiveWithBasedOnSubjectSpec from spek to kotest 

jakubriegel:
Refers to https://github.com/robstoll/atrium/issues/1194

### DIFF
--- a/misc/atrium-specs/build.gradle.kts
+++ b/misc/atrium-specs/build.gradle.kts
@@ -24,6 +24,7 @@ kotlin {
                 apiWithExclude("io.kotest:kotest-runner-junit5:$kotestVersion")
                 // necessary in order that intellij sees the io.kotest symbols (runner-junit5 actually already depends on it)
                 apiWithExclude("io.kotest:kotest-framework-api:$kotestVersion")
+                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.3.9")
             }
         }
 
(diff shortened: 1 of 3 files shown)


misc/atrium-specs/build.gradle.kts
```diff
@@ -24,6 +24,7 @@ kotlin {
                 apiWithExclude("io.kotest:kotest-runner-junit5:$kotestVersion")
                 // necessary in order that intellij sees the io.kotest symbols (runner-junit5 actually already depends on it)
                 apiWithExclude("io.kotest:kotest-framework-api:$kotestVersion")
+                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.3.9")
```
jakubriegel, line 27: I'm sure if this should be done this way in Atrium's dependencies setup. Feel free to suggest better approach.

Without adding coroutines explicitly as implementation here, gradle is not able to compile the code.

jakubriegel:
@robstoll If review will take more time, is it possible, that we add `HACKTOBERFEST-ACCEPTED` label here?

misc/atrium-specs/build.gradle.kts
robstoll: All fine, I can still improve after merge

robstoll (approved):

robstoll:
@jakubriegel looks good to me, thanks for the contribution

PR #1264, opened 2022-12-10
Kt 54674 -> pull request for discussing if statement check

JordanllHarper:
Pull request on this branch because it's what I'm working with so far


logic/extensions/atrium-logic-kotlin_1_3/src/commonMain/kotlin/ch/tutteli/atrium/logic/kotlin_1_3/impl/DefaultResultAssertions.kt
```diff
@@ -29,16 +26,36 @@ class DefaultResultAssertions : ResultAssertions {
...
+            container: AssertionContainer<out Result<*>>,
+            expectedType: KClass<TExpected>) : SubjectChangerBuilder.ExecutionStep<Throwable?, TExpected> {
+
+
         println("heeere and subject is: ${container.maybeSubject.getOrElse { "no subject?" }}")
-        return container.manualFeature(EXCEPTION) {
-            println("suddenly no longer a Failure?: $this")
+
+        return container.manualFeature(EXCEPTION) {println("suddenly no longer a Failure?: $this")
             exceptionOrNull()
```
robstoll, line 40: check needs to be in here and you need to check, if exceptionOrNull() != null and then check that `container.maybeSubject.map { exceptionOrNull() }` was not null as well

PR #1312, opened 2023-01-28
Default result assertions fix 

JordanllHarper:
Unsure on how to perform the unwrapping part of the process.

### DIFF
--- a/logic/extensions/atrium-logic-kotlin_1_3/src/commonMain/kotlin/ch/tutteli/atrium/logic/kotlin_1_3/impl/DefaultResultAssertions.kt
+++ b/logic/extensions/atrium-logic-kotlin_1_3/src/commonMain/kotlin/ch/tutteli/atrium/logic/kotlin_1_3/impl/DefaultResultAssertions.kt
@@ -38,9 +38,10 @@ class DefaultResultAssertions : ResultAssertions {
 
 
             if(exceptionOrNull() == null && container.maybeSubject.map { exceptionOrNull() } != null){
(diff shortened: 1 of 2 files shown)


logic/extensions/atrium-logic-kotlin_1_3/src/commonMain/kotlin/ch/tutteli/atrium/logic/kotlin_1_3/impl/DefaultResultAssertions.kt
```diff
@@ -34,7 +34,20 @@ class DefaultResultAssertions : ResultAssertions {
         container: AssertionContainer<out Result<*>>,
         expectedType: KClass<TExpected>
     ):  SubjectChangerBuilder.ExecutionStep<Throwable?, TExpected> =
-        container.manualFeature(EXCEPTION) { exceptionOrNull() }.transform().let { previousExpect ->
+        container.manualFeature(EXCEPTION) {
+
+
+            if(exceptionOrNull() == null && container.maybeSubject.map { exceptionOrNull() } != null){
+                //unwrap somehow (?) todo
```
robstoll, line 41: Checkout https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/ to get to know Result better.
Also, `container.maybeSubject.map { exceptionOrNull() }` will never be null but always `Option<Throwable?>` you need to use `getOrElse { null }` in addition to turn it into a `Throwable?`

PR #1331, opened 2023-02-16
Update AssertJ syntax in README.md

vlsi:
AssertJ does provide SoftAssertions.assertSoftly(block) so users don't accidentally forget calling `assertAll`. Unfortunately, they do not provide Kotlin wrapper yet, so `softly.assertThat(..)` have to be duplicated.


vlsi:
>Maybe worth fixing there as well (in case you are a maintainer)

I'm not 🤷

Issue #1359, opened 2023-03-17
"following elements were mismatched" hides the reason for the failure


vlsi:
### Affected Version

0.18.0

### API

fluent-en_GB

### Platform

jvm

### Kotlin Version

1.8

### How to reproduce the problem?

```kotlin
data class User(val login: String, val password: String)

expect(listOf(User("joe", "qwerty"), User("q", "qwerty1"))) {
    toHaveElementsAndAll {
        feature("login", { login })
            .feature("length", { length })
            .because("login should be longer than one letter") {
                toBeGreaterThan(1)
            }
        feature("password", { password })
            .because("password should be secure") {
                notToEqual("qwerty")
            }
    }
}
```

=>

```
I expected subject: [User(login=joe, password=qwerty), User(login=q, password=qwerty1)]        (java.util.Arrays.ArrayList <1453882535>)
◆ elements need all: 
    » ▶ login: 
        ◾ ▶ length: 
            ◾ to be greater than: 1        (kotlin.Int <2086658927>)
            ℹ because: login should be longer than one letter
    » ▶ password: 
        ◾ not to equal: "qwerty"        <151959367>
        ℹ because: password should be secure
    ❗❗ following elements were mismatched: 
       ⚬ index 0: User(login=joe, password=qwerty)        (ch.tutteli.atrium.api.fluent.en_GB.samples.ThrowableFeatureExtractorSamples$messageFeature2$User <1649737945>)
       ⚬ index 1: User(login=q, password=qwerty1)        (ch.tutteli.atrium.api.fluent.en_GB.samples.ThrowableFeatureExtractorSamples$messageFeature2$User <1943864595>)
```

### Describe the bug

The output prints mismatched records, however, it does not report which expectations failed.

I see how printing all (failed or not) expectations for every element might produce a lot of output, so I would suggest that "following elements were mismatched" should hide the first level of successful expectations.

The following would be better:

```
I expected subject: [User(login=joe, password=qwerty), User(login=q, password=qwerty1)]        (java.util.Arrays.ArrayList <1453882535>)
◆ elements need all: 
    » ▶ login: 
        ◾ ▶ length: 
            ◾ to be greater than: 1        (kotlin.Int <2086658927>)
            ℹ because: login should be longer than one letter
    » ▶ password: 
        ◾ not to equal: "qwerty"        <151959367>
        ℹ because: password should be secure
    ❗❗ following elements were mismatched: 
[...]

PR #1362, opened 2023-03-18
fix: display mismatch reasons in "following elements were mismatched"

vlsi:
Fixes https://github.com/robstoll/atrium/issues/1359

### DIFF
--- a/atrium-core/src/commonMain/kotlin/ch/tutteli/atrium/creating/impl/BaseExpectImpl.kt
+++ b/atrium-core/src/commonMain/kotlin/ch/tutteli/atrium/creating/impl/BaseExpectImpl.kt
@@ -63,8 +63,7 @@ abstract class BaseExpectImpl<T>(
             representationInsteadOfFeature?.let { provider ->
                 maybeSubject.fold({ null }) { provider(it) }
             } ?: maybeSubject.getOrElse {
-                // a RootExpect without a defined subject is almost certain a bug
-                Text(SHOULD_NOT_BE_SHOWN_TO_THE_USER_BUG)
+                Text.EMPTY
             }
     }
 }
--- a/atrium-core/src/commonMain/kotlin/ch/tutteli/atrium/reporting/text/impl/TextFeatureAssertionGroupFormatter.kt
+++ b/atrium-core/src/commonMain/kotlin/ch/tutteli/atrium/reporting/text/impl/TextFeatureAssertionGroupFormatter.kt
@@ -61,6 +61,6 @@ class TextFeatureAssertionGroupFormatter(
         parameterObject: AssertionFormatterParameterObject
     ) : AssertionGroup by assertionGroup {
         override val description: Translatable = newName
-        override val representation: Any = if(parameterObject.isNotInExplanatoryFilterGroup()) assertionGroup.representation else Text.EMPTY
+        override val representation: Any = assertionGroup.representation
     }
 }
--- a/logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
+++ b/logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
@@ -4,6 +4,7 @@ import ch.tutteli.atrium.assertions.Assertion
 import ch.tutteli.atrium.assertions.AssertionGroup
 import ch.tutteli.atrium.assertions.builders.*
 import ch.tutteli.atrium.core.Option
+import ch.tutteli.atrium.core.Some
 import ch.tutteli.atrium.core.getOrElse
 import ch.tutteli.atrium.creating.AssertionContainer
 import ch.tutteli.atrium.creating.Expect
@@ -20,8 +21,10 @@ import ch.tutteli.atrium.logic.creating.iterable.contains.steps.impl.EntryPointS
 import ch.tutteli.atrium.logic.creating.iterable.contains.steps.notCheckerStep
 import ch.tutteli.atrium.logic.creating.transformers.FeatureExtractorBuilder
 import ch.tutteli.atrium.logic.creating.typeutils.IterableLike
+import ch.tutteli.atrium.reporting.Text
 import ch.tutteli.atrium.reporting.translating.Translatable
 import ch.tutteli.atrium.reporting.translating.TranslatableWithArgs
+import ch.tutteli.atrium.translations.DescriptionBasic
 import ch.tutteli.atrium.translations.DescriptionBasic.NOT_TO_HAVE
 import ch.tutteli.atrium.translations.DescriptionBasic.TO_HAVE
 import ch.tutteli.atrium.translations.DescriptionIterableLikeExpectation.*
@@ -135,9 +138,25 @@ class DefaultIterableLikeAssertions : IterableLikeAssertions {
 
         val assertions = ArrayList<Assertion>(2)
         assertions.add(createExplanatoryAssertionGroup(container, assertionCreatorOrNull))
-        val mismatches = createIndexAssertions(list) { (_, element) ->
-            mismatchIf(allCreatedAssertionsHold(container, element, assertionCreatorOrNull))
+        val mismatches = list.mapIndexedNotNull { index, e ->
+            val assertion = if (assertionCreatorOrNull != null) {
+                container.collectBasedOnSubject(Some(e as E), assertionCreatorOrNull)
+            } else {
+                // TODO: assertionCreatorOrNull should be non-nullable, and this "to be null" should be a part of expectation
+                assertionBuilder.createDescriptive(DescriptionBasic.TO_BE, Text.NULL) {
+                    e == null
+                }
+            }
+
+            assertion.takeIf { mismatchIf(it.holds()) }?.let {
+                assertionBuilder
+                    .feature
+                    .withDescriptionAndRepresentation(TranslatableWithArgs(INDEX, index), e)
+                    .withAssertion(it)
+                    .build()
+            }
         }
+
         if (mismatches.isNotEmpty()) assertions.add(createExplanatoryGroupForMismatches(mismatches))
 
         decorateAssertion(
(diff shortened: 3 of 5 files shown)


logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
```diff
@@ -135,9 +138,25 @@ class DefaultIterableLikeAssertions : IterableLikeAssertions {
...
-        val mismatches = createIndexAssertions(list) { (_, element) ->
-            mismatchIf(allCreatedAssertionsHold(container, element, assertionCreatorOrNull))
+        val mismatches = list.mapIndexedNotNull { index, e ->
+            val assertion = if (assertionCreatorOrNull != null) {
+                container.collectBasedOnSubject(Some(e as E), assertionCreatorOrNull)
+            } else {
+                // TODO: assertionCreatorOrNull should be non-nullable, and this "to be null" should be a part of expectation
+                assertionBuilder.createDescriptive(DescriptionBasic.TO_BE, Text.NULL) {
+                    e == null
+                }
```
vlsi, line 148: It is strange to handle `e == null` here. I wish `assertionCreatorOrNull` was non-nullable.

vlsi:
The fix makes 

```kotlin
        expect(listOf(User("joe", "qwerty"), User("j", "qwerty1"))) {
            toHaveElementsAndAll {
                feature("login", { login })
                    .feature("size", { length })
                    .because("login should be longer than one letter") {
                        toBeGreaterThan(1)
                    }
                feature("password", { password })
                    .because("password should be secure") {
                        notToEqual("qwerty")
                    }
            }
        }
```

to produce

```
I expected subject: [User(login=joe, password=qwerty), User(login=j, password=qwerty1)]        (java.util.Arrays.ArrayList <1937662030>)
◆ elements need all: 
    » ▶ login: 
        ◾ ▶ size: 
            ◾ to be greater than: 1        (kotlin.Int <987792975>)
            ℹ because: login should be longer than one letter
    » ▶ password: 
        ◾ not to equal: "qwerty"        <952055826>
        ℹ because: password should be secure
    ❗❗ following elements were mismatched: 
       ⚬ ▶ index 0: User(login=joe, password=qwerty)        (ch.tutteli.atrium.api.fluent.en_GB.samples.ThrowableFeatureExtractorSamples$messageFeature2$User <1635068416>)
           ◾ ▶ login: "joe"        <1130906588>
               ◾ ▶ size: 3        (kotlin.Int <439323450>)
                   ◾ to be greater than: 1        (kotlin.Int <987792975>)
                   ℹ because: login should be longer than one letter
           ◾ ▶ password: "qwerty"        <952055826>
               ◾ not to equal: "qwerty"        <952055826>
               ℹ because: password should be secure
       ⚬ ▶ index 1: User(login=j, password=qwerty1)        (ch.tutteli.atrium.api.fluent.en_GB.samples.ThrowableFeatureExtractorSamples$messageFeature2$User <1133787119>)
           ◾ ▶ login: "j"        <44754983>
               ◾ ▶ size: 1        (kotlin.Int <987792975>)
                   ◾ to be greater than: 1        (kotlin.Int <987792975>)
                   ℹ because: login should be longer than one letter
           ◾ ▶ password: "qwerty1"        <1254614872>
               ◾ not to equal: "qwerty"        <952055826>
               ℹ because: password should be secure
```

logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
robstoll: I agree, this functionality could be refactored. Is due to some legacy behaviour introduced in Atrium 0.2.0 or the like and wouldn't be necessary any more. Going to create an issue for it

atrium-core/src/commonMain/kotlin/ch/tutteli/atrium/creating/impl/BaseExpectImpl.kt
```diff
@@ -63,8 +63,7 @@ abstract class BaseExpectImpl<T>(
             representationInsteadOfFeature?.let { provider ->
                 maybeSubject.fold({ null }) { provider(it) }
             } ?: maybeSubject.getOrElse {
-                // a RootExpect without a defined subject is almost certain a bug
-                Text(SHOULD_NOT_BE_SHOWN_TO_THE_USER_BUG)
+                Text.EMPTY
```
robstoll, line 66: Please don't change this in the end, if you saw this text in reporting, then it is most likely an implementation bug :wink:

atrium-core/src/commonMain/kotlin/ch/tutteli/atrium/reporting/text/impl/TextFeatureAssertionGroupFormatter.kt
```diff
@@ -61,6 +61,6 @@ class TextFeatureAssertionGroupFormatter(
         parameterObject: AssertionFormatterParameterObject
     ) : AssertionGroup by assertionGroup {
         override val description: Translatable = newName
-        override val representation: Any = if(parameterObject.isNotInExplanatoryFilterGroup()) assertionGroup.representation else Text.EMPTY
+        override val representation: Any = assertionGroup.representation
```
robstoll, line 64: please revert in the end

logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
```diff
@@ -135,9 +138,25 @@ class DefaultIterableLikeAssertions : IterableLikeAssertions {
...
+            }
+
+            assertion.takeIf { mismatchIf(it.holds()) }?.let {
+                assertionBuilder
+                    .feature
+                    .withDescriptionAndRepresentation(TranslatableWithArgs(INDEX, index), e)
+                    .withAssertion(it)
+                    .build()
+            }
```
robstoll, line 157: You should modify createIndexAssertions - I suggest you rename it to `createMismatch` as it is used only for this purpose

robstoll (commented):
The output already looks good :+1: but the implementation needs to be adopted

atrium-core/src/commonMain/kotlin/ch/tutteli/atrium/reporting/text/impl/TextFeatureAssertionGroupFormatter.kt
vlsi: What would be your suggestion then?

Apparently, the current implementation in Atrium is "whatever is displayed in explanatory goes WITHOUT representation".
At the same time, the only way to put "warning sign" is to wrap it in explanatory.

That creates a deadlock, and I think it is wrong that "explanatory" implies "remove all representations". I just do not understand why it removes the representations. I believe the ones who want to display message "without representation" should call the relevant APIs "without representation", and it should not be "explanatory" that hides representations.

atrium-core/src/commonMain/kotlin/ch/tutteli/atrium/creating/impl/BaseExpectImpl.kt
vlsi: Well, what is your recommendation then?
Currently, Atrium uses an explicit `None` subject in order to render `elements need all:` block.: https://github.com/robstoll/atrium/blob/65227add3c31604a64cf4ef86671eadc78be3bde/logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/containsHelpers.kt#L45

logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
vlsi: Of course, it should replace the current `createIndexAssertions`, however, I tried to implement something minimal to see what are other changes required.

One of the takeaways, for now, is the current implementation shows both successful and failing assertions, and it does not distinguish between successful and failing ones.
For instance `login: "joe"` was successful, however, it was displayed the same as `password: "qwerty"` which failed.

atrium-core/src/commonMain/kotlin/ch/tutteli/atrium/reporting/text/impl/TextFeatureAssertionGroupFormatter.kt
robstoll: >  I think it is wrong that "explanatory" implies "remove all representations"

The reason why we remove the representation is because usually we explain what expectation was defined without it being bound to a subject. For instance, the `◆ elements need all: ` part. The expectations shown there are not bound to a specific element in the Iterable.

> At the same time, the only way to put "warning sign" is to wrap it in explanatory.

I see now better where your request for a non-fluent assertionBuilder comes from :)

<hr/>

I see overlapping topics here. I think one way forward would be that we can define that a certain `Assertion` should be shown as is within an explanatory group, i.e. kind of escape the explanatory mechanism. This way we would also only show failing expectations etc. as it was not within an explanatory group. 

Another approach would be that we allow more freely what bullet points are used and then you would not use an explanatory group in your case but just a regular list group with another bullet point. 

Have to go will, think about it later on,

robstoll: We need to take the approach with escaping because otherwise we loose context. for this to work, we need:
- change from NonFiltering count to FilteringStack
- A new GroupAssertionType EscapeNonFiltering which behaves like an invisible group but additionally modifies the stack

robstoll:
closing this as the issue as such will be resolved with the new reporter based on Proof. The reporting for the case in the issue will look as follows:
```text
I expected subject : [User(login=joe, password=qwerty), User(login=q, password=qwerty1)]
✘ elements need all : 
  » ▶ login : 
      • ▶ length               : 
          • to be greater than : 1
  » ▶ password       : 
      • not to equal : "qwerty"
      ℹ️ because : password should be secure
  ❗❗ following elements were mismatched : 
     • index 0 : User(login=joe, password=qwerty)
       ✘ ▶ password       : "qwerty"
           ✘ not to equal : "qwerty"
           ℹ️ because : password should be secure
     • index 1 : User(login=q, password=qwerty1)
       ✘ ▶ login : "q"
           ✘ ▶ length               : should not be shown
               ✘ to be greater than : 1
```
Stay tuned...

PR #1416, opened 2023-04-26
add relocation pom for removing -en_GB

robstoll:

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -265,50 +265,50 @@ configure(subprojectsWithoutToolAndSmokeTestProjects) {
     val languageSuffixes = setOf("-en_GB", "-de_CH")
     if (languageSuffixes.any { name.contains(it) }) {
         configure<PublishingExtension> {
-            val pubs = publications.toList()
-            publications {
-                pubs
-                    .filterIsInstance<MavenPublication>()
-                    .forEach { pub ->
-                        val artifactIdWithoutLanguage = languageSuffixes.fold(pub.artifactId) { id, suffix ->
-                            id.replace(suffix, "")
+            publications.withType<MavenPublication>()
+                .matching { !it.name.endsWith("-relocation") }
+                .all {
+                    val artifactIdBeforeLanguageRemoval = artifactId
+                    val artifactIdWithoutLanguage = languageSuffixes.fold(artifactId) { id, suffix ->
+                        id.replace(suffix, "")
+                    }
+
+                    // TODO 1.1.0 consider to remove en_GB from the project name
+                    artifactId = artifactIdWithoutLanguage
+
+                    //TODO 1.1.0 remove again, consider to rename the project-name and directory
+                    // we did not have a -metadata jar before, so no need for a relocation publication
+                    if (!artifactId.endsWith("metadata")) {
+                        val oldArtifactId = if (artifactIdBeforeLanguageRemoval == project.name) {
+                            "${project.name}-common"
+                        } else if (artifactId.endsWith("-jvm")) {
+                            project.name
+                        } else {
+                            artifactIdBeforeLanguageRemoval
                         }
-                        pub.artifactId = artifactIdWithoutLanguage
 
-                        //TODO 1.1.0 remove again, consider to rename the project-name and directory
-                        // we did not have a -metadata jar before, so no need for a relocation publication
-                        if (!pub.artifactId.endsWith("metadata")) {
-                            val oldArtifactId = if (pub.artifactId == name) {
-                                "$name-common"
-                            } else if (pub.artifactId.endsWith("-jvm")) {
-                                name
-                            } else {
-                                pub.artifactId
-                            }
+                        println("old: $oldArtifactId, new: $artifactId")
+                        val pub = this
+                        publications.register<MavenPublication>("${pub.name}-relocation") {
+                            pom {
+                                // Old artifact coordinates
+                                groupId = pub.groupId
+                                artifactId = oldArtifactId
+                                version = pub.version
 
-                            pub.artifactId = artifactIdWithoutLanguage
-
-                            create<MavenPublication>("${pub.name}-relocation") {
-                                pom {
-                                    // Old artifact coordinates
-                                    groupId = pub.groupId
-                                    artifactId = oldArtifactId
-                                    version = pub.version
-
-                                    distributionManagement {
-                                        relocation {
-                                            // New artifact coordinates
-                                            groupId.set(pub.groupId)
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -258,13 +258,63 @@ configure(subprojectsWithoutToolAndSmokeTestProjects) {
     apply(plugin = "ch.tutteli.gradle.plugins.dokka")
     apply(plugin = "ch.tutteli.gradle.plugins.publish")
 
-    project.extensions.getByType<ch.tutteli.gradle.plugins.publish.PublishPluginExtension>().apply {
+    extensions.getByType<ch.tutteli.gradle.plugins.publish.PublishPluginExtension>().apply {
```
vlsi, line 261: configure<PublishPluginExtension> { ...

build.gradle.kts
```diff
@@ -258,13 +258,63 @@ configure(subprojectsWithoutToolAndSmokeTestProjects) {
...
         resetLicenses("EUPL-1.2")
     }
+
+    val languageSuffixes = setOf("-en_GB", "-de_CH")
+    if (languageSuffixes.any { name.contains(it) }) {
+        extensions.getByType<PublishingExtension>().apply {
```
vlsi, line 267: `configure<PublishingExtension> { ...`

build.gradle.kts
```diff
@@ -258,13 +258,63 @@ configure(subprojectsWithoutToolAndSmokeTestProjects) {
...
+            val pubs = publications.toList()
+            publications {
+                pubs
+                    .filterIsInstance<MavenPublication>()
```
vlsi, line 271: pubs.withType<MavenPublication>()

build.gradle.kts
robstoll: First of all, thanks for the review and the suggestions :slightly_smiling_face: :+1: 

Out of curiosity, why do you suggest to use `configure`? we don't use convention plugins (so far) and I don't want to magically refer to something different than the extension, so why not use it directly?

build.gradle.kts
robstoll: using `publications.toList()` and filterIsIsntance on purpose because I don't want that the newly created publication is re-processed (I think we would even run into a ConcurrentModificationException or the like)

build.gradle.kts
vlsi: Well, you do specify `ch.tutteli.gradle.plugins.publish.PublishPluginExtension` type here; the only possibility is that this type represents project extension.
Note: "convention plugins != project conventions", and `configure` has little to do with `convention plugins`.
With convention plugins, direct accessors should be used instead of referring to the extensions by names.

`configure<Type> { ... }` is exactly the syntax for that: https://github.com/gradle/gradle/blob/100a43bd3e0079db6225a8652cacb8784d652240/subprojects/kotlin-dsl/src/main/kotlin/org/gradle/kotlin/dsl/ProjectExtensions.kt#L82-L83

`configure<...>` is something that suggested in Gradle documentation for the cases when accessors are not available: https://docs.gradle.org/current/userguide/kotlin_dsl.html#sec:kotlin_using_standard_api

`extensions.getByType...apply.{` is pure noise.

build.gradle.kts
vlsi: Basically, whenever you have `.apply {` in `gradle.kts` it means there's something wrong going on.

build.gradle.kts
vlsi: >I don't want that the newly created publication is re-processed

It creates a fragile logic that forces the reader to determine the created publications' order.

It would be easier to reason about if the relocation was like "ok, whenever a new publication arrives, we add a corresponding relocation".
The scripts that expect a specific initialization order are fragile, and they are hard to maintain.

robstoll: if we use `withType` then behind the scene gradle does something along the line of
```
publications.forEach {
  ...
}
```
if we do a create within this forEach, then we run into a ConcurrentModificationException.
See for instance (random google hit: https://rollbar.com/blog/java-concurrentmodificationexception/)

build.gradle.kts
robstoll: > the only possibility is that this type represents project extension.

exactly. if you use `configure<...>` then the project convention (introduced via a plugin such as java-library) does not necessarily need to load the extension but could replace it by something custom. I don't like such magic things.

But I see that conventions are already deprecated at will be removed with 9.0.0. My guess is, that after the removal, this will always only load the extension. Therefore it does then the same thing and I agree that it is more gradle like to use `configure`, so I am going to rewrite it

build.gradle.kts
vlsi: The following works for me:

```kotlin
        configure<PublishingExtension> {
            publications.withType<MavenPublication>()
                .matching { !it.name.endsWith("-relocation") }
                .all {
                    val pub = this
                    val artifactIdWithoutLanguage = languageSuffixes.fold(artifactId) { id, suffix ->
                        id.replace(suffix, "")
                    }
                    artifactId = artifactIdWithoutLanguage
                    if (!pub.artifactId.endsWith("metadata")) {
                        val oldArtifactId = if (pub.artifactId == name) {
                            "$name-common"
                        } else if (pub.artifactId.endsWith("-jvm")) {
                            name
                        } else {
                            pub.artifactId
                        }

                        pub.artifactId = artifactIdWithoutLanguage

                        publications.create<MavenPublication>("${pub.name}-relocation") {
                            pom {
                                // Old artifact coordinates
                                groupId = pub.groupId
                                artifactId = oldArtifactId
                                version = pub.version

                                distributionManagement {
                                    relocation {
                                        // New artifact coordinates
                                        groupId.set(pub.groupId)
                                        artifactId.set(artifactIdWithoutLanguage)
                                        version.set(pub.version)
                                        message.set("artifactId has changed, removed -en_GB as we drop support for a translated report and jvm has now the suffix -jvm and common no longer has the suffix -common")
                                    }
                                }
                            }
                        }
                    }
                }
```

vlsi: An alternative option would be

```kotlin
// We want to create relocations for all the publications, so we wait for them to settle
afterEvaluate {
  val pubs = publications.withType<MavenPublication>().toList().
  // ...
}
```

However, it would be better to avoid `afterEvaluate` when possible.

robstoll: good to know, I also prefer the one without afterEvaluate. I first used withType but used the overload with the lamba and used an if inside instead of matching (is nicer). I wonder why I ran into the concurrent MidificationException. 🤔
anyway, will adopt the code and merge afterwards

PR #1428, opened 2023-05-06
Default result assertions fix 

JordanllHarper:
Implemented a fix which pasts the tests.
Using exception or null discarded inner exception value, so manually mapped to retain that value.
Let me know what I'm needing to do to improve the code

### DIFF
--- a/apis/fluent-en_GB/extensions/atrium-api-fluent-en_GB-kotlin_1_3/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/kotlin_1_3/samples/ResultExpectationSamples.kt
+++ b/apis/fluent-en_GB/extensions/atrium-api-fluent-en_GB-kotlin_1_3/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/kotlin_1_3/samples/ResultExpectationSamples.kt
@@ -4,6 +4,7 @@ import ch.tutteli.atrium.api.fluent.en_GB.*
 import ch.tutteli.atrium.api.fluent.en_GB.kotlin_1_3.toBeAFailure
 import ch.tutteli.atrium.api.fluent.en_GB.kotlin_1_3.toBeASuccess
 import ch.tutteli.atrium.api.verbs.internal.expect
+import kotlin.math.exp
 import kotlin.test.Test
 
 class ResultExpectationSamples {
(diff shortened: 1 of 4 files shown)


apis/fluent-en_GB/extensions/atrium-api-fluent-en_GB-kotlin_1_3/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/kotlin_1_3/samples/ResultExpectationSamples.kt
```diff
@@ -57,69 +58,85 @@ class ResultExpectationSamples {
...
+        val failure = Result.failure<Int>(IllegalArgumentException(message))
+
+
+
+        expect(failure) {
+            toBeAFailure<IllegalArgumentException>()
+        }
+
+
+    }
```
robstoll, line 76: Please remove this one again, suffices if we have it in ResultExpecationsSpec, here we only want samples and not regression tests. Also, the samples below already failed due to the bug, so we do not even need an additional regression here.

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/ResultExpectationsSpec.kt
```diff
@@ -54,73 +54,73 @@ abstract class ResultExpectationsSpec(
...
-                    }.toThrow<AssertionError> {
-                        messageToContain("value: $isNotSuccessDescr")
-                        if (hasExtraHint) messageToContain("$toEqualDescr: 1")
-                    }
-                }
-            }
-
-            //TODO 0.20.0 activate once we have the workaround for #1234 implemented
+//
+//    fun describeFun(vararg pairs: SpecPair<*>, body: Suite.() -> Unit) =
```
robstoll, line 58: I guess that was accidental, you should not comment out the tests

robstoll (changes requested):

robstoll:
@JordanllHarper make sure to rebase on main since you have conflicts

apis/fluent-en_GB/extensions/atrium-api-fluent-en_GB-kotlin_1_3/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/kotlin_1_3/samples/ResultExpectationSamples.kt
JordanllHarper: Do you want me to comment out all the tests to the bottom or just the test?

robstoll (commented):
btw. no need for a new PR, you can keep this one open. we are almost good 😊

JordanllHarper:
If it's okay can we continue on the new one. As you can see, I was having some issues with git last night 😅.

robstoll:
sure

PR #1430, opened 2023-05-08
simplify fix for unwrap Result bug

robstoll:


robstoll:
@JordanllHarper see how I simplified it. Note especially, that I use `it` inside of `Option.map`, you were still using `this.exceptionOrNull()`

JordanllHarper:
Ah I see with the fix now. My thoughts were a little misguided when I experimented with the idea of doing the exception or null check manually. Thank you for the learning experience!

PR #1436, opened 2023-05-13
Change report for not to equal null

JordanllHarper:

### DIFF
--- a/translations/atrium-translations-en_GB/src/commonMain/kotlin/ch/tutteli/atrium/translations/DescriptionAnyExpectation.kt
+++ b/translations/atrium-translations-en_GB/src/commonMain/kotlin/ch/tutteli/atrium/translations/DescriptionAnyExpectation.kt
@@ -29,5 +29,5 @@ enum class DescriptionAnyExpectation(override val value: String) : StringBasedTr
     BECAUSE("because"),
 
     //#330
-    NOT_TO_BE_NULL_BUT_TO_BE_THE_INSTANCE("not to equal: null but to be an instance of: ") ,
+    NOT_TO_BE_NULL_BUT_TO_BE_THE_INSTANCE("not to equal: null but to be an instance of") ,
 }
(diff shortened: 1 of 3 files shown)


robstoll (changes requested):

translations/atrium-translations-en_GB/src/commonMain/kotlin/ch/tutteli/atrium/translations/DescriptionAnyExpectation.kt
```diff
@@ -26,5 +26,8 @@ enum class DescriptionAnyExpectation(override val value: String) : StringBasedTr
     NOT_TO_EQUAL_ONE_IN("not to equal one in"),
 
     /** @since 0.18.0 */
-    BECAUSE("because")
+    BECAUSE("because"),
+
+    //#330
+    NOT_TO_BE_NULL_BUT_TO_BE_THE_INSTANCE("not to equal: null but to be an instance of: ") ,
```
robstoll, line 32:
```suggestion
    NOT_TO_BE_NULL_BUT_TO_BE_THE_INSTANCE("not to equal: null but to be an instance of") ,
```
Please take a look how the output looks like, with your current implementation there is a double `:` at the end

you need to: 
- add the translation enum also to de_CH (use `entsprich nicht: null sondern ist eine Instanz vom Typ`)
- adopt the tests
- also adopt notToEqual in case one passes `null` (also during runtime)
- run `./gradlew :readme-examples:build` => there should be changes in README.md due to the change in the output

JordanllHarper: Just in regards to the tests, just to clarify what you mean by adopt? Do you want me to write tests or are there some already?

robstoll: I guess you are not familiar with Github Actions. See below, the card with "Some checks were not successful". You can click on the one with a red cross :x: You will see that inter alia AnyExpectationsSpec failed (https://github.com/robstoll/atrium/actions/runs/5006785006/jobs/8972507160?pr=1436#step:4:435) because notToEqualNull changed. Try to fix this test and all others

JordanllHarper: Ah okay yeah I've never used github actions before so I'll try to get those passing.

robstoll:
closing this for now, re-open once you have new changes

PR #1453, opened 2023-05-30
Feat/1400 add not to be an instance of

AncutaIoan:


apis/fluent/atrium-api-fluent/src/commonMain/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyExpectations.kt
```diff
@@ -167,6 +167,14 @@ internal fun <TSub : Any> Expect<*>.toBeAnInstanceOf(kClass: KClass<TSub>): Subj
 inline fun <reified TSub : Any> Expect<*>.toBeAnInstanceOf(noinline assertionCreator: Expect<TSub>.() -> Unit): Expect<TSub> =
     toBeAnInstanceOf(TSub::class).transformAndAppend(assertionCreator)
 
+inline fun <reified TSub : Any> Expect<*>.notToBeAnInstanceOf(): Expect<Boolean> = notToBeAnInstanceOf(TSub::class).transform()
```
robstoll, line 170: IMO we should not transform to Boolean but keep the same subject as before. I would imagine that someone is using it like the following:
```
sealed interface A {
  val someInt: Int
}
class B: A { ... }
class C: A { ... }
class D: A { ... }

expect(someA).because("can be any of A but not D because ...") {
  notToBeAnInstanceOf<D>()
    .feature(A::someInt)  // extract feature of A, would not be possible if we transform to Boolean
    .toBeEqualTo(5)
}
```

<hr/>

please add a KDOC and link to your sample (see other functions in this file as guidance)

translations/atrium-translations-de_CH/src/commonMain/kotlin/ch/tutteli/atrium/translations/DescriptionAnyExpectation.kt
```diff
@@ -26,5 +26,8 @@ enum class DescriptionAnyExpectation(override val value: String) : StringBasedTr
     NOT_TO_EQUAL_ONE_IN("entspricht keinem in"),
 
     /** @since 0.18.0 */
-    BECAUSE("denn")
+    BECAUSE("denn"),
+
+    /** @since 1.0.0 */
```
robstoll, line 31:
```suggestion
    /** @since 1.1.0 */
```

translations/atrium-translations-en_GB/src/commonMain/kotlin/ch/tutteli/atrium/translations/DescriptionAnyExpectation.kt
```diff
@@ -26,5 +26,8 @@ enum class DescriptionAnyExpectation(override val value: String) : StringBasedTr
     NOT_TO_EQUAL_ONE_IN("not to equal one in"),
 
     /** @since 0.18.0 */
-    BECAUSE("because")
+    BECAUSE("because"),
+
+    /** @since 1.0.0 */
```
robstoll, line 31:
```suggestion
    /** @since 1.1.0 */
```

robstoll (changes requested):

robstoll:
@AncutaIoan thanks for the contribution, looks good, we just need one change and some small details

logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultAnyAssertions.kt
```diff
@@ -98,4 +98,11 @@ class DefaultAnyAssertions : AnyAssertions {
                 .build()
         ).build()
     }
+
+
+    override fun <T, TSub : Any> notToBeAnInstanceOf(container: AssertionContainer<T>, expected: KClass<TSub>): SubjectChangerBuilder.ExecutionStep<T, Boolean> =
+        container.changeSubject.reportBuilder()
+            .checkCast(expected)
+            .build()
```
robstoll, line 106: Since we don't need a transformation, it suffices to use an implementation similar to `toBe` above

logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/creating/transformers/SubjectChangerBuilder.kt
```diff
@@ -81,7 +81,16 @@ interface SubjectChangerBuilder {
...
+        /**
+         * Uses [DescriptionAnyExpectation.NOT_TO_BE_THE_INSTANCE_OF] as description of the change,
+         * the given [type] as representation and checked agaisnt the container [container]'s
+         * [AssertionContainer.maybeSubject] to the given type [TSub]
+         */
+        fun <TSub : Any> checkCast(type: KClass<TSub>): FailureHandlerStep<T, Boolean> =
+            withDescriptionAndRepresentation(DescriptionAnyExpectation.NOT_TO_BE_THE_INSTANCE_OF, type)
+                .withTransformation {
+                    Option.someIf(!type.isInstance(it)) { !type.isInstance(it) }
+                }
```
robstoll, line 93: remove again

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/AnyExpectationsSpec.kt
```diff
@@ -53,6 +53,8 @@ abstract class AnyExpectationsSpec(
     toBeAnInstanceOfSuperType: Feature1<out Any?, Expect<SuperType>.() -> Unit, SuperType>,
     toBeAnInstanceOfSubTypeFeature: Feature0<out Any?, SubType>,
     toBeAnInstanceOfSubType: Feature1<out Any?, Expect<SubType>.() -> Unit, SubType>,
+    notToBeInstanceOfFeature:Feature0<out Any?, Boolean>,
+    notToBeAnInstanceOf: Feature1<out Any?, Expect<SuperType>.() -> Unit, Boolean>,
```
robstoll, line 57: please also add a test-case

apis/fluent/atrium-api-fluent/src/commonMain/kotlin/ch/tutteli/atrium/api/fluent/en_GB/anyExpectations.kt
AncutaIoan: This is actually quite interesting, I just now realize it is possible to chain tests without regards to one another. Thanks for sharing this knowledge! But just one question : Won't this make the test coupled? If say, for instance, we sometime along the design realize that we have to change the implementation of class A (due to some random reason). I'll do the changes you requested, but I'm just curios of this aspect!

robstoll: not sure I get what you mean by coupled. The above example is of course fictional but if it is true that for some (business) case all implementations (maybe expect for D) need to return 5 for `someInt` then I don't see coupling here but merely testing the behaviour of the interface. But maybe you meant something else?

AncutaIoan: No, you explained it fine, was just a bit confuse about something

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/AnyExpectationsSpec.kt
AncutaIoan: Here I runned into a problem, I can't actually find a way to test this out. Could you maybe hint me out how I should go about adding a unifySignatures that would work for feature0/1 with same return type?

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/testUtils.kt
```diff
@@ -182,6 +181,20 @@ fun <T, R> unifySignatures(
...
 
+// TODO I cant make this work any way I try, it seems to me like I wont be able to unify like this
+//@JvmName("unifySignatures0Feature")
+//fun <T, R> unifySignatures(
+//    f0: Feature0<T, T>,
+//    f1: Feature1<T, Expect<R>.() -> Unit, T>
+//): List<Triple<String, Expect<T>.(Expect<R>.() -> Unit) -> Expect<T>, Boolean>> {
+//    val f0WithSubAssertion: Expect<T>.(Expect<R>.() -> Unit) -> Expect<T> =
+//        { f: Expect<R>.() -> Unit -> (f0.lambda)() }
+//    return listOf(
```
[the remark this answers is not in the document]
robstoll: *Edit* see comment to your question above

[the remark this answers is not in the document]
AncutaIoan: Don't think it would fix it, my problem begin with the signature of the method itself, the inside I got it in a good state. (And I think we should actually -> Expect<T> since we don't use transform/others as state in other comment

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/AnyExpectationsSpec.kt
robstoll: this function already exists actually and is already used by `toBeAnInstanceOfSubTypeFeature` and `toBeAnInstanceOfSubType` in AnyExpectationsSpec.

Following the things you should do:
- add notToBeInstanceOfFeature and notToBeAnInstanceOf to both includes of of the SubjectLessSpec
- add a `describeFun` (see line 683 toBeAnInstanceOfIntFeature) and add similar test cases

robstoll:
@AncutaIoan do you need help somehow?

robstoll:
@AncutaIoan thanks for the pre-work and your first contribution to Atrium :+1: 

I am going to take over and fix the rest.

PR #1502, opened 2023-08-06
suppress warning about upper bounds for TSub and Any

robstoll:


robstoll:
doesn't work out as we leave out T on purpose in API so that we can write `isA<TSub>` instead of `isA<T, TSub>`

PR #1512, opened 2023-08-19
Restrict E at OptionalAssertions.isPresent with Any

Nikos-Tsiougranas:
Restricts E at OptionalAssertions.isPresent with Any.

Also adapts the `toBePresent` in `apis/fluent/atrium-api-fluent/src/jvmMain/kotlin/ch/tutteli/atrium/api/fluent/en_GB/optionalExpectations.kt` based on the previous added restriction and deletes tests that do not comply with the new restrictions since the receiver was Optional<Any?>.

Also adapts the `toBePresent` in `apis/infix/atrium-api-infix/src/jvmMain/kotlin/ch/tutteli/atrium/api/infix/en_GB/optionalExpectations.kt` based on the previous added restriction and deletes tests that do not comply with the new restrictions since the receiver was Optional<Any?> as well.


robstoll (approved):
@Nikos-Tsiougranas thanks for your first contribution to Atrium 🙂👍

that was an easy one, isn't it. how about another one?

and don't forget to start Atrium if you like it

PR #1531, opened 2023-08-27
Add butAtMost sample test and link it with the corresponding function

Nikos-Tsiougranas:
Adds sample for the butAtMost function and links using @sample with the corresponding function.


robstoll:
@Nikos-Tsiougranas LGTM thanks 😊

robstoll (approved):

PR #1535, opened 2023-09-03
Feature/#330-Not to equal null report change

JordanllHarper:
Draft pr - think I've got the feature correct but getting an error regarding the DescriptionAnyExpectation field not being found


robstoll:
Take a look at your old PR: https://github.com/robstoll/atrium/pull/1436/files you also need to add it to de_CH. Also you still need to adapt the tests (as mentioned in the old PR)

JordanllHarper:
Oh sorry it's been a while and I completely forgot! Will get those amended.

JordanllHarper:
Okay managed to get tests passing locally.
Is the version number right + is the translation correct? Used the old pr for it.

robstoll:
as mentioned in the old PR:  
run ./gradlew :readme-examples:build => there should be changes in README.md due to the change in the output

but since the tests did not fail, something must be wrong

JordanllHarper:
I've changed the tests to use the descriptions and amended the readme.

Just to clarify by "I see you changed the translation, the one before was correct", you mean that the translation I've put in is all good?

README.md
```diff
@@ -1574,10 +1574,8 @@ expect(Paths.get("/root/.ssh/config")).toBeWritable()
 ```text
 I expected subject: /root/.ssh/config        (sun.nio.fs.UnixPath <1234789>)
 ◆ to be: writable
-    » failure at parent path: /root        (sun.nio.fs.UnixPath <1234789>)
-      » access was denied
-      » the owner is root, the group is root
-      » the permissions are u=rwx g= o=
+    » no file system entry exists at this location
```
robstoll, line 1577: please revert that change in the end (same same for other path related changes)

translations/atrium-translations-de_CH/src/commonMain/kotlin/ch/tutteli/atrium/translations/DescriptionAnyExpectation.kt
```diff
@@ -19,6 +19,9 @@ enum class DescriptionAnyExpectation(override val value: String) : StringBasedTr
     /** @since 0.18.0 */
     TO_BE_THE_INSTANCE("ist dieselbe Instanz wie"),
```
robstoll, line 21: since 1.1.0

robstoll (changes requested):
first you had:  
not to equal: null but to be an instance of

now you have:
not to equal: null but to be an instance of type

The first version is what we want in the end.

btw. the following point of the issue is still open:
change the implementation of notToEqual treat null specially so that expect(null as List<Int>?).notToBe(null) also results in the reporting above (without to be: empty of course)

but you can do this also in a subsequent PR

JordanllHarper:
So this won't build but I just wanted to push and see what you thought of these changes, in regards to adapting the not to be module. Is this on the right lines and if so, how do I go about fixing the compile errors?

README.md
```diff
@@ -398,7 +398,7 @@ expect {
 I expected subject: () -> kotlin.Nothing        (readme.examples.MostExamplesSpec$1$6$1 <1234789>)
 ◆ ▶ thrown exception when called: java.lang.IllegalArgumentException
     ◾ ▶ message: null
-        ◾ to be an instance of type: String (kotlin.String) -- Class: java.lang.String
+        ◾ not to equal: null but to be an instance of type: String (kotlin.String) -- Class: java.lang.String
```
robstoll, line 401: don't forget to run `./gradlew :readme-examples:build`  this should change if you adopt the translation

logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultAnyAssertions.kt
```diff
@@ -18,8 +17,17 @@ class DefaultAnyAssertions : AnyAssertions {
     override fun <T> toBe(container: AssertionContainer<T>, expected: T): Assertion =
         container.createDescriptiveAssertion(TO_EQUAL, expected) { it == expected }
 
-    override fun <T> notToBe(container: AssertionContainer<T>, expected: T): Assertion =
-        container.createDescriptiveAssertion(NOT_TO_EQUAL, expected) { it != expected }
+    override fun <T> notToBe(container: AssertionContainer<T?>, expected: T?): Assertion {
```
robstoll, line 20: There isn't a need to change `expected: T`  to `expected: T?` because `T` has not `Any`  as upper bound constraint, i.e. it is already possible to use e.g. `String?` as substitute for `T, thus one can already pass `null`  as expected since `null`  is a valid value of ` String?` 
```suggestion
    override fun <T> notToBe(container: AssertionContainer<T>, expected: T): Assertion {
```

logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultAnyAssertions.kt
```diff
@@ -18,8 +17,17 @@ class DefaultAnyAssertions : AnyAssertions {
...
+
+        if (container.maybeSubject.getOrElse { null } == null || expected == null) {
```
robstoll, line 22: We are only interested in the case where `expected` is ` null`  not where the subject is null. In the end we want more or less the same report for 
```
expect(null as Int?).notToEqualNull { ... }
``` 
and
``` 
expect(null as Int?).notToEqual(null)
```
(but not for `expect(null as Int?).notToEqual(2)` 
However, I just realised, that we would need to break notToBe and expect `subType: KClass<T>`  in addition in order that you can report the type. My bad, we won't do this at this point. so please revert all changed related to this requirement (all changes in this file basically)

robstoll (changes requested):

logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultAnyAssertions.kt
JordanllHarper: Okay no problem so we are just keeping the translation and initial `notToEqualNull` ?

robstoll: I think the code as it was after commit d02abb19c6a69e883fea17fd71536dac25096cd7 was everything we needed. I can also cherry-pick from there I totally get that it starts to get confusing now

JordanllHarper: Yeah that would be helpful thank you

robstoll:
Closing this in favour of #1572 (cherry-picked the relevant changes). 
@JordanllHarper thanks for your contribution :+1:  and looking forward to the next  :slightly_smiling_face:

PR #1559, opened 2023-09-29
Add samples for Iterable.toContain.exactly

Nikos-Tsiougranas:
Adds samples for Iterable.toContain.exactly

### DIFF
--- a/apis/fluent/atrium-api-fluent/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableLikeToContainCheckerSamples.kt
+++ b/apis/fluent/atrium-api-fluent/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableLikeToContainCheckerSamples.kt
@@ -5,6 +5,26 @@ import ch.tutteli.atrium.api.verbs.expect
 import kotlin.test.Test
 
 class IterableLikeToContainCheckerSamples {
+    @Test
+    fun exactly() {
+        expect(listOf("A", "B", "A")).toContain.inAnyOrder.exactly(2).entry {
+            toEqual("A")
+        }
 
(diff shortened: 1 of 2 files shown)


apis/fluent/atrium-api-fluent/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableLikeToContainCheckerSamples.kt
```diff
@@ -5,6 +5,26 @@ import ch.tutteli.atrium.api.verbs.expect
 import kotlin.test.Test
 
 class IterableLikeToContainCheckerSamples {
+    @Test
+    fun exactly() {
+        expect(listOf("A", "B", "A")).toContain.inAnyOrder.exactly(2).entry {
+            toBeEqualComparingTo("A")
```
robstoll, line 11: Please use `toEqual` instead of toBeEqualComparingTo, there is not need for a compareTo comparison in case of a String:
```suggestion
            toEqual("A")
```
Same same further below

robstoll (changes requested):

apis/fluent/atrium-api-fluent/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableLikeToContainCheckerSamples.kt
Nikos-Tsiougranas: Sure!

robstoll (approved):

robstoll:
@Nikos-Tsiougranas thanks for the contribution 🙏

PR #1560, opened 2023-10-05
introduce expectGrouped (ExpectGrouping) and group

robstoll:


vlsi:
@robstoll , I'm testing this, and it looks like there's no easy way to provide a custom subject description for `ExpectGrouping.expect(subject: Expect<R>): Expect<R>`.

I'm trying to wrap it like

```kotlin
fun <T> ExpectGrouping.expect(
    subject: T,
    @CallerArgumentExpression("subject") description: String = "",
): Expect<T> =
    atriumExpect(subject) // <-- How could I pass "description" to the subject?
```

robstoll:
You don't want to see `I expected subject` right? expect is just a delegate to manual feature in this context, so you could use the following instead: `_logic.manualFeature(description){ subject }.transform()`

PR #1575, opened 2023-10-15
add Map.toContain...entries infix samples

danherrera:
Add `MapLikeToContain...CreatorSamples.kt` entries infix samples.

Thanks for highlighting that these were missing and providing an example @robstoll! I should have highlighted that the infix samples were mismatched compared to the fluent module.


robstoll:
:+1: thanks for the supplement

PR #1616, opened 2023-11-26
add samples for Iterable.toContain.notOrAtMost

Zuretto:
### Outline
Added samples for _Iterable.toContain.notOrAtMost_.

### Issue
#1549 

### DIFF
--- a/apis/fluent/atrium-api-fluent/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableLikeToContainCheckerSamples.kt
+++ b/apis/fluent/atrium-api-fluent/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableLikeToContainCheckerSamples.kt
@@ -27,4 +27,28 @@ class IterableLikeToContainCheckerSamples {
             }
         }
     }
+
+    @Test
+    fun notOrAtMost() {
+        expect(listOf("A", "A", "B", "C")).toContain.inAnyOrder.notOrAtMost(2).entry {
+            toEqual("A")
+        }
+
+        expect(listOf(1, 2, 3)).toContain.inAnyOrder.notOrAtMost(2).entry {
+            toBeGreaterThan(1)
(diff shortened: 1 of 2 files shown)


apis/fluent/atrium-api-fluent/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableLikeToContainCheckerSamples.kt
```diff
@@ -27,4 +27,28 @@ class IterableLikeToContainCheckerSamples {
...
     }
+
+    @Test
+    fun notOrAtMost() {
+        expect(listOf("A", "A", "B", "C")).toContain.inAnyOrder.notOrAtMost(2).entry {
+            toEqual("A")
+        }
+
+        expect(listOf(1, 2, 3)).toContain.inAnyOrder.notOrAtMost(2).entry {
+            toBeGreaterThan(1)
```
robstoll, line 38: how about you change this to 4 and add a comment why this expectation holds:
```suggestion
            // none fulfils the expectation which is fine because we use notOrAtMost
            // use atMost if you want that at least one element matches
            toBeGreaterThan(4) 
```

apis/fluent/atrium-api-fluent/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableLikeToContainCheckerSamples.kt
```diff
@@ -27,4 +27,28 @@ class IterableLikeToContainCheckerSamples {
...
+        }
+
+        fails {
```
robstoll, line 41: would be nice if you explain why it fails (same same below)
```suggestion
        fails { // because "B" is 3 times in the List
```

robstoll (changes requested):

PR #1686, opened 2024-01-30
feat: add default variance to Expect<out T>, AssertionContainer<out T>, FeatureExpect<out T, out R>

vlsi:
It enables usages as follows:

```kotlin
    val Expect<Number>.numberOfDigits: FeatureExpect<Number, Int>
        get() = feature("Number of digits") {
            toString().length
        }

    expect(42).numberOfDigits.toEqual(2)
```

Another usage pattern:

```kotlin
interface Named {
    val name: String
}

data class Item(override val name: String, ...): Named
data class Person(override val name: String, ...): Named

val Expect<Named>.name: FeatureExpect<Named, String>
    get() = feature("name") { name }

...

expect(Person("Fred")).name.toEqual("Fred")
```

### DIFF
--- a/apis/fluent/atrium-api-fluent/src/jvmTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/BigDecimalExpectationsSpec.kt
+++ b/apis/fluent/atrium-api-fluent/src/jvmTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/BigDecimalExpectationsSpec.kt
@@ -14,7 +14,7 @@ class BigDecimalExpectationsSpec : Spek({
         @Suppress("DEPRECATION") fun1<BigDecimal?, BigDecimal?>(Expect<BigDecimal?>::toEqual).withNullableSuffix(),
         fun1<BigDecimal?, Nothing?>(Expect<BigDecimal?>::toEqual).withNullableSuffix(),
         Expect<Any>::toEqual,
-        @Suppress("DEPRECATION") (Expect<BigDecimal>::notToEqual.name to  Expect<BigDecimal>::notToEqual),
+        @Suppress("DEPRECATION") (Expect<BigDecimal>::notToEqual.name to { notToEqual(it) }),
         Expect<Any>::notToEqual,
         Expect<BigDecimal>::toEqualNumerically.name to Expect<BigDecimal>::toEqualNumerically,
         Expect<BigDecimal>::notToEqualNumerically.name to Expect<BigDecimal>::notToEqualNumerically,
--- a/misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/AnyExpectationsSpec.kt
+++ b/misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/AnyExpectationsSpec.kt
@@ -599,7 +599,7 @@ abstract class AnyExpectationsSpec(
 
 
     describeFun(notToEqualNullFeature, notToEqualNull) {
-        val notToEqualNullFunctions = unifySignatures(notToEqualNullFeature, notToEqualNull)
+        val notToEqualNullFunctions = unifySignatures<Int?, Int>(notToEqualNullFeature, notToEqualNull)
 
         context("subject is null") {
             notToEqualNullFunctions.forEach { (name, notToEqualNullFun, hasExtraHint) ->
--- a/apis/fluent/atrium-api-fluent/src/commonMain/kotlin/ch/tutteli/atrium/api/fluent/en_GB/arraySubjectChangers.kt
+++ b/apis/fluent/atrium-api-fluent/src/commonMain/kotlin/ch/tutteli/atrium/api/fluent/en_GB/arraySubjectChangers.kt
@@ -17,7 +17,7 @@ import kotlin.jvm.JvmName
  *
  * @since 0.9.0
  */
-fun <E> Expect<out Array<out E>>.asList(): Expect<List<E>> =
+fun <E> Expect<Array<out E>>.asList(): Expect<List<E>> =
     _logic.changeSubject.unreported { it.asList() }
 
 /**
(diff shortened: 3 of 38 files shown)


vlsi:
@robstoll , WDYT?

misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/AnyExpectationsSpec.kt
```diff
@@ -599,7 +599,7 @@ abstract class AnyExpectationsSpec(
 
 
     describeFun(notToEqualNullFeature, notToEqualNull) {
-        val notToEqualNullFunctions = unifySignatures(notToEqualNullFeature, notToEqualNull)
+        val notToEqualNullFunctions = unifySignatures<Int?, Int>(notToEqualNullFeature, notToEqualNull)
```
vlsi, line 602: I'm not sure why this does not work without explicit type parameters, however, `unifySignatures` looks more like a corner-case rather than real-life usage.

apis/fluent/atrium-api-fluent/src/jvmTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/BigDecimalExpectationsSpec.kt
```diff
@@ -14,7 +14,7 @@ class BigDecimalExpectationsSpec : Spek({
         @Suppress("DEPRECATION") fun1<BigDecimal?, BigDecimal?>(Expect<BigDecimal?>::toEqual).withNullableSuffix(),
         fun1<BigDecimal?, Nothing?>(Expect<BigDecimal?>::toEqual).withNullableSuffix(),
         Expect<Any>::toEqual,
-        @Suppress("DEPRECATION") (Expect<BigDecimal>::notToEqual.name to  Expect<BigDecimal>::notToEqual),
+        @Suppress("DEPRECATION") (Expect<BigDecimal>::notToEqual.name to { notToEqual(it) }),
```
vlsi, line 17: This change was needed to call the deprecated `Expect<BigDecimal?>.notToEqual(BigDecimal?)`.
I do not know if there's a way to disambiguate `Expect<BigDecimal>::notToEqual` between `notToEqual(BigDecimal?) `and `notToEqual(Nothing?)`

I think having a lambda here should be good enough.

robstoll:
We had `Expect<out T>` in the beginning and then moved to an invariant Expect because it allows that the compiler can already warn you about wrong usage via a compile error. For instance, we don't want that someone is able to write:
```
expect(1.0).toEqual(1)
```
Because it is already clear at compile-time that this will never hold.


Your example works if you a type parameter with an upper bound instead of `Number`
```
val <T: Number> Expect<T>.numberOfDigits: FeatureExpect<T, Int>
        get() = feature("Number of digits") {
            toString().length
        }

    expect(42).numberOfDigits.toEqual(2)
```

or in case you don't care about the initial type of the subject (in the sense you don't need it in the signature), then you can also use out, such as in:
```
fun <E> Expect<out Array<out E>>.asList(): Expect<List<E>>
```

I hope that helps, I am going to close this PR, please comment in case I missed something and there are other use cases we would like to support which cannot be supported right now

vlsi:
The relevant Kotlin issue is https://youtrack.jetbrains.com/issue/KT-13198/Expose-OnlyInputTypes-annotation-to-allow-proper-checks-of-parameters-of-generic-methods-same-as-those-used-Collection.contains

robstoll:
Thanks, I know this one but this alone doesn't help unfortunately. The problem is that we don't have methods but extension methods. If we define `Expect<out T>` then an extension method `fun Expect<Number>` can always be upcasted to this type. In other words if you we define `expect(1)` then we have a `Expect<Int>` but this is also a subtype of `Expect<Number>`. So if you do `toEqual(1.0)` aftewards, then it is OK to infer that expect is `Expect<Number>` and 1.0 is a `Number` to. All good, the input type was `Number` and the output type is `Number` to. 

What we would need is somethigng like https://youtrack.jetbrains.com/issue/KT-24799/restrict-use-site-variance-or-introduce-lower-bounds

vlsi:
@robstoll , thanks for the explanation.

What do you think of the following then?

1. Add the notice on variance to the `Expect` type interface. Suggest 
2. Add a sample of implementing at custom feature right into `Expect` documentation
3. Use `val <T: Person> Expect<T>` instead of `val Expect<Person>` in [the readme](https://github.com/robstoll/atrium?tab=readme-ov-file#ex-own-compose-4)
4. Consider adding a compiler plugin that would verify the use of Atrium APIs. For instance, if a user declares `val Expect<Person>` with `Person` being a non-final class, the plugin could emit a warning "consider using `val <T: Person> Expect<T>` instead"

robstoll:
1. :+1: 
2. :+1: 
3. this one would show an error in intellij as Person is final, so I don't think it would be good to change. Just a bit further above in the [readme](https://github.com/robstoll/atrium?tab=readme-ov-file#compose-expectation-functions) there is "Why is a type parameter used in the above examples?" explaining it. Maybe we should make this more prominent? And I think it makes sense to add something here as well along the line of "Why is no type parameter used in this example?"
4. :+1: though I am not sure if it wouldn't be better if we add it to an existing linter project such as Detekt. WDYT?

vlsi:
> this one would show an error in intellij as Person is final, so I don't think it would be good to change.

It would be great to have a copy-paste-ready source for "defining your own extension".

For instance:

```kotlin
val <T: SQLException> Expect<T>.errorCode: FeatureExpect<T, Int>
    get() = feature("errorCode") { errorCode }
```

> if we add it to an existing linter project such as Detekt

Yet another option could be adding OpenRewrite rule to rewrite `val Expect<SQLException>` into `val <T: SQLException> Expect<T>`.

robstoll:
I agree and I don't mind to add more or better examples to the README. Maybe we should remove examples which are not ideal as people tend to copy without reading :thinking: For instance, the first example in https://github.com/robstoll/atrium?tab=readme-ov-file#compose-expectation-functions is not ideal, I'll remove it and keep only the second

robstoll:
tried to improve it (https://github.com/robstoll/atrium/pull/1697), more improvements are of course welcome

Issue #1569, opened 2023-10-15
samples for Iterable feature extractors


robstoll:
*Platform* (all, jvm, js): all
*Extension* (none, kotlin 1.3): none

## Code related feature
*api-fluent*
*api-fluent*
- [ ] add a `minFeature` method in IterableFeatureExtractorSamples for expect(iterable).min()
- [ ] add a `min` method IterableFeatureExtractorSamples for expect(iterable).min { } (i.e. the overload with the assertionCreator-lambda)
- [ ] add a `maxFeature` method in IterableFeatureExtractorSamples for expect(iterable).max()
- [ ] add a `max` method IterableFeatureExtractorSamples for expect(iterable).max { } (i.e. the overload with the assertionCreator-lambda)
- [ ] link in the KDoc of the corresponding function in iterableLikeToContain...Creators.kt to the samples via @sample (see charSequenceToContainCreators.kt)

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime, we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/main/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

constantine-zubrilin:
Can I take it?

PR #1708, opened 2024-02-18
Add samples for Iterable feature extractors

constantine-zubrilin:
Issue: https://github.com/robstoll/atrium/issues/1569


robstoll:
@constantine-zubrilin good catch, I suggest we do this in another PR. Also in the same go you could move the `last` functions from List to Iterable (my fault, figured the `last` fun is not defined for Iterable in Kotlin)

robstoll (changes requested):
would be nice if you could add comments as you did for `last`

constantine-zubrilin:
> @constantine-zubrilin good catch, I suggest we do this in another PR. Also in the same go you could move the `last` functions from List to Iterable (my fault, figured the `last` fun is not defined for Iterable in Kotlin)

I've started it here: https://github.com/robstoll/atrium/pull/1710

Issue #1718, opened 2024-03-06
Hyperlink for atrium-fluent in Examples section of README routes to non-existant file


lkorsman:
In the project's overall README, clicking on "atrium-fluent" hyperlink in the first sentence of the Examples section routes to a non-existent build.gradle file. I assume the link should route to the build.gradle.kts (https://github.com/robstoll/atrium/blob/main/bundles/fluent/atrium-fluent/build.gradle.kts).

![Screen Shot 2024-03-06 at 10 32 09 AM](https://github.com/robstoll/atrium/assets/20802668/76b44a5a-e26f-4751-8ab8-8c95a5ea6340)

lkorsman:
I am working on this

PR #1719, opened 2024-03-06
Update atrium-fluent hyperlink in README

lkorsman:
Fixes [#1718](https://github.com/robstoll/atrium/issues/1718)

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ Are you using a different runner? A PR would be appreciated 😊.
 
 # Examples
 We are using the API provided by the bundle module 
-[atrium-fluent](https://github.com/robstoll/atrium/blob/main/bundles/fluent/atrium-fluent/build.gradle.kts)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -172,7 +172,7 @@ Are you using a different runner? A PR would be appreciated 😊.
 
 # Examples
 We are using the API provided by the bundle module 
-[atrium-fluent](https://github.com/robstoll/atrium/tree/main/bundles/fluent/atrium-fluent/build.gradle)
+[atrium-fluent](https://github.com/robstoll/atrium/blob/main/bundles/fluent/atrium-fluent/build.gradle.kts)
```
robstoll, line 175: can you keep tree please, although blob makes more sense in this case. we have a rewrite in place which only works with tree currebtly

robstoll:
@lkorsman thanks for the fix 🙂👍

PR #1720, opened 2024-03-06
Move listFeature to iterableLike

constantine-zubrylin:

### DIFF
--- a/apis/fluent/atrium-api-fluent/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableFeatureExtractorSamples.kt
+++ b/apis/fluent/atrium-api-fluent/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/samples/IterableFeatureExtractorSamples.kt
@@ -91,7 +91,7 @@ class IterableFeatureExtractorSamples {
     fun lastFeature() {
         val iterable = sequenceOf(1, 2, 3).asIterable()
 
-        expect(list).last.toEqual(3) // subject is 3
+        expect(iterable).last.toEqual(3) // subject is 3
 
         expect(iterable).last // Subject is 3
             .toBeGreaterThan(2) // subject is still 3


robstoll (commented):
did not take a deep look as it is a draft. let me know in case I shall already review

apis/fluent/atrium-api-fluent/src/commonMain/kotlin/ch/tutteli/atrium/api/fluent/en_GB/iterableFeatureExtractors.kt
```diff
@@ -59,3 +61,29 @@ fun <E : Comparable<E>, T : Iterable<E>> Expect<T>.max(): Expect<E> =
...
+/**
+ * Expects that Iterable is not empty and returns an [Expect] for the last element in it.
+ *
+ * @return The newly created [Expect] for last element.
+ *
+ * @sample ch.tutteli.atrium.api.fluent.en_GB.samples.ListFeatureExtractorSamples.lastFeature
+ *
+ * @since 1.2.0
+ */
+val <E, T: Iterable<E>> Expect<T>.last: Expect<E?>
```
robstoll, line 74: no idea why you started adding `?` here and there but the implementation you had in List was just fine :)
```suggestion
val <E, T: Iterable<E>> Expect<T>.last: Expect<E>
```

logic/atrium-logic/src/generated/commonMain/ch/tutteli/atrium/logic/iterableLike.kt
```diff
@@ -45,6 +45,9 @@ fun <T : IterableLike, E : Comparable<E>> AssertionContainer<T>.max(converter: (
 
 fun <T : IterableLike, E> AssertionContainer<T>.containsNoDuplicates(converter: (T) -> Iterable<E>): Assertion = impl.containsNoDuplicates(this, converter)
 
+fun <T : IterableLike, E : Comparable<E>> AssertionContainer<T>.last(converter: (T) -> Iterable<E>): FeatureExtractorBuilder.ExecutionStep<T, E> = impl.last(this, converter)
```
constantine-zubrylin, line 48: I could not manage to get rid of `E : Comparable<E>`. Need help with it.

robstoll: hm. `last` should not require comparable, what error do you get when you just use `E`

constantine-zubrylin: Screenshot 2024-03-10 at 23 21 33
Not sure if it's the best format to show an error :)

I've pushed a commit so you can have a look.

logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultIterableLikeAssertions.kt
```diff
@@ -250,4 +249,31 @@ class DefaultIterableLikeAssertions : IterableLikeAssertions {
...
+    ): FeatureExtractorBuilder.ExecutionStep<T, E> =
+        container.extractFeature
+            .methodCall(method)
+            .withRepresentationForFailure(NO_ELEMENTS)
+            .withFeatureExtraction {
+                val iterable = converter(it)
+                Option.someIf(iterable.iterator().hasNext()) {
+                    iterable.collect() ?: throw IllegalStateException(
+                        "Iterable does not hasNext() even though checked before! Concurrent access?"
+                    )
```
robstoll, line 273: this is the problem, the last element could also be `null` we don't restrict it. Hence you need to change this to:
```suggestion
                    iterable.collect()
```

constantine-zubrylin: Oh, such a silly mistake. Thank you! I've pushed all the changes.

robstoll:
@constantine-zubrylin thanks :+1:

PR #1758, opened 2024-05-09
use bc-validator, apiDump based on tag v1.2.0

robstoll:
we are producing two dumps, once for main and once when using Kotlin 1.9 or newer for the forward-compatibility tests. Kotlin outputs `entries` for Enums since Kotlin 1.9

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -83,7 +83,7 @@ jobs:
       KOTLIN_VERSION: ${{ matrix.kotlin_version }}
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         kotlin_version: [ '1.5', '1.6', '1.7', '1.8', '1.9' ]
     steps:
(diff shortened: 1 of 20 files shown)


vlsi:
Is there a need for two dumps?
What do you think of validating backward compatibility only when building with Kotlin 1.9+?

In any case, we should build release binaries with the latest compiler versions, so I do not see the value in producing and storing two api dumps.

robstoll:
Yes, as outlined in the description. Maybe some more background. We use the Kotlin 1.9.24 compiler but produce Kotlin 1.4 compatible code. This way a project using kotlin 1.4 can still use Atrium -- as long as there is not a good reason to drop an older kotlin version I try to stick to it. 
With the forward compatibility tests we check that it would also work if we used a newer api-/languageVersion 
than 1.4 The reason why we do this is to lower the pain to drop support for older kotlin versions. So far I always found regressions when updating api-/languageVersion and had to search for workarounds.

Dumping two apis was the simplest solution I could think of. Dumping only 1.9 would not be enough as it then would always fail with api-/languageVersion < 1.9 and vice versa.

Does that make sense to you?

vlsi:
Ah, you mean creating a dump per apiTarget, not a dump depending on which compiler to use. Now I understand what you mean.

PR #1822, opened 2024-08-25
testFactory as preparation to get rid of spek

robstoll:


robstoll:
we would need to establish js coverage to cover turnTestNodesIntoExpectGrouping and co which are currently only used in JS

PR #1886, opened 2024-12-27
#1884 take surrogate code points into account in RegexSearcher

### LINKED ISSUE #1884: toContain.exactly.matchFor doesn't work if BnMS Node is used in Pattern
### Affected Version

1.2.0

### API

fluent, infix

### Platform

jvm

### Kotlin Version

2.0.21

### How to reproduce the problem?

```kotlin
expect("🚩️ size        : 0").toContain.exactly(1).regex("🚩️ s")
```

### Describe the bug

fails with (from proof branch):
``` 
I expected subject : "🚩️ size        : 0"
🚩️ to contain : 
   🚩️ string matching regex : 🚩️ s        (kotlin.text.Regex <162083492>)
      🚩️ ▶ number of matches : 2
           🚫️ is exactly     : 1
```            

This is an edge case, it works if the pattern does not resolve to BnMS Node. For instance, the following works as expected:
```kotlin
expect("🚩️ size        : 0").toContain.exactly(1).regex("🚩️")
```

### Expected behaviour

should also find just one occurrence

robstoll:
because if a Pattern (in java) uses BnMS then chars with two code points (high and low surrogate) match even if the start index points to the lower surrogate.

fixes #1884 

### DIFF
--- a/logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/creating/charsequence/contains/searchers/impl/RegexSearcher.kt
+++ b/logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/creating/charsequence/contains/searchers/impl/RegexSearcher.kt
@@ -11,10 +11,14 @@ import ch.tutteli.atrium.logic.creating.charsequence.contains.searchbehaviours.N
 class RegexSearcher : Searcher<NoOpSearchBehaviour, Regex> {
     override fun search(searchIn: CharSequence, searchFor: Regex): Int {
         var counter = 0
-        var matchResult = searchFor.find(searchIn)
+        val searchInString = searchIn.toString()
+        var matchResult = searchFor.find(searchInString)
         while (matchResult != null) {
-            matchResult = searchFor.find(searchIn, matchResult.range.first + 1)
             ++counter
+            val startIndex = matchResult.range.first.let { startIndex ->
+                startIndex + if (searchInString[startIndex].isHighSurrogate()) 2 else 1
+            }
+            matchResult = searchFor.find(searchIn, startIndex)
         }
         return counter
     }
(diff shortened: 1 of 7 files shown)


logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/creating/charsequence/contains/searchers/impl/RegexSearcher.kt
```diff
@@ -11,10 +11,14 @@ import ch.tutteli.atrium.logic.creating.charsequence.contains.searchbehaviours.N
 class RegexSearcher : Searcher<NoOpSearchBehaviour, Regex> {
     override fun search(searchIn: CharSequence, searchFor: Regex): Int {
         var counter = 0
-        var matchResult = searchFor.find(searchIn)
+        val searchInString = searchIn.toString()
```
vlsi, line 14: Just wondering: does `searchIn.findAll(searchFor): Sequence<MatchResult>` work?
`MatchResult.next(): MatchResult?` work?

It looks a bit weird to hand-craft it

robstoll: findAll would be a disjoint search, i.e. search `aa` in `aaaa` would find 2 occurrences. we use a non-disjoint search i.e. it finds it 3 times. Don't ask me why I went with a non-disjoint search tough. Looking at it now it seems a bit weird but back in 2017 I had a use case where it was relevant and it seemed natural to implement it that way. I guess it was an edge case as I cannot remember it. Most times it probably doesn't matter, so... I would keep it that way

vlsi: MatchResult.next() and friends go for disjoint matches which is what I would have expected if I used `toContain o exactly 3 regex`.

If you implement overlapping occurrences, then the hand-rolled impl is ok. I was just curios, there were not comments, so I asked.

robstoll: > disjoint matches which is what I would have expected if I used toContain o exactly 3 regex.

I get that as I had the same assumption some time ago. It's documented in the KDoc (also for `toContain o exactly 3 regex`) but it's not intuitive. I guess I will change the behaviour with 2.0.0

PR #1921, opened 2025-03-19
Migrate js-stubs from kotlin-js to kotlin-multiplatform with target js

### LINKED ISSUE #1887: migrate js-stubs from kotlin-js to mpp with target js

## Code related feature
When executing `./gradlew :js-stubs:build` you can see the following warning in the output:
```
> Configure project :js-stubs
w: 'kotlin-js' Gradle plugin is deprecated and will be removed in the future. 
Please use 'kotlin("multiplatform")' plugin with a 'js()' target instead. See the migration guide: https://kotl.in/t6m3vu
```

Migrate misc/js-stubs/build.gradle.kts accordingly

## Your first contribution?
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/main/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  

ragul-engg:
resolves #1887 


robstoll (approved):

robstoll:
@ragul-engg thanks for the changes. I think js-stubs was the only project using build-logic.kotlin-js. Could you please check if we can delete gradle/build-logic/dev/src/main/kotlin/build-logic.kotlin-js.gradle.kts

ragul-engg:
Yeah, will look into that. BTW in console there is a different warning being produced do you have any idea about this ?

```
> Task :js-stubs:compileKotlinJs
w: Language version 1.4 is deprecated and its support will be removed in a future version of Kotlin
```

ragul-engg:
Done as you have pointed out the `gradle/build-logic/dev/src/main/kotlin/build-logic.kotlin-js.gradle.kts` file is not used, have removed them

robstoll:
> Yeah, will look into that. BTW in console there is a different warning being produced do you have any idea about this ?
> 
> ```
> > Task :js-stubs:compileKotlinJs
> w: Language version 1.4 is deprecated and its support will be removed in a future version of Kotlin
> ```

The warning is legit as we produce kotlin 1.4 compatible code. However, normally we suppress it. I guess some setup we had in build-logic.kotlin-js is missing now.

Thinking about it a bit more, we should not delete build-logic.kotlin-js 🙈
Instead you should use kotlin-js in js-stubs again and make the adjustments in kotlin-js, i.e base it on the multiplatform-convention.

ragul-engg:
I did base the build-logic.kotlin-js on multiplatform-conventions. I am still seeing the same warning. I don't think we have anything specific to suppress those compileKotlinJs warnings in kotlin-js previously. Dropping the file for reference, I just replaced it to refer the mp-conventions

```
plugins {
    id("build-logic.kotlin-multiplatform-conventions") // changed here
}

kotlin {
    js(IR) {
        nodejs {
        }
    }
}
```

ragul-engg:
@robstoll I found it the kotlin-js build logic was lacking the kotlin-conventions, adding that suppressed the warning.

ragul-engg:
looks like its just based on just gradle-conventions. should i change that to kotlin-conventions as it imports gradle-conventions anyway. I think it should be fine, what do you think?

robstoll:
I think I missed to adjust it when I introduce kotlin-convention. Please change it, we will get a CI error if our assumption is wrong

robstoll (approved):
Looks good 🙂👍

robstoll:
Thanks for fixing more than required 🙂

Issue #1934, opened 2025-03-20
fix update to junit 5.12.1 in maven sample project


robstoll:
## Code related feature
See the PR https://github.com/robstoll/atrium/pull/1930 and its build https://github.com/robstoll/atrium/actions/runs/13973303812/job/39120511357?pr=1930

For some reason the update for maven fails (note that the same update in the gradle build for the junit5 project succeeds). 

Analyse why and fix it. Note that you need to change into the sample project directory to find it out, i.e.

```bash
cd ./samples/jvm/maven
./mvnw -B clean verify
```

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime, we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/main/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

robstoll:
@henriquedefreitas can you please leave a comment so that I can assign the issue to you. Thanks

PR #1936, opened 2025-03-30
#1934 fix updating junit version to 5.12.1

henriquedefreitas:
After some research, it seems that including the junit-jupiter-engine dependency inside the plugin tag is not necessary as junit-jupiter inside <dependencies> tag is a meta dependency that already includes the engine artifact, and it was probably causing some kind of conflict. The tests run fine without it.


robstoll (approved):

robstoll:
@henriquedefreitas thanks for your first contribution :+1:

PR #1941, opened 2025-04-05
Add iterable like to contain in any order only creator samples to api infix

Onaizit3:

### DIFF
--- a/apis/infix/atrium-api-infix/src/commonTest/kotlin/ch/tutteli/atrium/api/infix/en_GB/samples/IterableLikeToContainInAnyOrderOnlyCreatorSamples.kt
+++ b/apis/infix/atrium-api-infix/src/commonTest/kotlin/ch/tutteli/atrium/api/infix/en_GB/samples/IterableLikeToContainInAnyOrderOnlyCreatorSamples.kt
@@ -1,0 +1,138 @@
+package ch.tutteli.atrium.api.infix.en_GB.samples
+
+import ch.tutteli.atrium.api.verbs.expect
+import kotlin.test.Test
+import ch.tutteli.atrium.api.infix.en_GB.inAny
+import ch.tutteli.atrium.api.infix.en_GB.o
+import ch.tutteli.atrium.api.infix.en_GB.*
+
+class IterableLikeToContainInAnyOrderOnlyCreatorSamples {
+    @Test
+    fun value() {
+        expect(listOf("A")) toContain o inAny order but only value("A")
+
+        fails { // because the List does not contain expected value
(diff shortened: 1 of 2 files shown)


apis/infix/atrium-api-infix/src/commonTest/kotlin/ch/tutteli/atrium/api/infix/en_GB/samples/IterableLikeToContainInAnyOrderOnlyCreatorSamples.kt
```diff
@@ -0,0 +1,136 @@
+package ch.tutteli.atrium.api.infix.en_GB.samples
+
+import ch.tutteli.atrium.api.verbs.expect
+import kotlin.test.Test
+import ch.tutteli.atrium.api.infix.en_GB.*
+
+class IterableLikeToContainInAnyOrderOnlyCreatorSamples {
+    @Test
+    fun value() {
+        expect(listOf("A")) toContain o inAny order but only value("A")
```
robstoll, line 10: almost, you can also remove the last parentheses:
```suggestion
        expect(listOf("A")) toContain o inAny order but only value "A"
```
Same same below for other `value` cases, you need the parentheses for `values` though

robstoll (changes requested):
See review comment and note that you missed the second point written in the description of the issue:
> link to the samples in iterableLikeToContainInAnyOrderOnlyCreators.kt in api-infix

I.e. you need to add `@sample` to the KDoc, see other files in the api-infix

apis/infix/atrium-api-infix/src/commonTest/kotlin/ch/tutteli/atrium/api/infix/en_GB/samples/IterableLikeToContainInAnyOrderOnlyCreatorSamples.kt
Onaizit3: ok Rob, I made the changes

robstoll (approved):

robstoll:
@Onaizit3 thanks for your first contribution to Atrium :slightly_smiling_face: :+1: 
How about another one?

Onaizit3:
@robstoll,  I am ready for more contributions

robstoll:
@Onaizit3 nice to hear, there are more issues where we could use your help:
https://github.com/robstoll/atrium/issues?q=is%3Aissue%20state%3Aopen%20no%3Aassignee%20label%3A%22help%20wanted%22

Let me know in case you shouldn't find a suitable one, there are other things which could be done in addition.

Issue #1935, opened 2025-03-20
fix update to Kotlin 2.1.20 in the multiplatform and js sample projects


robstoll:
## Code related feature

Following the things you need to do:

- [ ] analyse error in the PR https://github.com/robstoll/atrium/pull/1933 
- [ ] checkout the corresponding branch and fix the error
- [ ] same same for https://github.com/robstoll/atrium/pull/1938 (should now be an easy one)

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime, we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/main/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

amasotti:
Hi @robstoll 

I discovered today this nice project and thought to take one open issue and dive into the source code.
The fix for the two errors should be relatively easy, if I see correctly. The only thing the pipeline was complaining about was that the `yarn.lock` file was not updated and commited (required if we update the kotlin version).

Hope it helps

Issue #1456, opened 2023-06-04
migrate check-dexer.gradle from groovy to kotlin i.e. to check-dexer.gradle.kts


robstoll:
## Code related feature
We check that Atrium works with android sdk api level 26 during CI by running d8 (dexer) --> see .github/workflows/build-ubuntu.yml => step `download Android SDK's android.jar (or use cached file)`

The configuration for this is done in rootProjectFolder/gradle/scripts/check-dexer.gradle

Note that we apply this script only if CI=true (see rootProjectFolder/build.gradle.kts). I recommend you comment out the `if` (~Line 560) during development so that intellij correctly highlights the syntax etc.

In order that you can run `./gradlew checkDexer` you need to define ATRIUM_ANDROID_JAR which needs to point to the android.jar of the SDK. See .github/build-windows.yml to see how you can download the sdk.

Note that we are not compatible with API level 25. So it is best to make a negative test, i.e. check if `./gradlew checkDexer` still fails if you point it to the android-25.jar

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
  This way we get the chance to revise the description in case things have changed in the meantime, we might give you additional hints and we can assign the task to you, so that others do not start as well.
- See [Your first code contribution](https://github.com/robstoll/atrium/blob/main/.github/CONTRIBUTING.md#your-first-code-contribution) for guidelines.  
- Do not hesitate to ask questions here or to contact us via [Atrium's slack channel](https://kotlinlang.slack.com/team/U3DE1TXKP) if you need help
  ([Invite yourself](https://slack.kotlinlang.org/) in case you do not have an account yet).

RealOkabe:
Sorry I went on a vacation and I wasn't able to work on this. I am back now. Starting work on this again.

RealOkabe:
I give up. Just setting up the environment is a pain and no one wants to help.

robstoll:
@RealOkabe I can still try to help. Last thing you mentioned, when we tried to tackle your Windows issues was that you will see if it works out with a linux distro. As mentioned when you sought help via slack, you can ignore the failing PathExpectation tests and focus on rewriting groovy to kotlin. We will see in CI if you broke something by accident. In any case, I am here to help in case you still want to give it try. If not, then please un-assign yourself from the issue

amasotti:
@robstoll I had a look at the groovy script. My first impressions is however that we cannot directly translate groovy into a kotlin DSL script (`.kts` file). 
[...]

PR #1948, opened 2025-04-12
chore(#1935): update to kotlin 2.1.20

amasotti:
Should solve #1935 

There was just one step missing (not automatically executed by renovatebot): update the `yarn.lock` file. 
Looks almost too easy to be true, but it was enough to run `kotlinUpgradeYarnLock` via gradle to fix the issue.

### DIFF
--- a/samples/js/kotlin-test/build.gradle.kts
+++ b/samples/js/kotlin-test/build.gradle.kts
@@ -22,7 +22,7 @@ dependencies {
 }
 
 plugins {
-    kotlin("js") version "2.0.20"
+    kotlin("js") version "2.1.20"
 }
 
 group = "org.atriumlib.samples"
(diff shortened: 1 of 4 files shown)


robstoll (approved):

robstoll:
@amasotti thanks for your first contribution to Atrium :slightly_smiling_face: :+1: 

Yeah... it's a pity that renovate is not doing it automatically, we basically run into this issue whenever kotlin changes some JS dependencies. There is another good first issue which is a bit more involved in case you are interested.

amasotti:
@robstoll Sure, I can have a look. 
I've worked much more with Kotlin for Backend than with kotlin-js, but I can give it a try :)

robstoll:
I was thinking about this one https://github.com/robstoll/atrium/issues/1456 which is currently the last remaining good first issue. Of course, you are also more than welcome to tackle another issue where we can use your help:
https://github.com/robstoll/atrium/issues?q=is%3Aissue%20state%3Aopen%20no%3Aassignee%20label%3A%22help%20wanted%22

PR #1949, opened 2025-04-14
refactor: migrate check dexer to kotlin

amasotti:
Solves #1456 

Migrated the last left groovy gradle script to Kotlin. In doing so I've also tried to introduce a more modular approach to the Atrium Android dexer compatibility check, by splitting the previously monolithic split in some file for better maintainability and readability.

**Key Changes:**

* **Central Plugin File** (`gradle/build-logic/check-dexer/src/main/kotlin/build-logic.check-dexer.gradle.kts`):
    * Defines the `AtriumDexerExtension` to configure the list of subprojects to be checked.
    * Registers the main `dexerPreCheck` task to validate the `ATRIUM_ANDROID_JAR` environment variable.
    * Resolves the list of `dexerProjects` based on the extension configuration.
    * Iterates through the configured projects and applies the dexer configuration using the `configureDexerForProject` utility function.
    * Registers the `preCheckDexerProjects` task to ensure the prerequisite check runs before any per-project dexer tasks.
    * Registers the aggregating `checkDexer` task that depends on the `checkDexer` task of each configured subproject.


gradle/build-logic/check-dexer/build.gradle.kts
```diff
@@ -0,0 +1,4 @@
+plugins {
+    base
```
robstoll, line 2: as far as I can tell you don't use anything of the base plugin, right? If so, I would remove it

gradle/build-logic/check-dexer/src/main/kotlin/DexerUtils.kt
```diff
@@ -0,0 +1,128 @@
...
+        mavenCentral()
+    }
+
+    // Setup r8 configuration
+    if (project.configurations.findByName("r8") == null) {
+        project.configurations.create("r8")
+        project.dependencies.add("r8", "com.android.tools:r8:8.7.18")
+    }
+
+    // Setup d8 configuration
```
robstoll, line 35: - please remove the comment, your method name already tells us the same
- same same below for registerCheckDexerTask
- for consistency reasons it would maybe make sense to create a configureR8Configuration as well? I would at least remove the comment as well, doesn't add

gradle/build-logic/check-dexer/build.gradle.kts
```diff
@@ -0,0 +1,4 @@
...
+    id("build-logic.kotlin-dsl-gradle-plugin")
+}
```
robstoll, line 4:
```suggestion
}

dependencies {
    api(projects.basics) 
}
```
basics adds the Project.prefixedProject extension

gradle/build-logic/check-dexer/src/main/kotlin/DexerUtils.kt
```diff
@@ -0,0 +1,128 @@
...
+    // Register the checkDexer task
+    registerCheckDexerTask(project, dexerPreCheckTask)
+}
+
+/**
+ * Configures the D8 configuration with appropriate attributes.
+ */
+private fun configureD8Configuration(project: Project) {
+    if (project.configurations.findByName("d8") == null) {
+        val d8Configuration = project.configurations.create("d8").apply {
```
robstoll, line 47: I think you could write it as follows (same for next apply) or is there a difference?
```suggestion
        val d8Configuration = project.configurations.create("d8") {
```

gradle/build-logic/check-dexer/src/main/kotlin/DexerUtils.kt
```diff
@@ -0,0 +1,128 @@
...
+    // Setup d8 configuration
+    configureD8Configuration(project)
+
+    // Register the checkDexer task
+    registerCheckDexerTask(project, dexerPreCheckTask)
+}
+
+/**
+ * Configures the D8 configuration with appropriate attributes.
+ */
```
robstoll, line 44: I would also remove this comment

gradle/build-logic/check-dexer/src/main/kotlin/DexerUtils.kt
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Registers and configures the checkDexer task that compiles JVM bytecode
+ * into DEX bytecode using the D8 compiler.
+ */
+private fun registerCheckDexerTask(project: Project, dexerPreCheckTask: Provider<Task>) {
+    project.tasks.register<JavaExec>("checkDexer") {
+        description = "Compiles android class files into dex bytecode"
+        inputs.property("ATRIUM_ANDROID_JAR", System.getenv("ATRIUM_ANDROID_JAR"))
+
+        val outputPath = "${project.layout.buildDirectory.get()}/d8"
```
robstoll, line 78: I think more gradle idiomatic would be:
```suggestion
        val outputPath = "${project.layout.buildDirectory.dir("d8")
```
But then you need to use a lazy approach for args further below. Up-to-you if you want to go down that path (see also comment regarding `get`

gradle/build-logic/check-dexer/src/main/kotlin/DexerUtils.kt
```diff
@@ -0,0 +1,128 @@
...
+        outputs.dir(outputPath)
+
+        // Set up dependencies
+        val jvmJar = project.tasks.named<org.gradle.jvm.tasks.Jar>("jvmJar").get()
```
robstoll, line 82: I don't mind if you keep it that way. The "new" way of configuring things in gradle would be to do everything lazily, i.e. don't use `get` and instead `map` and `flatMap` on a `Provider`. as far as I remember you can then use `args(...)` to pass the provider, but something to try tout

gradle/build-logic/check-dexer/src/main/kotlin/ProjectExtension.kt
```diff
@@ -0,0 +1,12 @@
+import org.gradle.api.Project
+
+/**
+ * Returns a project with the given name prefixed with the root project name.
+ * For example, for module name "core" and root project "atrium", returns project ":atrium-core"
+ */
+fun Project.getPrefixedProjectName(name: String, rootProject: String): Project =
```
robstoll, line 7: no need for this file once you depend on projects.basics

gradle/build-logic/check-dexer/src/main/kotlin/build-logic.check-dexer.gradle.kts
```diff
@@ -0,0 +1,79 @@
...
+val extension = extensions.create<AtriumDexerExtension>("dexer")
+
+/**
+ * Validates that the ATRIUM_ANDROID_JAR environment variable is properly set
+ * to a valid android.jar file path.
+ *
+ * @throws GradleException if the environment variable is missing or points to a non-existent file
+ */
+fun validateAndroidJarPath() {
+    val androidJarPath = System.getenv("ATRIUM_ANDROID_JAR")
```
robstoll, line 18: We could move ATRIUM_ANDROID_JAR to a constant, we use it in several places, but up to you, I also guess that it won't change for the next 10 years or so :laughing:

gradle/build-logic/check-dexer/src/main/kotlin/DexerUtils.kt
```diff
@@ -0,0 +1,128 @@
...
+ * Configures the dexer settings for a given project:
+ * - Adds required repositories
+ * - Sets up r8 and d8 configurations
+ * - Registers the checkDexer task that compiles the project's bytecode with D8
+ *
+ * @param project The project to configure
+ * @param dexerPreCheckTask The prerequisite check task
+ */
+fun configureDexerForProject(project: Project, dexerPreCheckTask: Provider<Task>) {
+    // Setup repositories
```
robstoll, line 23: please remove the comment I don't think it adds value

gradle/build-logic/check-dexer/src/main/kotlin/DexerUtils.kt
```diff
@@ -0,0 +1,128 @@
...
+    project.repositories.apply {
```
robstoll, line 24: same question as elsewhere, is there a difference in using apply or just `repositories {` if there isn't then I would prefer the later (also in the rest of the code)

gradle/build-logic/check-dexer/src/main/kotlin/build-logic.check-dexer.gradle.kts
```diff
@@ -0,0 +1,79 @@
...
+        )
+    }
+
+    if (!File(androidJarPath).exists()) {
+        throw GradleException(
+            "ATRIUM_ANDROID_JAR environment variable points to a non-existing file: $androidJarPath"
+        )
+    }
+
+    logger.lifecycle("Using android.jar from: $androidJarPath")
```
robstoll, line 33: nice :+1:

gradle/build-logic/check-dexer/src/main/kotlin/build-logic.check-dexer.gradle.kts
```diff
@@ -0,0 +1,79 @@
...
+
+// Configure prerequisite check task
+val dexerPreCheckTask = tasks.register("dexerPreCheck") {
+    doFirst {
+        validateAndroidJarPath()
+    }
+}
+
+// Resolve the list of projects to check for dexing compatibility
+val dexerProjects: List<Project> = extension.subprojects
```
robstoll, line 44: could again be lazily then it would be something along the line of
```
extension.subprojects.map { list -> list.map { prefixedProject(it) } }
```

robstoll (commented):
@amasotti thanks for your work, I have a few comments here and there. We can also merge it as is, see them as suggestions for improvment. Let me know what you want to do

gradle/build-logic/check-dexer/build.gradle.kts
amasotti: done :)

gradle/build-logic/check-dexer/src/main/kotlin/DexerUtils.kt
amasotti: done :)

gradle/build-logic/check-dexer/build.gradle.kts
amasotti: nice, thanks!

gradle/build-logic/check-dexer/src/main/kotlin/DexerUtils.kt
amasotti: done :)

gradle/build-logic/check-dexer/src/main/kotlin/DexerUtils.kt
amasotti: done :)

gradle/build-logic/check-dexer/src/main/kotlin/DexerUtils.kt
amasotti: This is I think the only suggestions I wouldn't follow. The two lines below (`dependsOn`) refer to the checkDexer task, not the jvmJar. 
I tried with a lazy loading, but I start getting problems in compiling Kotlin. :(

gradle/build-logic/check-dexer/src/main/kotlin/ProjectExtension.kt
amasotti: api.basics imported and removed this file all together

gradle/build-logic/check-dexer/src/main/kotlin/build-logic.check-dexer.gradle.kts
amasotti: declared as `const val` in the AtriumDexerExtension class.

gradle/build-logic/check-dexer/src/main/kotlin/DexerUtils.kt
amasotti: I removed it below, you're totally right. In those cases `.apply` is redundant. In this case however it didn't work for some reasons. Repositories seem to require the scope function.

gradle/build-logic/check-dexer/src/main/kotlin/build-logic.check-dexer.gradle.kts
amasotti: Done, thanks for the hint!

amasotti:
@robstoll Thanks a lot for the code review!

I learned quite a few things in this PR and improvement suggestions are ALWAYS welcome ;)

gradle/build-logic/check-dexer/src/main/kotlin/DexerUtils.kt
```diff
@@ -0,0 +1,118 @@
...
+import java.io.OutputStream
+
+fun configureDexerForProject(project: Project, dexerPreCheckTask: Provider<Task>) {
+
+    project.repositories.apply {
+        google()
+        mavenCentral()
+    }
+
+    configureR8(project)
```
robstoll, line 21: Nitpick: I would use twice the same naming:
```suggestion
    configureR8Configuration(project)
```

or twice without Configuration

gradle/build-logic/check-dexer/src/main/kotlin/DexerUtils.kt
robstoll: Fine with me, let me know if you want to try to solve the compilation problem together. Alternatively, I might take a look and create a follow-up PR if I have time and can share my solution if I should find one.

robstoll (approved):

gradle/build-logic/check-dexer/src/main/kotlin/DexerUtils.kt
antonio-masotti: Agree, but I preferred `configureR8` and `configureD8` as consistent naming. 
Somehow `configure..Configuration` sounds redundant :)

robstoll (approved):

robstoll:
@amasotti finally, we don't have any groovy code left 🎉 thanks

there aren't any good first issues left but I have some in mind which I could write done. Do you fancy to do a third one?

PR #1972, opened 2025-05-12
chore: Replace usage of kotlin js plugin in `samples/js/kotlin-test` with custom kotlin-js build logic

ragul-engg:
- As Kotlin js plugin is deprecated, the custom build logic has been used which in turns uses multiplatform plugin
- Reference: https://kotlinlang.org/docs/multiplatform-compatibility-guide.html#migration-from-kotlin-js-gradle-plugin-to-kotlin-multiplatform-gradle-plugin
- resolves: #1887


samples/js/kotlin-test/build.gradle.kts
```diff
@@ -22,7 +22,7 @@ dependencies {
 }
 
 plugins {
-    kotlin("js") version "2.1.20"
+    id("build-logic.kotlin-js")
```
robstoll, line 25: We cannot use our internal build-logic in samples. The intend is that someone could copy the sample project to to have an easier start.

robstoll (changes requested):

samples/js/kotlin-test/build.gradle.kts
ragul-engg: Cool, didn't know that. I have fixed it now 👍

samples/js/kotlin-test/build.gradle.kts
```diff
@@ -43,4 +33,22 @@ kotlin {
             }
         }
     }
+
+    sourceSets {
+        val jsMain by getting {
+            dependencies {
+                implementation("org.jetbrains.kotlin:kotlin-stdlib-js")
+            }
```
robstoll, line 41: is this one necessary? I thought the kotlin gradle plugin is setting this one up automatically

samples/js/kotlin-test/build.gradle.kts
```diff
@@ -43,4 +33,22 @@ kotlin {
...
+        }
+
+        val jsTest by getting {
+            dependencies {
+                // setup for Atrium:
+                implementation("ch.tutteli.atrium:$atriumApi:$atriumVersion")
+
+                // setup for mocha:
+                implementation("org.jetbrains.kotlin:kotlin-test-js")
```
robstoll, line 50: same here, is this really required?

samples/js/kotlin-test/build.gradle.kts
ragul-engg: In the reference doc i attached in commit, it says to add the js deps has to be added inside the sourcesets.

samples/js/kotlin-test/build.gradle.kts
ragul-engg: Same as above

robstoll:
```suggestion
                implementation(kotlin("test"))
```

samples/js/kotlin-test/build.gradle.kts
ragul-engg: Yeah, you are right the multiplatform plugin is doing it for us. It worked when i remove this stdlib dep.

samples/js/kotlin-test/build.gradle.kts
ragul-engg: done fixed it

robstoll (approved):

robstoll:
@ragul-engg thanks 🙂👍

PR #2031, opened 2025-06-27
#1960 migrate char sequence to contain not to contain expectations spec

anson-ryea:
Migrate CharSequencetoContainNotToContainExpectationsSpec to kotlin-test #1960

Replace CharSequencetoContainNotToContainExpectationsSpec tests with AbstractCharSequencetoContainNotToContainExpectationsTest base class
CharSequencetoContainNotToContainExpectations tests in atrium/api/fluent/en_GB/and atrium/api/infix/en_GB/ has been separated into CharSequencetoContainNotToContainBuilderExpectationsTest and CharSequencetoContainNotToContainShortcutExpectationsTest
All build tests has passed


misc/atrium-specs/src/commonMain/kotlin/ch/tutteli/atrium/specs/integration/AbstractCharSequenceToContainNotToContainExpectationsTest.kt
```diff
@@ -0,0 +1,223 @@
...
+    fun notToContain__subject_empty_string() = testFactory(notToContainSpec) { notToContainFun ->
+        val fluentEmptyString = expect("" as CharSequence)
+
+        it("${notToContainSpec.name} 'Hello' does not throw") {
+            fluentEmptyString.notToContainFun("Hello")
+        }
+    }
+
+    @TestFactory
+    fun toContain__subject_text_search_for_hello_and_robert() = testFactory(toContainSpec) { toContainFun ->
```
robstoll, line 58: not sure, would it make sense to rename to the following to make it clear in which case we search? (of course then you would need to put `'hello' and 'robert' into an own method
```suggestion
    fun toContain__subject_text_search_for_Hello_and_Robert() = testFactory(toContainSpec) { toContainFun ->
```

anson-ryea: Yea, I think so. In this sense let me split Hello and notInThere as well.

Issue #2051, opened 2025-07-22
Expect.toBeInRange


robstoll:
*Platform* (all, jvm, js): all
*Extension* (none, kotlin 1.3): none

## Code related feature
```kotlin
expect(x).toBeInRange(1..10)
expect(x).toBeInRange(1 until 10)

//instead of

expect(x) {
  toBeGreaterThanOrEqualTo(1)
  toBeLessThanOrEqualTo(10)
}
expect(x) {
  toBeGreaterThanOrEqualTo(1)
  toBeLessThan(10)
}
```

Following the things you need to do:

*logic*
- [ ] create RangeAssertions with a function `toBeInRange` (see e.g. FloatingPointAssertions as a guideline)
- [ ] implement toBeInRange in DefaultRangeAssertions.kt - make sure both bounds are evaluated (no fail fast behaviour)

*api-fluent*
- [ ] create rangeExpectations.kt with 3 times fun `Expect<T>.toBeInRange(t: T): Expect<T>` for T= CharRange, IntRange, LongRange
- [ ] create a Test named AbstractRangeExpectationsSpec in atrium-specs -> commonMain  (see for instance AbstractComparableExpectationsTest) and extend it in atrium-api-fluent -> commonTest (name it RangeExpectationsSpec)
- [ ] create RangeExpectationSamples.kt with 3 functions in commonTest -> samples (see other samples as guideline -- try to provide the reason why an expectation fails)
- [ ] add `@sample` with link to your sample method to the 3 functions in rangeExpectations.kt
- [ ] add `@since 1.3.0` (adapt to current [milestone](https://github.com/robstoll/atrium/milestones)) to KDOC of the three functions in RangeExpectations.kt

*api-infix*
- [ ] create rangeExpectations.kt with 3 times fun `Expect<T>.toBeInRange(t: T): Expect<T>` for T= CharRange, IntRange, LongRange
- [ ] create RangeExpectationsTest in atrium-api-infix which extends AbstractRangeExpectationsSpec
- [ ] add samples to RangeExpectationSamples.kt (3 fun again)
- [ ] add `@sample` with link to your sample method to the 3 functions in rangeExpectations.kt
- [ ] add `@since 1.3.0` (adapt to current [milestone](https://github.com/robstoll/atrium/milestones)) to KDOC of the two functions in RangeExpectations.kt

## Your first contribution?
- Write a comment `I'll work on this` if you would like to take this issue over. 
[...]

PR #2055, opened 2025-07-25
#2051 add toBeInRange expectations

matcha2smiley:
add toBeInRange expectations
---

### DIFF
--- a/apis/fluent/atrium-api-fluent/src/commonMain/kotlin/ch/tutteli/atrium/api/fluent/en_GB/rangeExpectations.kt
+++ b/apis/fluent/atrium-api-fluent/src/commonMain/kotlin/ch/tutteli/atrium/api/fluent/en_GB/rangeExpectations.kt
@@ -1,0 +1,41 @@
+package ch.tutteli.atrium.api.fluent.en_GB
+
+import ch.tutteli.atrium.logic.isInRange
+import ch.tutteli.atrium.creating.Expect
+import ch.tutteli.atrium.logic._logicAppend
+
+/**
+ * Expects that the subject of `this` expectation is within the [expected] range (inclusive).
+ *
+ * @return an [Expect] for the subject of `this` expectation.
+ *
+ * @sample ch.tutteli.atrium.api.fluent.en_GB.samples.RangeExpectationSamples.toBeInRangeChar
+ *
+ * @since 1.3.0
+ */
+fun Expect<Char>.toBeInRange(expected: CharRange): Expect<Char> =
+    _logicAppend { isInRange(expected) }
+
+/**
+ * Expects that the subject of `this` expectation is within the [expected] range (inclusive).
+ *
+ * @return an [Expect] for the subject of `this` expectation.
+ *
+ * @sample ch.tutteli.atrium.api.fluent.en_GB.samples.RangeExpectationSamples.toBeInRangeInt
+ *
+ * @since 1.3.0
+ */
+fun Expect<Int>.toBeInRange(expected: IntRange): Expect<Int> =
+    _logicAppend { isInRange(expected) }
+
+/**
+ * Expects that the subject of `this` expectation is within the [expected] range (inclusive).
+ *
+ * @return an [Expect] for the subject of `this` expectation.
+ *
+ * @sample ch.tutteli.atrium.api.fluent.en_GB.samples.RangeExpectationSamples.toBeInRangeLong
+ *
+ * @since 1.3.0
+ */
+fun Expect<Long>.toBeInRange(expected: LongRange): Expect<Long> =
+    _logicAppend { isInRange(expected) }
--- a/apis/fluent/atrium-api-fluent/src/commonMain/kotlin/ch/tutteli/atrium/api/fluent/en_GB/rangeExpectations.kt
+++ b/apis/fluent/atrium-api-fluent/src/commonMain/kotlin/ch/tutteli/atrium/api/fluent/en_GB/rangeExpectations.kt
@@ -1,11 +1,11 @@
 package ch.tutteli.atrium.api.fluent.en_GB
 
-import ch.tutteli.atrium.logic.isInRange
+import ch.tutteli.atrium.logic.toBeInRange
 import ch.tutteli.atrium.creating.Expect
 import ch.tutteli.atrium.logic._logicAppend
 
 /**
- * Expects that the subject of `this` expectation is within the [expected] range (inclusive).
+ * Expects that the subject of `this` expectation is within the [expected] range.
  *
  * @return an [Expect] for the subject of `this` expectation.
  *
@@ -14,10 +14,10 @@ import ch.tutteli.atrium.logic._logicAppend
  * @since 1.3.0
  */
 fun Expect<Char>.toBeInRange(expected: CharRange): Expect<Char> =
-    _logicAppend { isInRange(expected) }
+    _logicAppend { toBeInRange(expected) }
 
 /**
- * Expects that the subject of `this` expectation is within the [expected] range (inclusive).
+ * Expects that the subject of `this` expectation is within the [expected] range.
  *
  * @return an [Expect] for the subject of `this` expectation.
  *
@@ -26,10 +26,10 @@ fun Expect<Char>.toBeInRange(expected: CharRange): Expect<Char> =
  * @since 1.3.0
  */
 fun Expect<Int>.toBeInRange(expected: IntRange): Expect<Int> =
-    _logicAppend { isInRange(expected) }
+    _logicAppend { toBeInRange(expected) }
 
 /**
- * Expects that the subject of `this` expectation is within the [expected] range (inclusive).
+ * Expects that the subject of `this` expectation is within the [expected] range.
  *
  * @return an [Expect] for the subject of `this` expectation.
  *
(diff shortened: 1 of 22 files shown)


logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultRangeAssertions.kt
```diff
@@ -0,0 +1,63 @@
...
+            assertionBuilder.explanatoryGroup
+                .withDefaultType
+                .withAssertions(explanatoryAssertionCreator(subject))
+                .build()
+        }
+        .withDescriptionAndRepresentation(
+            TranslatableWithArgs(
+                DescriptionComparableExpectation.TO_BE_IN_RANGE,
+                "${range.start}..${range.endInclusive}"
+            ),
```
robstoll, line 60: this will result in an error as your translatable does not define 2 placeholders. But I don't think you need them. otherwise the report will look as follows:
```
I expected that the subject: 100
- to be in range 1...20: 1..20
```
I.e. it is enough to just use your description as description as follow:
```suggestion
                DescriptionComparableExpectation.TO_BE_IN_RANGE,
```

logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultRangeAssertions.kt
```diff
@@ -0,0 +1,63 @@
...
+    container: AssertionContainer<out T>,
+    range: ClosedRange<T>,
+    explanatoryAssertionCreator: (T) -> List<Assertion>
+): Assertion =
+    assertionBuilder.descriptive
+        .withTest(container.toExpect()) { it >= range.start && it <= range.endInclusive }
+        .withHelpOnFailureBasedOnDefinedSubject(container.toExpect()) { subject ->
+            assertionBuilder.explanatoryGroup
+                .withDefaultType
+                .withAssertions(explanatoryAssertionCreator(subject))
```
robstoll, line 53: note, this will show both hints, even though one of it is not necessary IMO as it hols. Following an example:
```
expect(10).toBeInRange(1 .. 2)
```
would show:
```
I expected subject: 10
- to be in range: 1..2
    » to be greater than or equal to: 1 
    » to be less than or equal to: 2
```    
Yet, if we just put it like that, I am not sure the user will understand the hint immediately. How about we turn this into:
```
I expected subject: 10
- to be in range: 1..2
  » was not less than or equal to: 2
```
Looking at it, I am not even sure we need the extra hint (I guess you came up with it because I pointed you to DefaultFloatingPointAssertions, that was a mistake on my part, I should have pointed you to DefaultComparableAssertions.kt -- I fixed it for the tests but forgot it apparently for logic

WDYT, would 
```
I expected subject: 10
- to be in range: 1..2
 ```
 be enough as well or do we need the hint?

**Review by robstoll — COMMENTED** [review]
Looks already good :slightly_smiling_face: 
One thing I forgot to write down in the issue. Once you are done with the logic part you need to execute `./gradlew generate` this will create a range.kt into logic which you then can use in your apis.

File: logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultRangeAssertions.kt (hunk 2 above)
Comment (matcha2smiley) [reply logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultRangeAssertions.kt]: Really appreciate your review!

I completely agree — I also think the minimal version is clearer and sufficient.
Thank you for pointing it out!

The reason I chose the minimal implementation is that, in the case of the toBeInRange assertion, I believe it's usually enough to simply know that the value is outside the specified range.
In most use cases, it's not particularly important whether the value is below the lower bound or above the upper bound — just knowing that it's out of range is enough to achieve the intended purpose.

That said, I can imagine that in the future, there might be scenarios — for example during debugging — where it could be helpful to know which direction the value deviates.
In such cases, it might make sense to add an option for showing more detailed output. That’s just a thought I had, though — nothing concrete at this point.
Given that there’s no specific use case for it at the moment and considering the implementation cost, I think sticking with the minimal version is the right choice for now.

File: logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultRangeAssertions.kt (hunk 1 above)
Comment (matcha2smiley) [reply logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultRangeAssertions.kt]: Thanks for the suggestion. It’s been fixed!

File: apis/fluent/atrium-api-fluent/src/commonMain/kotlin/ch/tutteli/atrium/api/fluent/en_GB/rangeExpectations.kt [hunk 3]
```diff
@@ -0,0 +1,41 @@
+package ch.tutteli.atrium.api.fluent.en_GB
+
+import ch.tutteli.atrium.logic.isInRange
+import ch.tutteli.atrium.creating.Expect
+import ch.tutteli.atrium.logic._logicAppend
+
+/**
+ * Expects that the subject of `this` expectation is within the [expected] range (inclusive).
```
robstoll, line 8: detail, you can also keep as is. Personally, I would remove the `(inclusive)` although it is correct
```suggestion
 * Expects that the subject of `this` expectation is within the [expected] range.
```
because this could confuse someone who writes
 ```
expect('d').toBeInRange('a' until 'c')
```
Where `c` is exclusive (behind the seen it is still a ClosedRange but endInclusive is `b`.) This comes apparent in error reporting, the user will see `a .. b` instead of `a ..< c` (I wonder if kotlin will introduce an own type at some point, something like IntOpenEndRange).

File: logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/RangeAssertions.kt [hunk 4]
```diff
@@ -0,0 +1,8 @@
+package ch.tutteli.atrium.logic
+
+import ch.tutteli.atrium.assertions.Assertion
+import ch.tutteli.atrium.creating.AssertionContainer
+
+interface RangeAssertions {
+    fun <T : Comparable<T>> isInRange(container: AssertionContainer<T>, range: ClosedRange<T>): Assertion
```
robstoll, line 7: please keep toBeInRange, it's a legacy that some methods in other Assertions start with is. Using toBeInRange will simplify migration in the future. Thanks
```suggestion
    fun <T : Comparable<T>> toBeInRange(container: AssertionContainer<T>, range: ClosedRange<T>): Assertion
```

logic/atrium-logic/src/commonMain/kotlin/ch/tutteli/atrium/logic/impl/DefaultRangeAssertions.kt
```diff
@@ -0,0 +1,25 @@
...
+        container: AssertionContainer<T>,
+        range: ClosedRange<T>
+    ): Assertion =
+        assertionBuilder.descriptive
+            .withTest(container.toExpect()) { it >= range.start && it <= range.endInclusive }
+            .withDescriptionAndRepresentation(
+                DescriptionComparableExpectation.TO_BE_IN_RANGE,
+                range,
+            )
+            .build()
```
robstoll, line 24:
```suggestion
    ): Assertion = container.createDescriptiveAssertion(DescriptionComparableExpectation.TO_BE_IN_RANGE, range) {
        it >= range.start && it <= range.endInclusive
    }
```

robstoll (commented):
I'll will no stop reviewing until you give a sign again, happy coding :slightly_smiling_face:

matcha2smiley:
Hi @robstoll , Thak you for your kindness👋
I was looking at this issue and had a quick question regarding the test implementation plan.

Looking at recent migration PRs like [#2031](https://github.com/robstoll/atrium/pull/2031), it seems that the project is moving away from the *Spec style (e.g., RangeExpectationsSpec using Spek) towards a *Test style with @TestFactory and standard JUnit5 classes.

In this issue, the current implementation plan suggests creating AbstractRangeExpectationsSpec and extending it in RangeExpectationsSpec.
Would it be better to follow the newer *Test style instead?

Just wanted to make sure I'm aligned with the latest testing conventions before I proceed.
Thanks in advance!

robstoll:
@matcha2smiley I took a look at the good-first-issue template again and realised that it did not include the mention of apiDump. See updated issue description, you need to execute `./gradlew apiDump & KOTLIN_VERSION=1.9 ./gradlew apiDump` whenever you add something new otherwise `./gradlew build` will fail at jvmApiCheck

robstoll:
@matcha2smiley thanks for your first contribution to Atrium :slightly_smiling_face: :+1: 

How about another one? And don't forget to :star: Atrium if you liked it, thanks

matcha2smiley:
> @matcha2smiley thanks for your first contribution to Atrium 🙂 👍
> 
> How about another one?

@robstoll I'm happy to contribute to Atrium! 😆
Actually, I'm considering working on one of the migration issues next.
Do you have any migration issues you'd prefer me to tackle first?
If not, I'll just start from the top. 🙂

PR #2081, opened 2025-10-16
#2012 migrate IterableExpectationsSpec to kotlin-test

FelixSchwer:

### DIFF
--- a/apis/fluent/atrium-api-fluent/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/IterableExpectationsTest.kt
+++ b/apis/fluent/atrium-api-fluent/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/IterableExpectationsTest.kt
@@ -1,45 +1,47 @@
 package ch.tutteli.atrium.api.fluent.en_GB
 
+import ch.tutteli.atrium.api.verbs.internal.expect
 import ch.tutteli.atrium.creating.Expect
 import ch.tutteli.atrium.specs.*
+import kotlin.test.Test
 
-object AbstractIterableExpectationsTest : ch.tutteli.atrium.specs.integration.AbstractIterableExpectationsTest(
+object IterableExpectationsTest : ch.tutteli.atrium.specs.integration.AbstractIterableExpectationsTest(
     fun0(Expect<Iterable<Int>>::toHaveElements),
     fun0(Expect<Iterable<Int>>::notToHaveElements),
-    feature0<Iterable<Int>, Int>(Expect<Iterable<Int>>::min),
-    fun1<Iterable<Int>, Expect<Int>.() -> Unit>(Expect<Iterable<Int>>::min),
-    feature0<Iterable<Int>, Int>(Expect<Iterable<Int>>::max),
-    fun1<Iterable<Int>, Expect<Int>.() -> Unit>(Expect<Iterable<Int>>::max),
+    feature0(Expect<Iterable<Int>>::min),
+    fun1(Expect<Iterable<Int>>::min),
+    feature0(Expect<Iterable<Int>>::max),
+    fun1(Expect<Iterable<Int>>::max),
     fun0(Expect<Iterable<Int>>::toHaveElementsAndNoDuplicates),
-    property<Iterable<Int>, Int>(Expect<Iterable<Int>>::last),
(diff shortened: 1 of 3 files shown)


apis/fluent/atrium-api-fluent/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/IterableExpectationsTest.kt
```diff
@@ -0,0 +1,49 @@
...
+    fun1(Expect<Iterable<Int>>::last),
+) {
+
+    @Test
+    fun ambiguityTest() {
+        val a1 = expect(listOf(1.0, 2.0))
+        val a1b = expect(setOf(1.0, 2.0, null))
+        val star = expect(listOf(1, 2, 3) as Collection<*>)
+
+        a1.toHaveElements()
```
robstoll: its quite some time ago where we had kotlin type inference issues (bugs) where kotlin (I think kotlin 1.3.x) inferred a wrong type and we were not be able to assign this expression to a1. Kotlin improved quite a bit since then but still the ambiguityTest is meant to detect compiler ambiguities. So I suggest we keep the assignments even when it means we need to use `var` and add a `@Suppress("AssignedValueIsNeverRead")` on the function

apis/fluent/atrium-api-fluent/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/IterableExpectationsTest.kt
```diff
@@ -0,0 +1,49 @@
+package ch.tutteli.atrium.api.fluent.en_GB
+
+import ch.tutteli.atrium.api.verbs.internal.expect
+import ch.tutteli.atrium.creating.Expect
+import ch.tutteli.atrium.specs.*
+import kotlin.test.Test
+
+object IterableExpectationsTest : ch.tutteli.atrium.specs.integration.AbstractIterableExpectationsTest(
```
robstoll:
```suggestion
class IterableExpectationsTest : ch.tutteli.atrium.specs.integration.AbstractIterableExpectationsTest(
```
kotlin-test doesn't support `object` that's something I have to note in the description

apis/fluent/atrium-api-fluent/src/commonTest/kotlin/ch/tutteli/atrium/api/fluent/en_GB/IterableExpectationsTest.kt
```diff
@@ -0,0 +1,49 @@
...
+) {
+
+    @Test
+    fun ambiguityTest() {
+        val a1 = expect(listOf(1.0, 2.0))
+        val a1b = expect(setOf(1.0, 2.0, null))
+        val star = expect(listOf(1, 2, 3) as Collection<*>)
+
+        a1.toHaveElements()
+        a1.notToHaveElements()
```
robstoll: this expectation will fail as a1 has elements (also other expectations here, execute the test to ensure everything passes)

apis/infix/atrium-api-infix/src/commonTest/kotlin/ch/tutteli/atrium/api/infix/en_GB/IterableExpectationsTest.kt
```diff
@@ -26,10 +28,10 @@ class IterableExpectationsSpec : ch.tutteli.atrium.specs.integration.IterableExp
         private fun getNotToHaveElementsPair() = "${notToHave.name} ${elements::class.simpleName}" to Companion::notToHaveElements
         private fun notToHaveElements(expect: Expect<Iterable<Int>>) = expect notToHave elements
 
-        private fun minFeaturePair() = feature1<Iterable<Int>, o, Int>(Expect<Iterable<Int>>::min).name to ::minFeature
+        private fun minFeaturePair() = feature1(Expect<Iterable<Int>>::min).name to ::minFeature
```
robstoll: I run into compiler ambiguity issues, re-add the type hints (same same for max and last below)

robstoll:
@FelixSchwer No worries, we will take over from here. Thanks for the pre-work 🙂👍

PR #2164, opened 2026-05-21
correct capitalization in a usage hint

liblit:
`Text.Empty` does not exist. The recommendation should refer to [`Text.EMPTY`](https://github.com/robstoll/atrium/blob/7b4aecc290f415db1931c9a133fbc0e85041c863/atrium-core/src/commonMain/kotlin/ch/tutteli/atrium/reporting/Text.kt#L34-L37):

https://github.com/robstoll/atrium/blob/7b4aecc290f415db1931c9a133fbc0e85041c863/atrium-core/src/commonMain/kotlin/ch/tutteli/atrium/reporting/Text.kt#L34-L37

### DIFF
--- a/atrium-core/src/commonMain/kotlin/ch/tutteli/atrium/reporting/Text.kt
+++ b/atrium-core/src/commonMain/kotlin/ch/tutteli/atrium/reporting/Text.kt
@@ -22,7 +22,9 @@ data class Text private constructor(val string: String) {
     companion object {
 
         operator fun invoke(string: String): Text {
-            require(string.isNotEmpty()) { "use Text.EMPTY instead" }
+            require(string.isNotEmpty()) {
+                "use ${Text::class.simpleName}.${Text::EMPTY.name} instead"
+            }
             return Text(string)
         }
 
--- a/atrium-core/src/commonMain/kotlin/ch/tutteli/atrium/reporting/Text.kt
+++ b/atrium-core/src/commonMain/kotlin/ch/tutteli/atrium/reporting/Text.kt
@@ -22,7 +22,7 @@ data class Text private constructor(val string: String) {
     companion object {
 
         operator fun invoke(string: String): Text {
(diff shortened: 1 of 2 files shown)


robstoll:
@liblit thanks for the fix, there is even a test which checks that the the wrong hint is thrown 😄 can you fix it as well please

liblit:
> there is even a test which checks that the the wrong hint is thrown 😄 can you fix it as well please

Ah, sorry for missing that earlier. Now fixed.

liblit:
FYI, I considered:

```kotlin
require(string.isNotEmpty()) {
    "use ${Text::class.simpleName}.${Text::EMPTY.name} instead"
}
```

This style has the advantage that it will fail to compile if `Text.EMPTY` changes its name again in the future. But it also seems rather over-engineered.

robstoll:
> FYI, I considered:
> 
> ```kotlin
> require(string.isNotEmpty()) {
>     "use ${Text::class.simpleName}.${Text::EMPTY.name} instead"
> }
> ```
> 
> This style has the advantage that it will fail to compile if `Text.EMPTY` changes its name again in the future. But it also seems rather over-engineered.

I usually do this as well, then wen you rename a field (maybe it was actually named Text.Empty in the very beginning of Atrium) the error message is renamed automatically as well. But we can also keep it as is, up-to-you.

liblit:
> I usually do this as well, then wen you rename a field (maybe it was actually named Text.Empty in the very beginning of Atrium) the error message is renamed automatically as well.

My thoughts exactly! I've updated this pull request accordingly.

robstoll:
@liblit thanks for your second contribution to Atrium :slightly_smiling_face: :+1:

robstoll (approved):

liblit:
> @liblit thanks for your second contribution to Atrium :slightly_smiling_face: :+1: 

My pleasure. Thank _you_ for creating and maintaining this delightful library!

PR #2177, opened 2026-06-25
ci: generate the test matrix with @vlsi/github-actions-random-matrix

vlsi:
## Why

The CI test matrix was built from a vendored copy of `matrix_builder.js` plus the tegonal-commons `matrix_commons.js` helper. The random-matrix logic had to be kept in sync by hand, and the matrix was assembled with several `generateRow()` calls whose order determined the final job count. The upstream project now ships the same logic as the npm package `@vlsi/github-actions-random-matrix`, which also added a `generateRows(maxRows, {require})` API that builds the whole matrix in one call.

## What

- Add `.github/workflows/package.json` and `package-lock.json` pinning `@vlsi/github-actions-random-matrix@2.1.0`.
- Add `.github/workflows/matrix.mjs`: a self-contained ESM script that defines the `java_distribution`, `java_version`, and `os` axes, the existing exclusions (`graalvm` + Java 11/17, `dragonwell` + macOS), and builds the matrix in a single `generateRows(Number(process.env.MATRIX_JOBS || 4), {require: [...]})` call. The `require` list keeps the previous guarantees: a row each for the oldest and newest Java version and for one Linux and one Windows job. The remaining `MATRIX_JOBS` budget goes to pairwise coverage, so the job count is now exactly `MATRIX_JOBS` regardless of the requirement order. It uses a seeded RNG (`createGitHubMatrixBuilder`), so a build can be reproduced from the seed printed in the step summary.
- Update `quality-assurance.yml` to run `npm ci --prefix .github/workflows` and `node .github/workflows/matrix.mjs` (the runner's preinstalled Node is enough; no `setup-node` step needed).
- Remove `matrix.js`, `matrix_builder.js` (replaced by the package), and `matrix_commons.js` (no longer used).
- Drop the `matrix_commons.js` entry from `.gt/remotes/tegonal-gh-commons/pulled.tsv`. It was the only file vendored from that remote; without this, a future `gt update` would restore a file that references the removed `matrix_builder.js`. The remote registration itself is left in place.

Java 25 is still held back (it needs Gradle 9.x, which drops Java 11); the explanatory comment moved into `matrix.mjs`.

## How to verify

```
npm ci --prefix .github/workflows
node .github/workflows/matrix.mjs            # default: 4 jobs
MATRIX_JOBS=6 node .github/workflows/matrix.mjs   # 6 jobs
```

Each generated row carries `java_version`, `java_distribution`, and `os`, matching what the `build-and-test` job consumes. Across 200 seeds the output had no exclusion violations and no missing requirement (oldest/newest Java, ubuntu, windows).


.github/workflows/quality-assurance.yml
```diff
@@ -15,7 +15,9 @@ jobs:
 
       - name: Generate matrix
         id: set-matrix
-        run: node .github/workflows/matrix.js
+        run: |
+          npm ci --prefix .github/workflows
```
robstoll, line 19: @vlsi I am a bit confused, vlsi/github-actions-random-matrix isn't a proper gh action or did the llm just make things up?

vlsi: I haven't considered creating an action for it.
The current state is that I published `matrix-builder` to npmjs so I could reuse the same matrix-builder code across multiple projects.
It is a typical npm library, so the usage is that you fetch the library with npm ci as per `project.json`.

Frankly, I'm not sure what would be the benefits of creating an action to factor this step.

Having package.json in git enables humans and agents to test the matrix locally.

Does that clarify?

robstoll: Thanks for this bump, I'll update tegonal/github-commons first and then it will update here more or less automatically. Hence closing this PR once this is done (keep it open right now as a reminder for me)

